PR #35, opened 2018-09-03
TF doc: guide/datasets: int32 is not supported in Example proto.

wangtz:
Error:
TypeError: Value passed to parameter 'dense_defaults' has DataType int32 not in list of allowed values: float32, int64, string


wangtz:
@MarkDaoust

PR #44, opened 2018-09-06
Created using Colaboratory

MarkDaoust:
This autograph guide is broken on nightly. replacing `range` with `tf.range` works 

cc: @mdanatg


mdanatg:
We are working on a fix, sorry about that.

MarkDaoust:
Thanks Dan.

I just wanted to be sure it was a known issue.

PR #49, opened 2018-09-10
Add plt.show() instruction

Ge0:
Without the `plt.show()` instruction, one does not always know how to actually display the GUI through matplotlib/tkinter.

This PR updates the documentation accordingly.


Ge0:
Signed.

MarkDaoust:
Hi @Ge0, 

You're [not the first to suggest this](https://github.com/tensorflow/models/pull/5166#issuecomment-415468020)

But I'm not convinced of the value, as it's unenforced, extra code that has no effect for tpical users.

We'll look farther into this if it keeps coming up.

But thanks for taking the time to make the PR.

PR #63, opened 2018-09-18
Add missing return in example code

bertptrs:
Keras custom layers example code missed a return statement, which
caused layers to not have a config when the example was followed.


bertptrs:
I signed!

PR #74, opened 2018-09-21
Cleanup on the Hadoop page

superbobry:
* Removed an unrelated reference to HADOOP-10903. The issue describes the addition of the --glob flag to the `hadoop` command. It does not give any context on why this is necessary. In fact, I've failed to find any reference explaining this behaviour other than this SO answer: https://stackoverflow.com/a/9322747/262432.
* ${} are now used consistently in Bash snippets.

### DIFF
--- a/site/en/deploy/hadoop.md
+++ b/site/en/deploy/hadoop.md
@@ -31,13 +31,15 @@ be set:
     ```shell
     source ${HADOOP_HOME}/libexec/hadoop-config.sh
     ```
+    
+    The variable is optional if libhdfs.so is available in LD_LIBRARY_PATH.
 
 *   **LD_LIBRARY_PATH**: To include the path to libjvm.so, and optionally the path
     to libhdfs.so if your Hadoop distribution does not install libhdfs.so in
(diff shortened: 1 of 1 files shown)


site/en/deploy/hadoop.md
```diff
@@ -31,13 +31,15 @@ be set:
     ```shell
     source ${HADOOP_HOME}/libexec/hadoop-config.sh
     ```
+    
+    The variable is optional if libhdfs.so is available in LD_LIBRARY_PATH.
```
superbobry: Implemented in tensorflow/tensorflow#19336.

site/en/deploy/hadoop.md
```diff
@@ -46,11 +48,8 @@ be set:
     expanded as described in the libhdfs documentation:
 
     ```shell
-    CLASSPATH=$(${HADOOP_HDFS_HOME}/bin/hadoop classpath --glob) python your_script.py
+    CLASSPATH=$(${HADOOP_HOME}/bin/hadoop classpath --glob) python your_script.py
```
superbobry: HADOOP_HDFS_HOME is unlikely to contain the `hadoop` binary.

PR #75, opened 2018-09-22
remove link to lite subsite from install toc

lamberta:
Causing lower tabs to not show up on /lite


lamberta:
Cause that's [The Chicago Way](https://www.youtube.com/watch?v=2ScvAJG51V4)!

PR #76, opened 2018-09-24
Update basic_classification.ipynb

raylexlee:
Bug fix.

### DIFF
--- a/site/en/tutorials/keras/basic_classification.ipynb
+++ b/site/en/tutorials/keras/basic_classification.ipynb
@@ -770,8 +770,7 @@
         "  \n",
         "  plt.xlabel(\"{} {:2.0f}% ({})\".format(class_names[predicted_label],\n",
         "                                100*np.max(predictions_array),\n",
-        "                                class_names[true_label]),\n",
-        "                                color=color)\n",
+        "                                class_names[true_label])).set_color(color)\n"
         "\n",
         "def plot_value_array(i, predictions_array, true_label):\n",
         "  predictions_array, true_label = predictions_array[i], true_label[i]\n",


lamberta:
What's the reason fro this change? Are you getting a warning or error?

raylexlee:
Yesterday I was learning from the beginning up to https://www.tensorflow.org/tutorials/keras/basic_classification#make_predictions. I copied the codes to my python shell and studied the data. When I tried to copy the function block def plot_image(i, predictions_array, true_label, img) ...,  I recieved error messages. After some googling, I managed to get this workable solution. 
Thank you @lamberta . So I went through the steps one by one. This time the original version worked flawlessly. Then I realized that it was the blank lines in the def function block causing problem when the block of codes was pasted to the python shell.

raylexlee:
I am using Ubuntu 16.04.
```
raylex@fk:~$ cd audioplayer/
raylex@fk:~/audioplayer$ source ~/tensorflow/venv/bin/activate
(venv) raylex@fk:~/audioplayer$ python
Python 3.5.2 (default, Nov 23 2017, 16:37:01) 
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> # TensorFlow and tf.keras
... import tensorflow as tf
>>> from tensorflow import keras
```
My result differs in
```
i = 12
plt.figure(figsize=(6,3))
plt.subplot(1,2,1)
plot_image(i, predictions, test_labels, test_images)
plt.subplot(1,2,2)
plot_value_array(i, predictions,  test_labels)
```
Instead my first wrong match occurs at i = 17 with the help of this tiny function
```
def not_matches(predictions_array, true_label):
    for i in range(len(predictions_array)):
        if true_label[i] != np.argmax(predictions_array[i]):
            yield i
```
and the following block of codes
```
i=next(wrong_matches)
plt.figure(figsize=(6,3))
plt.subplot(1,2,1)
plot_image(i, predictions, test_labels, test_images)
plt.subplot(1,2,2)
plot_value_array(i, predictions,  test_labels)
i

plt.show()
```
Of course, `wrong_matches=not_matches()` .
I this manner, I did save the pictures.
pic_17

PR #81, opened 2018-09-24
Cleanup api_guides/

lamberta:
Start cleaning out the api_guides.

First round removes guides that don't don't any guidance. These just list functions, which should be unnecessary in TF 2.0.


lamberta:
Here's what's left: https://github.com/lamberta/docs/tree/remove-api_guides/site/en/api_guides/python

Want to look around for other easy pickings I can get in this CL?

site/en/api_guides/python/index.md
```diff
@@ -1,52 +0,0 @@
-# Python API Guides
```
MarkDaoust: This is an index file, not a list of ops... might want to keep, or add a redirect.

lamberta: I'll add a redirect to /guide
It doesn't look like you can access this page from: https://www.tensorflow.org/api_docs/python/tf ?

lamberta: cl/214344585

PR #93, opened 2018-09-29
Mention 64-bit python need on windows

ramazanpolat:
Tensorflow only supports 64-bit versions of Python 3.5.x and 3.6.x on Windows.
It was mentioned on previous install guides.


lamberta:
Thanks! Can you acknowledge the CLA so we can proceed with this PR?

ramazanpolat:
I signed.

PR #101, opened 2018-10-05
Created Image classification tutorial

kshitizrimal:
I have created Image classification tutorial using TF.Keras and ImageDataGenerator class, of which I am an owner. Please have a look and if everything is fine with the tutorial, I would like to make this pull request.


kshitizrimal:
I signed it!

lamberta:
Very nice, thank you!
Colab preview here: http://colab.research.google.com/github/kshitizrimal/docs/blob/image-classification/site/en/tutorials/keras/image_classification.ipynb

A couple of quick things I see:
- Please delete the .DS_Store and .ipynb_checkpoints/ files that crept into this PR.
- We're saving notebooks without output (output is rendered when imported into tensorflow.org)
- I think we want to keep the tutorials/keras collection limited to the examples from *Deep Learning with Python*, so maybe we can look at /tutorials/images for this?
- In Colab, save notebook with left table of contents open
- We'll need to add an Apache 2 license

kshitizrimal:
Ok thank you .. i will do that

kshitizrimal:
@lamberta about images in your 3rd point.. in the tutorials/images section there are use cases for CIFAR-10 and ImageNet.. does that mean i will have to only use either of them? or meant something else?

kshitizrimal:
I re-checked the book and even in the book, the image dataset used is of cats vs dogs, but there, a local directory path is used for that, but its the same dataset i used here in the tutorial only located in remote server so that anyone can download and use.

lamberta:
We're fine with using other datasets, we just don't want to host them on tensorflow.org :)
If the license is agreeable, we can download it from somewhere—we just need to make it clear where it's coming from.

lamberta:
Thank you for the context.

We need to discuss some things and come back to this. One issue is the use of the the [keras-preprocessing](https://github.com/keras-team/keras-preprocessing) package for your `ImageDataGenerator`. [tf.keras.preprocessing](https://www.tensorflow.org/api_docs/python/tf/keras/preprocessing) and [tf.keras.applications](https://www.tensorflow.org/api_docs/python/tf/keras/applications) are not actually in `tf.keras` but are wrappers around these external packages. They work, they're accessible from `tf.keras`, but they live elsewhere and are not subject to the same API stability guarantees.

While this is definitely useful, and I think a lot of people will find it helpful, we can't quite present this as an *official* example. But, maybe we can create a community section (or something like that) ... but then that opens a different set of questions.

We'll discuss further and see if we can come up with something. Thanks

kshitizrimal:
Sure, I am happy to replace ImageDataGenerator with tf.data api if required. Please let me know what other changes you might need in this notebook and I am happy to do so.

kshitizrimal:
@lamberta if I need to remove keras.preprocessing, I am considering using opencv and numpy to first load and resize the data and use tf.data api to make the generators.. so just wanted to check if this approach is acceptable for this tutorial?

random-forests:
Re: the ImageDataGenerator living elsewhere, my vote is to move this into core. Data augmentation (especially for images) is something basically every developer will need to do. 

Do we have an canonical utility with similar functionality? If not, I suggest we use these, they're proven. If they need to be rewritten to be backed by tf.data or similar, worth it. 

Also in favor of adding this to our official tutorials, rather than a community section, after a bit more cleanup.

random-forests:
@kshitizrimal, please hold off on re-writing that (hoping we can find a standard solution, so everyone doesn't need to roll their own for a common task like this).

kshitizrimal:
> @kshitizrimal, please hold off on re-writing that (hoping we can find a standard solution, so everyone doesn't need to roll their own for a common task like this).

@random-forests sure, just let me know when its all sort out and I will make the changes accordingly.

MarkDaoust:
Sorry about the delay here.

We have talked with the owners, and they agree that if `tf.keras.preprocessing` is the easiest way to do these things, we should show it.

We might re-organize/edit the content a little (maybe break it into smaller documents). But let's not do that in this PR. 

I'm going to merge this, It won't be visible on the site immediately but I'll post back here when it is.

PR #102, opened 2018-10-05
Reorg C++ and Java ref sections

lamberta:


lamberta:
with cl/215164098

site/en/api_docs/cc/_book.yaml
```diff
@@ -0,0 +1,19 @@
+upper_tabs:
+# Tabs left of dropdown menu
+- include: /_upper_tabs_left.yaml
+
+# Dropdown menu
+- name: API
+  path: /api_docs/
```
MarkDaoust: If we're keeping versioned cc and java docs, we either need to add versions/{version}/ at some point, or rely on relative paths.

PR #112, opened 2018-10-09
Remove code highlight from generated text.

MarkDaoust:


lamberta:
Nevermind, I'm just looking at the diff on my phone and misread the change

PR #121, opened 2018-10-12
updates markdown in uses file under about

aritra24:
The previous uses.md file made use of the unordered list and list items tags of html which caused the bold markdown to not render. replaced the &lt;ul&gt; and &lt;li&gt; tags with markdown equivalents.


MarkDaoust:
Thanks.

PR #128, opened 2018-10-17
A first draft of "loading data: images"

MarkDaoust:
It still needs an introduction and conclusion paragraph.

But this won't be visible on the site unless we add it to the "_toc.yaml"

Staging:

https://github.com/MarkDaoust/docs/blob/load_data/site/en/tutorials/load_data/images.ipynb

https://colab.sandbox.google.com/github/MarkDaoust/docs/blob/load_data/site/en/tutorials/load_data/images.ipynb

### DIFF
--- a/site/en/tutorials/load_data/images.ipynb
+++ b/site/en/tutorials/load_data/images.ipynb
@@ -1280,7 +1280,8 @@
       "cell_type": "code",
       "source": [
         "ds = image_label_ds.cache(filename='./cache.tf-data')\n",
-        "ds = ds.repeat().shuffle(buffer_size=4000)\n",
+        "ds = ds.apply(\n",
+        "  tf.data.experimental.shuffle_and_repeat(buffer_size=image_count))\n",
         "ds = ds.batch(BATCH_SIZE).prefetch(1)\n",
         "ds"
       ],
@@ -1340,6 +1341,8 @@
       },
       "cell_type": "markdown",
       "source": [
+        "#### Basic\n",
+        "\n",
         "TFRecord files are a simple format to store a sequence of binary blobs. In this case the fact that these are read linearly off disk, which may give some performance boost.\n",
         "\n",
         "It also gives you a portable data file that is apropriate, for example, for streaming over a network.\n",
@@ -1380,7 +1383,7 @@
       },
       "cell_type": "code",
       "source": [
-        "image_ds = tf.data.TFRecordDataset('images.tfrec').map(preprocess_image, num_parallel_calls=8)"
+        "image_ds = tf.data.TFRecordDataset('images.tfrec').map(preprocess_image)"
       ],
       "execution_count": 0,
       "outputs": []
@@ -1404,8 +1407,9 @@
       "cell_type": "code",
       "source": [
         "ds = tf.data.Dataset.zip((image_ds, label_ds))\n",
-        "ds = ds.repeat().shuffle(buffer_size=4000)\n",
-        "ds=ds.batch(BATCH_SIZE).prefetch(1)\n",
+        "ds = ds.apply(\n",
+        "  tf.data.experimental.shuffle_and_repeat(buffer_size=image_count))\n",
+        "ds=ds.batch(BATCH_SIZE).prefetch(AUTOTUNE)\n",
         "ds"
       ],
       "execution_count": 0,
@@ -1433,6 +1437,155 @@
       "source": [
         "This is slower than the `cache` version because we have not cached the preprocessing."
       ]
+    },
+    {
+      "metadata": {
+        "id": "NF9W-CTKkM-f",
+        "colab_type": "text"
+      },
+      "cell_type": "markdown",
+      "source": [
+        "#### preprocessed"
+      ]
+    },
+    {
+      "metadata": {
+        "id": "J9HzljSPkxt0",
+        "colab_type": "text"
+      },
+      "cell_type": "markdown",
+      "source": [
+        "To save some preprocessing to the TFRecord file, first make a dataset of the processed images, as before:"
+      ]
+    },
+    {
+      "metadata": {
+        "id": "OzS0Azukkjyw",
+        "colab_type": "code",
+        "colab": {}
+      },
+      "cell_type": "code",
+      "source": [
+        "paths_ds = tf.data.Dataset.from_tensor_slices(all_image_paths)\n",
+        "image_ds = paths_ds.map(load_and_preprocess_image)\n",
+        "image_ds"
+      ],
+      "execution_count": 0,
+      "outputs": []
+    },
+    {
+      "metadata": {
+        "id": "onWOwLpYlzJQ",
+        "colab_type": "text"
+      },
+      "cell_type": "markdown",
+      "source": [
+        "Now instead of a dataset of `.jpeg` strings, this is a dataset of tensors.\n",
+        "\n",
+        "To serialize this to a TFRecord file you first convert the dataset of tensors to a dataset of strings."
+      ]
+    },
+    {
+      "metadata": {
+        "id": "xxZSwnRllyf0",
+        "colab_type": "code",
+        "colab": {}
+      },
+      "cell_type": "code",
+      "source": [
+        "ds = image_ds.map(tf.serialize_tensor)\n",
+        "ds"
+      ],
+      "execution_count": 0,
+      "outputs": []
+    },
+    {
+      "metadata": {
+        "id": "w9N6hJWAkKPC",
+        "colab_type": "code",
+        "colab": {}
+      },
+      "cell_type": "code",
+      "source": [
+        "tfrec = tf.data.experimental.TFRecordWriter('images.tfrec')\n",
+        "tfrec.write(ds)"
+      ],
+      "execution_count": 0,
+      "outputs": []
+    },
+    {
+      "metadata": {
+        "id": "OlFc9dJSmcx0",
+        "colab_type": "text"
+      },
+      "cell_type": "markdown",
+      "source": [
+        "With the preprocessing cached, data can be loaded from the TFrecord file quite efficiently. Just remember to de-serialized tensor before trying to use it."
+      ]
+    },
+    {
+      "metadata": {
+        "id": "BsqFyTBFmSCZ",
+        "colab_type": "code",
+        "colab": {}
+      },
+      "cell_type": "code",
+      "source": [
+        "RESTORE_TYPE = image_ds.output_types\n",
+        "RESTORE_SHAPE = image_ds.output_shapes\n",
+        "\n",
+        "ds = tf.data.TFRecordDataset('images.tfrec')\n",
+        "\n",
+        "def parse(x):\n",
+        "  result = tf.parse_tensor(x, out_type=RESTORE_TYPE)\n",
+        "  result = tf.reshape(result, RESTORE_SHAPE)\n",
+        "  return result\n",
+        "\n",
+        "ds = ds.map(parse, num_parallel_calls=AUTOTUNE)\n",
+        "ds"
+      ],
+      "execution_count": 0,
+      "outputs": []
+    },
+    {
+      "metadata": {
+        "id": "OPs_sLV9pQg5",
+        "colab_type": "text"
+      },
+      "cell_type": "markdown",
+      "source": [
+        "Now, add the labels and apply the same standard operations as before: "
+      ]
+    },
+    {
+      "metadata": {
+        "id": "XYxBwaLYnGop",
+        "colab_type": "code",
+        "colab": {}
+      },
+      "cell_type": "code",
+      "source": [
+        "ds = tf.data.Dataset.zip((ds, label_ds))\n",
+        "ds = ds.apply(\n",
+        "  tf.data.experimental.shuffle_and_repeat(buffer_size=image_count))\n",
+        "ds=ds.batch(BATCH_SIZE).prefetch(AUTOTUNE)\n",
+        "ds"
+      ],
+      "execution_count": 0,
+      "outputs": []
+    },
+    {
+      "metadata": {
+        "id": "W8X6RmGan1-P",
+        "colab_type": "code",
+        "colab": {}
+      },
+      "cell_type": "code",
+      "source": [
+        "timeit(ds)"
+      ],
+      "execution_count": 0,
+      "outputs": []
     }
   ]
 }
--- a/site/en/tutorials/load_data/images.ipynb
+++ b/site/en/tutorials/load_data/images.ipynb
@@ -454,13 +454,31 @@
       },
       "cell_type": "code",
       "source": [
-        "def load_and_format_image(path, label):\n",
+        "def format_image(path):\n",
         "  image = tf.read_file(path)\n",
         "  image = tf.image.decode_jpeg(image, channels=3)\n",
         "  image = tf.image.resize_images(image, [192, 192])\n",
         "  image = (image/128.0) - 1\n",
         "\n",
-        "  return image, tf.cast(label, dtype=tf.int64)"
+        "  return image"
+      ],
+      "execution_count": 0,
+      "outputs": []
+    },
+    {
+      "metadata": {
+        "id": "DP7X54sEIYgu",
+        "colab_type": "code",
+        "colab": {}
+      },
+      "cell_type": "code",
+      "source": [
+        "def process_image_label(path,label):\n",
+        "  image = tf.read_file(path)\n",
+        "  image = format_image(image)\n",
+        "\n",
+        "  label = tf.cast(label, dtype=tf.int64)\n",
+        "  return image, label"
       ],
       "execution_count": 0,
       "outputs": []
@@ -486,7 +504,7 @@
       "cell_type": "code",
       "source": [
         "# many threads will be waiting on disk reads.\n",
-        "image_ds = path_ds.map(load_and_format_image, num_parallel_calls=32)\n",
+        "image_ds = path_ds.map(process_image_label, num_parallel_calls=16)\n",
         "image_ds"
       ],
       "execution_count": 0,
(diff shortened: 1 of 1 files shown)


random-forests:
Thank you for writing this! Good timing, too. Together with Kiran's notebook on tf records, this would make a great start to a "Reading data" collection.

Suggestions:
=========

Under "Load and format the images"

* Instead of normalizing the image by "image = (image/128.0) - 1", instead, how about "image = image / 255"? 

* This would also simplify the code to display the images (instead of "plt.imshow((image+1)/2)", we can write "plt.imshow(image)")

* How about demo'ing the "load_and_format_image" function, just to demonstrate it works like regular Python code when eager is enabled? To do so, you could add a short block under it, that calls the method using something like "load_and_format_image('/root/.keras/datasets/flower_photos/tulips/7166644048_b00a14f01b.jpg', 0)" (I know there's a bunch of output, maybe catch it in a object, show how to get the .numpy() value, and print out the shape

Under "Pipe to a model for training"

* Explain the prefetch parameter. 

Under "Quick Transfer learning with keras.Applications"

* Leave the depth_multiplier at the default value.

* Finally, split the dataset into train / test, so we can show a quick evaluation after the transfer learning section, which is really cool. 

Overall this is great!

MarkDaoust:
> Instead of normalizing the image by "image = (image/128.0) - 1", instead, how about "image = image / 255"?

Done.
The `tf.keras.applications.mobilenet` expects data to be normalized to [-1,1].
I've switched it to use [0,1] for most of it, with a transform to [-1,1] in the transfer learning section. 

>demo'ing the "load_and_format_image"

Done.

>Explain the prefetch parameter.

Done

>Leave the depth_multiplier at the default value.

Done

>Finally, split the dataset into train / test, so we can show a quick evaluation after the transfer learning section, which is really cool.

I'd rather avoid going into too much detail here. I just want to get to "see how easy that was!" and leave the rest for a real tutorial.

MarkDaoust:
+tf.data

Hey Derek, Jiri, 

Does this seem like a reasonable approach to loading image directories with `tf.data`?

site/en/tutorials/load_data/images.ipynb
```diff
@@ -0,0 +1,1356 @@
...
+    },
+    {
+      "metadata": {
+        "id": "D1iba6f4khu-",
+        "colab_type": "code",
+        "colab": {}
+      },
+      "cell_type": "code",
+      "source": [
+        "image_ds = path_ds.map(load_and_preprocess_image, num_parallel_calls=8)"
```
jsimsa: replace `num_parallel_calls=8` with `num_parallel_calls=tf.data.experimental.AUTOTUNE`

site/en/tutorials/load_data/images.ipynb
```diff
@@ -0,0 +1,1356 @@
...
+      "cell_type": "code",
+      "source": [
+        "BATCH_SIZE = 32\n",
+        "\n",
+        "# Setting a shuffle buffer size larger than the dataset ensures that the data is completely shuffled.\n",
+        "ds = image_label_ds.repeat()\n",
+        "ds = ds.shuffle(buffer_size=4000) \n",
+        "ds = ds.batch(BATCH_SIZE)\n",
+        "# `prefecth` lets the dataset fetches batches, asynchronously while the model is training.\n",
+        "ds = ds.prefetch(1)\n",
```
jsimsa: replace `1` with `buffer_size=tf.data.experimental.AUTOTUNE`

site/en/tutorials/load_data/images.ipynb
```diff
@@ -0,0 +1,1356 @@
...
+      "metadata": {
+        "id": "SNOkHUGv3FYq",
+        "colab_type": "code",
+        "colab": {}
+      },
+      "cell_type": "code",
+      "source": [
+        "def change_range(image,label):\n",
+        "  image = 2*image-1\n",
+        "  return image, label\n",
```
jsimsa: this could just be `return 2*image - 1, label`

site/en/tutorials/load_data/images.ipynb
```diff
@@ -0,0 +1,1356 @@
...
+    {
+      "metadata": {
+        "id": "qwOJGxw_FkJU",
+        "colab_type": "code",
+        "colab": {}
+      },
+      "cell_type": "code",
+      "source": [
+        "ds = image_label_ds.repeat()\n",
+        "ds = ds.shuffle(buffer_size=4000) \n",
```
jsimsa: OOC, why did you pick 4000? as someone following the tutorial, I would want to know where this magic constant comes from. It might be worth saying that smaller values will result in faster startup (the buffer to initial fill is smaller) while higher values result in better shuffling (in the limit where buffer_size == input_size the shuffling is perfectly random).

Another thing to note is that because you `repeat` before `shuffle` some images might occur for the second time before other images occur at all. This can generally negatively affect accuracy. I suspect the reason you are doing this is to avoid filling up the shuffle buffer at the beginning of each epoch. To avoid doing that, while maintaining clean epoch boundaries, you can either 1) use the [tf.data.experimental.shuffle_and_repeat](https://www.tensorflow.org/api_docs/python/tf/data/experimental/shuffle_and_repeat) transformation or 2) do this:

```
ds = ds.shuffle(...).repeat().
options = tf.data.Options
options.experimental_shuffle_and_repeat_fusion = True
ds = ds.with_options(ds)
```

site/en/tutorials/load_data/images.ipynb
```diff
@@ -0,0 +1,1356 @@
...
+        "id": "UMVnoBcG_NlQ",
+        "colab_type": "text"
+      },
+      "cell_type": "markdown",
+      "source": [
+        "## Performance\n",
+        "\n",
+        "Note: This section just shows a couple of easy tricks that may help performance. For an in depth guide see [Input Pipeline Performance](https://www.tensorflow.org/guide/performance/datasets)\n",
+        "\n",
+        "The simple pipeline used above reads each file individually, on each epoch. This is fine for local training on CPU but may not be sufficient for GPU training, and is totally inapprpriate for any sort of distributed training. "
```
jsimsa: fix typo in `inapprpriate`

site/en/tutorials/load_data/images.ipynb
```diff
@@ -0,0 +1,1356 @@
...
+      ]
+    },
+    {
+      "metadata": {
+        "id": "lV1NOn2zE2lR",
+        "colab_type": "text"
+      },
+      "cell_type": "markdown",
+      "source": [
+        "Use `tf.data.Dataset.cache` to easily take advantage of the performance boost of working with in memory data.\n",
```
jsimsa: Make note of the fact that `cache` is only beneficial if the input dataset fits into memory. If it does not, then using `cache` does more harm then good.

jsimsa: nevermind, you already talk about this later

jsimsa:
Thanks @MarkDaoust did a pass and left some comments.

site/en/tutorials/load_data/images.ipynb
```diff
@@ -0,0 +1,1356 @@
...
+      "outputs": []
+    },
+    {
+      "metadata": {
+        "id": "uML4JeMmIAvO",
+        "colab_type": "text"
+      },
+      "cell_type": "markdown",
+      "source": [
+        "The `output_shapes` and `output_types` fields describe the content of each item in the dataset. In this case it is a set of scalar binary-strings"
```
mrry: We're going to remove these properties in V2. (They were rarely used by user code, and they aren't sufficient to express all of the things that can be in a dataset... especially the nested output of `Dataset.window()`.) The replacement will most likely be a single `Dataset.element_structure` property (exact name TBD) that serves the same purpose, and provides a superset of the same information.

site/en/tutorials/load_data/images.ipynb
```diff
@@ -0,0 +1,1356 @@
...
+      "outputs": []
+    },
+    {
+      "metadata": {
+        "id": "jHjgrEeTxyYz",
+        "colab_type": "text"
+      },
+      "cell_type": "markdown",
+      "source": [
+        "Since the datasets are in the same order we can just zip them together to get a dataset of `(image, label)` pairs."
```
mrry: Would it be worth mentioning that you can also call `from_tensor_slices()` with a pair of tensors directly?

site/en/tutorials/load_data/images.ipynb
```diff
@@ -0,0 +1,1356 @@
...
+      "outputs": []
+    },
+    {
+      "metadata": {
+        "id": "yA2F09SJLMuM",
+        "colab_type": "text"
+      },
+      "cell_type": "markdown",
+      "source": [
+        "The new dataset's `shapes` and `types`, are tuples of shapes and types as well, describing each field:"
```
mrry: Nit: no comma after `types`.

site/en/tutorials/load_data/images.ipynb
```diff
@@ -0,0 +1,1356 @@
...
+    {
+      "metadata": {
+        "id": "wwZavzgsIytz",
+        "colab_type": "text"
+      },
+      "cell_type": "markdown",
+      "source": [
+        "To train a model with this dataset you will want the data:\n",
+        "\n",
+        "* To be well shuffeled.\n",
```
mrry: s/shuffeled/shuffled/

site/en/tutorials/load_data/images.ipynb
```diff
@@ -0,0 +1,1356 @@
...
+        "* To be batched.\n",
+        "* To repeat forever.\n",
```
mrry: This part is TBD with the Keras API. I think we're leaning towards *not* repeating the dataset itself forever, but instead iterating over the same finite dataset for a given number of epochs inside Keras.

site/en/tutorials/load_data/images.ipynb
```diff
@@ -0,0 +1,1356 @@
...
+        "* Batches to available immediately when needed.\n",
```
mrry: "to be available as soon as possible"?

site/en/tutorials/load_data/images.ipynb
```diff
@@ -0,0 +1,1356 @@
...
+      },
+      "cell_type": "code",
+      "source": [
+        "BATCH_SIZE = 32\n",
+        "\n",
+        "# Setting a shuffle buffer size larger than the dataset ensures that the data is completely shuffled.\n",
+        "ds = image_label_ds.repeat()\n",
+        "ds = ds.shuffle(buffer_size=4000) \n",
+        "ds = ds.batch(BATCH_SIZE)\n",
+        "# `prefecth` lets the dataset fetches batches, asynchronously while the model is training.\n",
```
mrry: s/fetches batches, asynchronously/fetch batches asynchronously/

mrry: s/prefecth/prefetch/

mrry: Actually, for a novice user, I wonder if "in the background" would be more intuitive than "asynchronously"?

site/en/tutorials/load_data/images.ipynb
mrry: +1. I wonder if it would make sense to add this step inside Keras/Distribution Strategies as well?

site/en/tutorials/load_data/images.ipynb
```diff
@@ -0,0 +1,1356 @@
...
+      "outputs": []
+    },
+    {
+      "metadata": {
+        "id": "Y7NVWiLF3Vbf",
+        "colab_type": "text"
+      },
+      "cell_type": "markdown",
+      "source": [
+        "This model expects it's input to be normalized to the `[-1,1]` range:\n",
```
mrry: s/it's/its/

site/en/tutorials/load_data/images.ipynb
```diff
@@ -0,0 +1,1356 @@
...
+      ]
+    },
+    {
+      "metadata": {
+        "id": "CboYya6LmdQI",
+        "colab_type": "text"
+      },
+      "cell_type": "markdown",
+      "source": [
+        "So before the passing it to the mobilnet model, we need to convert the input from a range of `[0,1]` to `[-1,1]`."
```
mrry: s/mobilnet/MobileNet/ (or `mobile_net`)

site/en/tutorials/load_data/images.ipynb
```diff
@@ -0,0 +1,1356 @@
...
+    {
+      "metadata": {
+        "id": "OzAhGkEK6WuE",
+        "colab_type": "code",
+        "colab": {}
+      },
+      "cell_type": "code",
+      "source": [
+        "# The dataset may take a few seconds to start, as it fills its shuffle buffer.\n",
+        "image_batch, label_batch = next(iter(keras_ds))"
```
mrry: @jsimsa I forget: did we discuss adding a `Dataset.first()` method to make this a bit cleaner? (IIRC it was going to make some of the reducer examples a lot nicer.)

site/en/tutorials/load_data/images.ipynb
```diff
@@ -0,0 +1,1356 @@
...
+    },
+    {
+      "metadata": {
+        "id": "AnXPRNWoTypI",
+        "colab_type": "code",
+        "colab": {}
+      },
+      "cell_type": "code",
+      "source": [
+        "model.fit(ds, epochs=1, steps_per_epoch=3)  # steps_per_epoch=math.ceil(len(all_image_paths)/BATCH_SIZE)"
```
mrry: It doesn't exist yet, but I think the 2.0 recommendation for this will be `model.fit(ds.take(3), epochs=1)` (when the dataset input is finite).

site/en/tutorials/load_data/images.ipynb
```diff
@@ -0,0 +1,1356 @@
...
+      "outputs": []
+    },
+    {
+      "metadata": {
+        "id": "usIv7MqqZQps",
+        "colab_type": "text"
+      },
+      "cell_type": "markdown",
+      "source": [
+        "One disadvantage to using an im memory cache is that the cache must be rebuilt on each run, giving the same startup delay each time the dataset is started:"
```
mrry: I think @rohan100jain has some changes in flight that will help this no longer be the case.

site/en/tutorials/load_data/images.ipynb
```diff
@@ -0,0 +1,1356 @@
...
+      ]
+    },
+    {
+      "metadata": {
+        "id": "y1llOTwWFzmR",
+        "colab_type": "text"
+      },
+      "cell_type": "markdown",
+      "source": [
+        "TFRecord files are a simple format to store a sequence of binary blobs. In this case the fact that these are read linearly off disk, which may give some performance boost.\n",
```
mrry: The second sentence doesn't make much sense to me. Perhaps something like "By packing multiple examples into the same file, TensorFlow is able to read multiple examples at once, which is especially important for performance when using a remote storage service such as GCS."?

site/en/tutorials/load_data/images.ipynb
```diff
@@ -0,0 +1,1356 @@
...
+        "\n",
+        "It also gives you a portable data file that is apropriate, for example, for streaming over a network.\n",
```
mrry: s/apropriate/appropriate/

site/en/tutorials/load_data/images.ipynb
```diff
@@ -0,0 +1,1356 @@
...
+        "\n",
+        "First, build a `TFRecord` file from the raw image data:"
```
mrry: Nit: "TFRecord" isn't usually surrounded by backticks.

site/en/tutorials/load_data/images.ipynb
```diff
@@ -0,0 +1,1356 @@
...
+      "outputs": []
+    },
+    {
+      "metadata": {
+        "id": "wb7VyoKNOMms",
+        "colab_type": "text"
+      },
+      "cell_type": "markdown",
+      "source": [
+        "This is slower than the `cache` version because we have not cached the preprocessing."
```
mrry: This is kind of an anticlimax! Any chance of a "However, ..."?

MarkDaoust:
(not really merging, just testing copybara)

site/en/tutorials/load_data/images.ipynb
MarkDaoust: hmmm... the options approach doesn't seem to be working in tf-nightly. I'll send you a bug.

`shuffle_and_repeat` is.
Fixed.

jsimsa: The API for this has changed to:

```
ds = ds.shuffle(...).repeat().
options = tf.data.Options
options.experimental_optimization = tf.data.experimental.OptimizationOptions()
options.experimental_optimization.shuffle_and_repeat_fusion = True
ds = ds.with_options(ds)
```

site/en/tutorials/load_data/images.ipynb
MarkDaoust: I only did it this way because we had the images in a Dataset already.

I've added an example to the end of this section demonstrating:

```
ds = tf.data.Dataset.from_tensor_slices((all_image_paths, all_image_labels))

# The tuple elements are unpacked into the positional arguments of the mapped function 
def load_and_preprocess_from_path_label(path, label):
  return load_and_preprocess_image(path), label

image_label_ds = ds.map(load_and_preprocess_from_path_label)
image_label_ds
```

site/en/tutorials/load_data/images.ipynb
MarkDaoust: Acknowledged.

+1, that would be  a more intuitive API (I've already mentioned this to the keras team in b/109665994. Do you know the status of that plan?).

AFAIK, currently the only way it works is for the user to pass a `Dataset` that's long enough (typically using `.repeat()`) and use the `steps_per_epoch` argument to `fit`.

So I'll leave this for now, and gladly fix this when the behavior is changed.

site/en/tutorials/load_data/images.ipynb
MarkDaoust: Great. LMK when that happens and I can scrub `steps_per_epoch` from our docs.

site/en/tutorials/load_data/images.ipynb
MarkDaoust: !

How does that work?

site/en/tutorials/load_data/images.ipynb
MarkDaoust: SGTM

site/en/tutorials/load_data/images.ipynb
MarkDaoust: Sure.

I've split the TFRecord section into:

1. "Raw image data", where I dump the jpeg data directly to the TFRecord file.
2. "Serialized Tensors", where I decode and resize the images and `serialize_tensor` them into the TFRecord file.

PR #135, opened 2018-10-22
Use fewer epochs so this runs faster.

MarkDaoust:
Also switch to `model.save_weights`

Staging:  https://colab.sandbox.google.com/github/MarkDaoust/docs/blob/faster-RNN/site/en/tutorials/sequences/text_generation.ipynb


lamberta:
This diff isn't so good, anything on your end?

MarkDaoust:
> This diff isn't so good, anything on your end?

Rebased onto a commit where I saved the original version using colab.

So now it's only the first commit (which changes no content) that has a bad diff.

MarkDaoust:
> default to GPU in the Colab settings?

Yes. This tutorial benefits from it.

PR #152, opened 2018-10-24
Move XLA to subsite infra

lamberta:


lamberta:
Depends on cl/218410460

PR #155, opened 2018-10-26
Model Optimization Toolkit moved into TF Lite

lamberta:


lamberta:
Depends on cl/218810672

lamberta:
This is ready

PR #156, opened 2018-10-26
Pull docs tools from tensorflow/tensorflow.

MarkDaoust:
Import TensorFlow Docs tools, with history.
They don't work in this state, fix will follow shortly.

cd tensorflow
git subtree split --prefix tensorflow/tools/common/ -b tools_common
git subtree split --prefix tensorflow/tools/docs/ -b api_generator
cd ../docs/
git fetch ../tensorflow api_generator
git subtree add --prefix api_generator FETCH_HEAD
git fetch ../tensorflow tools_common
git subtree add --prefix tools_common FETCH_HEAD
cat tools_common/BUILD >> api_generator/BUILD
git rm tools_common/BUILD
git mv tools_common/* api_generator


MarkDaoust:
> Then they will need to sign a CLA as well, and confirm that they're okay with these being contributed to Google.

This is a straight move from another TensorFlow repo. All of the contributions are covered by the original submissions.

PR #160, opened 2018-10-29
Project import generated by Copybara.

MarkDaoust:
Sync from internal version, so this works outside of the main tensorflow repository.

PiperOrigin-RevId: 218592026


MarkDaoust:
> but were unable to find agreements for all the commit author(s) or Co-authors.

The" co-authors" is just the copybara exporter.

PR #161, opened 2018-10-29
Refactor books

lamberta:


lamberta:
Companion to cl/219158855

PR #162, opened 2018-10-29
Move performance guides under /guide section

lamberta:


lamberta:
Using @tfboyd's cutdown of the performance overview page from cl/198077351

PR #167, opened 2018-10-30
Add License, fix typos and punctuations in Image Classification tutorial

tanmay7270:
I added the Apache 2 license header. Also, while going through the tutorial, I found some typos so I fixed them along with some punctuations without modifying the original content. Could you please verify it?


lamberta:
Thanks, Tanmay! Very helpful.

@MarkDaoust This notebook was the one merged in #101 and is not ready to publish to tensorflow.org. Should it be marked with an underscore? This has the Francois MIT license I'm unclear if it came from him. Also, I think this belongs under the images section and not keras.

MarkDaoust:
Thanks man.

>Should it be marked with an underscore?
>I think this belongs under the images section and not keras.

yes. I'll do both in another PR

> This has the Francois MIT license I'm unclear if it came from him. 

He has a cats and dogs example .. but I don't see overlap between the two:

https://github.com/fchollet/deep-learning-with-python-notebooks/blob/master/5.3-using-a-pretrained-convnet.ipynb

I think we can remove that one.

MarkDaoust:
Thanks man.

>Should it be marked with an underscore?
>I think this belongs under the images section and not keras.

yes. I'll do both in another PR

> This has the Francois MIT license I'm unclear if it came from him. 

He has a cats and dogs example .. but I don't see overlap between the two:

https://github.com/fchollet/deep-learning-with-python-notebooks/blob/master/5.3-using-a-pretrained-convnet.ipynb

I think we can remove that one.

I'll make the edits.

MarkDaoust:
LGTM. Can one of you merge either:

https://github.com/tensorflow/docs/pull/168 or https://github.com/tanmay7270/docs/pull/1

PR #170, opened 2018-10-31
remove erroneous --config=opt flag

calid:
Using this flag results in the build failing with:

```
ERROR: Config value opt is not defined in any .rc file
```

and per https://github.com/tensorflow/probability/issues/176 this flag is not actually used.

### DIFF
--- a/site/en/guide/performance/overview.md
+++ b/site/en/guide/performance/overview.md
@@ -147,7 +147,7 @@ For TensorFlow source versions after 1.3.0:
 ```bash
 ./configure
 # Pick the desired options
-bazel build --config=mkl --config=opt //tensorflow/tools/pip_package:build_pip_package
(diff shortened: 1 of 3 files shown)


lamberta:
Thanks. You may need to rebase, the performance guide moved to: https://github.com/tensorflow/docs/blob/master/site/en/guide/performance/overview.md

@gunan, PTAL

calid:
updated, however after re-running configure it didn't complain about the flag, and bazel generated the necessary rc file... so looks like I misunderstood that other issue/PR, may have only been appropriate for the probability project.  I'll wait for confirmation but looking like this PR should be closed/not-merged.

calid:
@lamberta this was definitely a misunderstanding on my part, going to go ahead and close.  Sorry for the noise!

lamberta:
No worries, thanks for investigating!

PR #173, opened 2018-11-01
Move C++ guide to /guide/extend/cc

lamberta:
And remove the api_guides/cc directory.


lamberta:
Companion with: cl/219706680

PR #177, opened 2018-11-02
Update TF snippet to verify install

lamberta:
Fixes https://github.com/tensorflow/tensorflow/issues/23261


lamberta:
Stage: go/tfo-stage/install/pip

A little long, but we'll be able to remove the `tf.enable_eager_execution()` eventually

PR #180, opened 2018-11-05
Pull Style Transfer example

MarkDaoust:
Copy history from tensorflow/models to tensorflow/docs.

Images from: 
https://commons.wikimedia.org/wiki/File:Green_Sea_Turtle_grazing_seagrass.jpg
https://en.wikipedia.org/wiki/File:The_Great_Wave_off_Kanagawa.jpg


MarkDaoust:
These commits are just being transplanted between repos. Everything's Okay, googlebot.

PR #203, opened 2018-11-13
Update the TensorBoard section to write the graph to a file.

kevinbrooks-g:
The current instructions create a writer and add a graph but do not output the graph to a file. Adding the writer.flush() command will write the graph to a file.


kevinbrooks-g:
I redid the pull request to create a clean commit history.

PR #204, opened 2018-11-14
TFMA before TFS

rcrowe-google:
Clemens wanted TFMA before TFS, which I guess makes sense.


rcrowe-google:
So just waiting for merge

lamberta:
Merged. But you should be able to merge your own CL once approval is granted.

PR #217, opened 2018-11-20
Fix v1 loader shim

brilee:
Broke with the latest nightly due to updates in saved_model.export tags


lamberta:
Thanks, Brian.
While you're in here, can you add a note to the top of the notebook that says this notebook is work-in-progress? I'm concerned it might be taken out of context by someone driving by.

brilee:
@lamberta Added a note.

PR #221, opened 2018-11-26
Fix `Adding an op` links in the doc

Kh4L:
Some links were pointing to https://www.tensorflow.org/guide/extend/adding_an_op


Kh4L:
I signed it!

PR #231, opened 2018-12-04
Switched to use Keras optimizers in save_and_restore for tf v2 compatibility

tomerk:
Note: this is slightly suboptimal for tf v1 until we activate the v2 keras optimizers even in tf v1. Until then this warning gets printed:

```
        logging.warning(
            ('This model was compiled with a Keras optimizer (%s) but is being '
             'saved in TensorFlow format with `save_weights`. The model\'s '
             'weights will be saved, but unlike with TensorFlow optimizers in '
             'the TensorFlow format the optimizer\'s state will not be '
             'saved.\n\nConsider using a TensorFlow optimizer from `tf.train`.')
            % (optimizer,))
```


tomerk:
Done, thanks!

PR #236, opened 2018-12-06
change one-hot to multi-hot for consistency with overfit_and_underfit

rickiepark:
This particular `[3, 5]` sequence is used to explain multi-hot encoding in overfit_and_underfit notebook.
To prevent confusing, I want to change one-hot-encode to multi-hot-encode in basic_text_classification notebook.
Thanks. :)

### DIFF
--- a/site/en/tutorials/keras/basic_text_classification.ipynb
+++ b/site/en/tutorials/keras/basic_text_classification.ipynb
@@ -376,7 +376,7 @@
         "\n",
         "The reviews—the arrays of integers—must be converted to tensors before fed into the neural network. This conversion can be done a couple of ways:\n",
         "\n",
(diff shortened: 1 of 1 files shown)


lamberta:
It seems you are not alone based on this discussion of the book: https://forums.manning.com/posts/list/43280.page

lamberta:
Hi Haesun,
I have not been able to find the term "multi-hot encoding" used in the wild (beyond that page for the book), nor is it used in out [MLCC glossary](https://developers.google.com/machine-learning/glossary/). My preference is to use the wording from the book.
@MarkDaoust What do you think?

lamberta:
Chatted with Mark. Perhaps rewrite the sentence to something like:
Convert the arrays into vectors of 0s and 1s indicating ___, similar to a one-hot encoding.

This gets us to [indicator vector](https://en.wikipedia.org/wiki/Indicator_vector).

rickiepark:
Great.
How about "Convert the arrays into vectors of 0s and 1s indicating **word occurrence**, similar to a one-hot encoding"?
If you approve it, I will add a commit in this PR.
Thanks :)

lamberta:
That sounds good, thanks!

PR #237, opened 2018-12-06
Minor copy edits

colinmorris:


colinmorris:
@lamberta FYI, only users with write access to the repo are able to add or remove labels from issues, so a typical reader may be confused by that instruction.

lamberta:
@colinmorris Good point!

PR #238, opened 2018-12-08
Add object detection tutorial

polltooh:


lamberta:
Preview: https://colab.research.google.com/github/polltooh/docs/blob/master/site/en/tutorials/images/object_detection.ipynb

Hi Geoffrey,
This is a pretty neat example. Unfortunately, it does not run in Colab because it can't see your `object_detection_lib` module. We could either include all this code in the notebook (but minimize the cells because there's so much), or do something like a "curl and import". Or, since it's so large, add it to the examples repo and link off to it from /tutorials.
@MarkDaoust Can you take a look?

MarkDaoust:
Hi, 

While I do agree that we could use an object detection tutorial, I don't think this is quite how we want to do it. We would prefer to avoid pulling this much code into the repository. I'd be much more comfortable pointing to a full example somewhere else, since this is too much code to inline into the notebook.

Would it be possible to simplify this tutorial by, for example, using code from [tensorflow/models/object_detection](https://github.com/tensorflow/models/tree/master/research/object_detection)? 

Otherwise, maybe this would be a good candidate for the tensorflow/examples repository that @wolffg is working on? And then we could implement the tutorial here in terms of using that example.

I would also like to point people to the [hub object-detection example]( https://github.com/tensorflow/hub/blob/master/examples/colab/object_detection.ipynb) before they get down into the details like this. Because, if I understand correctly, this tutorial is "building an object detector" not just "using an object detector".

random-forests:
Working with Guanhang on this one. Agree it's still a ton o' code. He's been super helpful trimming it down (the problem seems to be there's a lot of utilities necessary for this domain, the model itself is tiny). The purpose of this example is to have end-to-end instructional code, written with modern APIs (in contrast to the OD API, which is heavier still).

Idea, how does this plan sound?  

1. We host this example on github.com/tensorflow/examples/object_detection, when that repo is up. 
2. Guanhang modifies his tutorial to be about "Object Detection in TensorFlow". In that tutorial (rather than having a bunch of runnable code), he can introduce key concepts in OD, then link to the various solutions we have, which look something like this:

OD API -> full industrial / extensive example
Hub -> easiest way to get started with a pre-trained model
Guanhang's -> easiest "from scratch" example.

So basically this can become a narrative document + links out to code.

MarkDaoust:
> 2\. Guanhang modifies his tutorial to be about "Object Detection in TensorFlow". In that tutorial (rather than having a bunch of runnable code), he can introduce key concepts in OD, then link to the various solutions we have

I like this plan. Especially if every part has a runnable Colab backing it. 

> The problem seems to be there's a lot of utilities necessary for this domain.

If they're used in many places maybe there's an argument for getting some into tf.Image.

### TF2

We also need to be careful here.. I see a bunch of code in the example that is clearly not tf2 compatible (name_scope, python_io). Similarly: is anyone converting models/od to tf2.

I don't think we're in a position to add anything new unless it's future-proofed for tf2.

```
!pip install -U tf-nightly-2.0-preview
```

lamberta:
Hi Geoffrey, you should now have access to the [examples repo](https://github.com/tensorflow/examples). Can you create this PR over there? I guess put it under `models/object_detection`, but that may change.
Thanks

polltooh:
Cool. Thanks for letting me know. Just submitted the [pull request](https://github.com/tensorflow/examples/pull/1/commits) to the examples repo. Please take a look. Thanks!!

tariq-hasan:
> Cool. Thanks for letting me know. Just submitted the [pull request](https://github.com/tensorflow/examples/pull/1/commits) to the examples repo. Please take a look. Thanks!!

The link appears to be dead!

polltooh:
> > Cool. Thanks for letting me know. Just submitted the [pull request](https://github.com/tensorflow/examples/pull/1/commits) to the examples repo. Please take a look. Thanks!!
> 
> The link appears to be dead!

This is a private github for now. Could you check to see if you have access to [examples](https://github.com/tensorflow/examples)

tariq-hasan:
> > > Cool. Thanks for letting me know. Just submitted the [pull request](https://github.com/tensorflow/examples/pull/1/commits) to the examples repo. Please take a look. Thanks!!
> > 
> > 
> > The link appears to be dead!
> 
> This is a private github for now. Could you check to see if you have access to [examples](https://github.com/tensorflow/examples)

I do not have access!

polltooh:
> > > > Cool. Thanks for letting me know. Just submitted the [pull request](https://github.com/tensorflow/examples/pull/1/commits) to the examples repo. Please take a look. Thanks!!
> > > 
> > > 
> > > The link appears to be dead!
> > 
> > 
> > This is a private github for now. Could you check to see if you have access to [examples](https://github.com/tensorflow/examples)
> 
> I do not have access!

This github would be made public soon. Stay tuned.

lamberta:
Tracking internally: b/135950856

PR #244, opened 2018-12-13
context fix

0101011:


mkaze:
@0101011 May I ask how did you modify the notebook? Since when I open a notebook of this repository on a local Jupyter server and modify it and wants to commit my changes, the diff generated by git covers all the lines (see [here][1]). Even though I have only say fixed a typo! Do I need to set any settings/configs?

[1]: https://github.com/mkaze/docs/commit/a031be5f4e9f24793e0c887616b756723ed611e9

0101011:
@mkaze Hi Masoud,

1. I use vanilla Jupyter for changes, no special configs

2. I can see the diff you've mentioned: do you face the same issue with other notebooks as well?

If it's the first time, try to replicate the problem with another ipynb file in your own repo to see if it's a thing

MarkDaoust:
Hi guys, 

The problem is most likely caused by Colab. We usually save our files using Colab, the online notebooking system. The output is formatted slightly differently from jupyter's.

For any notebook on github you can replace "github.com" with "colab.research.google.com/github" to open in Colab. (Or use [this chrome extension](https://chrome.google.com/webstore/detail/open-in-colab/iogfkhleblhcpcekbiedikdehleodpjo)) 

Colab's "save to github" option makes it easy. But you can also download and commit using the usual process.

For viewing the diffs, when it does make a mess... this app seems to work:

https://app.reviewnb.com/tensorflow/docs/pulls

0101011:
Got to try saving with Colab too :)

Thanks for the response Mark!

mkaze:
@0101011 @MarkDaoust Thank you for your replies. I used Colab, it is much better now but it still has some extra changes. Look at [this commit](https://github.com/mkaze/docs/commit/20dfd0cdefe9dfb5deddb961882702af709a08ec). It has changed the order of `"colab": {},` and `"cellView": "form"` for two of the cells. I did not touch those two cells at all and I have no idea why this happens?! I just opened the notebook from the original repo (i.e. TF docs) in the Colab and changed the content of a markdown cell and removed an extra cell and then used "Download .ipynb" from the File menu.

MarkDaoust:
Yes. This sort of thing happens to me sometimes too. I don't have a perfect solution.

0101011:
In the future all Jupyter files would be rewritten by Colab.

Just kidding, I'll give it a thorough look in 24 hours to find if the changes can be reverted painlessly with the proper diff generated, and hopefully get a workflow for that.

mkaze:
@MarkDaoust So is it OK to do a PR? No need to fix this manually?

MarkDaoust:
Yes, we'll still review and accept PRs where the diffs are (a bit) messy.

Don't worry about it too much. We're glad to have people who want to contribute, and we'd rather not waste your time worrying about little details like this.

PR #246, opened 2018-12-17
Update MNIST example

mdanatg:


mdanatg:
@brilee

mdanatg:
Note, all these changes are cosmetic - the notebook already runs at head.
Question about the final cell that is meant to load the saved model - should we leave it as-is, or remove altogether until loading is available?

MarkDaoust:
Thanks for the fixes Dan.

I prefer to name tutorials after what they demonstrate, instead of the dataset. 
Could we rename this to "tf2.ipynb" instead?

mdanatg:
How about something based on architecture, like "convolutional"? Alternatively it could be "training_and_saving" I think. Just "tf2" feels a bit too generic - all examples in this directory will be tf2 in the end. Once additional models are added, we'd need to differentiate them somehow.

lamberta:
Stage: https://colab.research.google.com/github/mdanatg/docs/blob/patch-3/site/en/2/guide/mnist.ipynb

Agree with both of you :) tf2 is too generic to fit into the guides, but I'm unclear of what this notebook is showing exactly.
Guides tend to be more focused and exhaustive of features. Would this be a more detailed saving guide under "HIgh Level APIs" [/guide section](https://www.tensorflow.org/guide/)? Will this have more detail than the [save and restore](https://www.tensorflow.org/guide/keras#save_and_restore) section of the Keras guide?

This kind of reads more like a [tutorial](https://www.tensorflow.org/tutorials/) (demonstrate an application of TensorFlow features), in that case, then maybe call it "convolutional" and find a tutorial group to put it under.

Or, is this guide meant to show what's required to migrate from TF1 to TF2? In that case, I might call it a tf2_migration_guide and add it to /guide.

mdanatg:
I agree that this directory is not ideal for a code sample (that's what the code was originally intended as). Given that, it would probably sit better in the model garden.

If you think it's worthwhile to keep it here and repurpose it as a "training and saving" tutorial that sounds good as well, but I wouldn't mind holding it off until we know for sure!

lamberta:
Hey Dan,
Another option: we've created an [examples](https://github.com/tensorflow/examples) repo. It's not yet public, but I've given you access if you'd like to submit this PR there (or we can still find a place here). @wolffg is cleaning up the models garden and moving some examples over.

mdanatg:
You have a better overview of both repositories - which alternative would be best to follow? Either sounds good to me.

PR #249, opened 2018-12-17
CUDA 10 install instructions.

tfboyd:
Should not be merged until 1.13.0 is RC at a minimum (so likely after the holiday). I do reference CUDA 10 being for >= 1.13.0 in the instructions but it is confusing. :-(


tfboyd:
Should not be merged until 1.13.0 is RC at a minimum (so likely after the holiday).  I do reference CUDA 10 being for >= 1.13.0 in the instructions but it is confusing.  :-(

PR #250, opened 2018-12-19
Fix basic classification tutorial

mkaze:
Issues fixed in this PR:
1. Casting is not done and there is no need to be done ==> removed redundant sentence.
2. There is no preprocessing function defined ==> removed redundant sentence.


MarkDaoust:
Thanks!

PR #254, opened 2018-12-21
Does {:.external} class in ipynb neccessary?

0101011:
Curious if we need {:.external} in the notebooks?
Reasoning: on tensorflow.org it renders as an external link symbol, but in the text it looks a bit messy.

Met those while doing keras translation.

### DIFF
--- a/site/en/guide/keras.ipynb
+++ b/site/en/guide/keras.ipynb
@@ -111,7 +111,7 @@
         "## Import tf.keras\n",
         "\n",
         "`tf.keras` is TensorFlow's implementation of the\n",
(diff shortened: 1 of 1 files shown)


MarkDaoust:
We like them, for tensorflow.org. But IIRC some infrastructure changes on our end will soon make these automatic. 

@lamberta  to confirm.

0101011:
Got it, then I won't touch them in the RU version

lamberta:
Mark is correct, though I'm fine taking them out now since they are ugly in the markdown and notebooks.
We should handle this on our side for tensorflow.org since I'm not sure if the proposed infrastructure fixes will help us anytime soon.

PR #255, opened 2018-12-21
added weights_only html reference tag

0101011:
Now the link works

### DIFF
--- a/site/en/guide/keras.ipynb
+++ b/site/en/guide/keras.ipynb
@@ -867,6 +867,7 @@
       },
       "cell_type": "markdown",
       "source": [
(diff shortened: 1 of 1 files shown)


MarkDaoust:
LGTM

Various viewers auto-add them for headings, but aren't consistent from platform to platform. So it's always good to explicitly add them where we're linking to a heading. Thanks.

0101011:
Hi Mark,

1. AFAIK there’s no way to auto-add them in vanilla Jupyter unless you use some IPython extensions

2.  Drop me some names, I’ll check those viewers - might be super handy!

PR #256, opened 2018-12-21
Add plt.show() to basic classification tuturial so something appears.

paylett:
When I was working though the tutorial:
https://www.tensorflow.org/tutorials/keras/basic_classification
the plot window did not appear for me.
Having not used matplotlib before, I spent quite a while trying to set up different configurations before discovering that the example simply needed to call plt.show()

### DIFF
--- a/site/en/tutorials/keras/basic_classification.ipynb
+++ b/site/en/tutorials/keras/basic_classification.ipynb
@@ -416,7 +416,8 @@
         "plt.figure()\n",
         "plt.imshow(train_images[0])\n",
         "plt.colorbar()\n",
-        "plt.grid(False)"
+        "plt.grid(False)\n",
+        "plt.show()"
       ],
       "execution_count": 0,
       "outputs": []
--- a/site/ko/tutorials/keras/basic_classification.ipynb
+++ b/site/ko/tutorials/keras/basic_classification.ipynb
@@ -402,7 +402,8 @@
     "plt.figure()\n",
     "plt.imshow(train_images[0])\n",
     "plt.colorbar()\n",
-    "plt.grid(False)"
+    "plt.grid(False)\n",
+    "plt.show()"
    ]
   },
   {
--- a/site/ru/tutorials/keras/basic_classification.ipynb
+++ b/site/ru/tutorials/keras/basic_classification.ipynb
@@ -404,7 +404,8 @@
     "plt.figure()\n",
     "plt.imshow(train_images[0])\n",
     "plt.colorbar()\n",
-    "plt.grid(False)"
+    "plt.grid(False)\n",
+    "plt.show()"
    ]
   },
   {


paylett:
Hi, I completed the Google Individual CLA prior to submitting the pull request. Is there some other step necessary to make this pass the CLA check?

0101011:
Hi @paylett,

1. plt.imshow does exactly that, it shows the image  😉

2. [Check the Colab outputs here](https://colab.research.google.com/github/tensorflow/docs/blob/master/site/en/tutorials/keras/basic_classification.ipynb) (Runtime > Run All)

3. You can perform even more operations on images using the arguments in matplotlib, [check more examples here](https://matplotlib.org/users/image_tutorial.html)

site/en/tutorials/keras/basic_classification.ipynb
```diff
@@ -416,7 +416,8 @@
         "plt.figure()\n",
         "plt.imshow(train_images[0])\n",
```
0101011: `plt.imshow = plt.show` in this case

lamberta:
Hello.
This has come up a couple of times and we decided against it: [here](https://github.com/tensorflow/docs/pull/49) and [here](https://github.com/tensorflow/models/pull/5166)

@MarkDaoust What do you think? Is there something we can do here to keep this from coming up?

0101011:
Couple of \n-s and a comment might be enough, looks clean in Colab:

```
plt.figure()

# Use imshow function to display the first training image
plt.imshow(train_images[0])

plt.colorbar()
plt.grid(False)
```

paylett:
Ah, I see now. The scenario/problem is as follows:

I have simply copy-pasted the code from the website into a .py file on my computer, and ran python from the command line.  (Python 3.6.3, Windows 10).  I need to call plt.show() to see anything. Whereas it doesn't seem to be necessary in Colab.

My audience profile is: experienced software developer; some python; no mathplotlib, Colab, Jupyter, Tensorflow; had an hour to kill and thought "hey, Tensorflow sounds cool, what can I learn". So I've reached this page in the first few minutes, saw the python and thought 'that looks fun - copy - paste'.

Now that you mention it Andrew, I see the 'Run in Colab' link at the top of the page - but I didn't see it at the time, and it's not mentioned inline in the tutorial. (Though I was just skimming looking at the code anyway).

Suggestion: how about having plt.show() inline, but commented out. It could also call out Colab. Simply seeing something like this would save people the confusion:
# plt.show()    # Insert if you are not using Colab

lamberta:
Thanks for the explanation, paylett@.

I agree, this sounds like a reasonable user journey: browse the webpage and copy/paste the code blocks as you go (assuming all the packages are installed). Commenting out `plt.show()` with an explanation could work, but if it doesn't hurt anything in Colab, maybe we should just include it. WDYT @MarkDaoust ? (will be back after the new year)

paylett:
> Try new things with Jupyter or Colab today

Absolutely - loving it already!  No worries, will leave it with you guys to decide if/how to change it.

0101011:
I've just tested this tutorial in a console, PyCharm and Colab.
Indeed, you can easily add `plt.show` without making a mess in the original guide:

1. In Colab, `plt.show()` doesn't change the output image we get with `plt.imshow()'
2. In console, `plt.imshow()` confirms there's an object, for example `<matplotlib.image.AxesImage object at 0x000001CAFFA6CB70>`

As in @paylett's PR:

```
plt.figure()
plt.imshow(train_images[0])
plt.colorbar()
plt.grid(False)
plt.show()
```

It should make things easier for newcomers.

0101011:
Also, pay attention that in this guide we're using `plt.imshow()` three times (2nd and 3rd being 484 and 764 accordingly).

jmhodges:
Considering how common it seems to be to use the website, and not the notebooks when you're a newbie, it seems like it would be nice to add at least some words in the prose about this line?

jmhodges:
If the `imshow` solves the problem, that's cool, too!

paylett:
Cool, thanks :)

PR #261, opened 2018-12-25
to update my tensorflow version?

darebooh:
Can I use this to update my tensorflow version?


lamberta:
Hello.
Closing since I don't think this is meant to be a pull request (if it is, you may need to rebase and submit only the changes).
For questions, you can try [StackOverflow](https://stackoverflow.com/questions/tagged/tensorflow).
Did you install TensorFlow [using pip](https://www.tensorflow.org/install/pip)? If so, you can update using: `pip install -U tensorflow`

PR #263, opened 2018-12-28
housekeeping with @rickiepark

0101011:
Random thoughts:

🔴 Got to find a way to bulk edit broken or outdated links across multiple folders
🔴 Or to automate somehow

### DIFF
--- a/site/ru/tutorials/keras/save_and_restore_models.ipynb
+++ b/site/ru/tutorials/keras/save_and_restore_models.ipynb
@@ -116,7 +116,7 @@
     "\n",
     "### Варианты\n",
     "\n",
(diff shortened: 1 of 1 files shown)


lamberta:
Yes, that'd be nice.

For API reference, if you wrap the symbol in an inline code block (eg \`tf.data.Dataset\`), it will automatically get converted to a link when tensorflow.org is generated.

For anchor links, it's a bit confusing because GitHub automatically uses '-' for spaces where our site uses '_'. During import we automatically convert the links, but it's still confusing. We have some linters that run on internal changes, but it we'd like to externalize some of this tooling (and we're working on it).

PR #266, opened 2018-12-30
colab: rendering all sections in the table of contents

0101011:
Name of the chapter and its sections should be in separated markdown cells in order to render each section in the table of contents in Colab.

Not necessarily a bug, but something to remember.

fix

### DIFF
--- a/site/en/guide/keras.ipynb
+++ b/site/en/guide/keras.ipynb
@@ -862,14 +862,22 @@
     },
     {
       "metadata": {
-        "id": "qnl7K-aI0E2z",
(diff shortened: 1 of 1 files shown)


0101011:
These could be found in some notebooks, especially in those converted from Jupyter.

I think it should be added as a rule for the tutorial style guide for other notebook authors to follow.

lamberta:
> I think it should be added as a rule for the tutorial style guide for other notebook authors to follow.

Agree! We're working on a Colab notebook style guide for tensorflow.org and will hopefully push an initial draft soon.

PR #269, opened 2019-01-01
KO translation - tutorials/keras/overfit_and_underfit, save_and_restore_models

rickiepark:
Running with colab is OK.
Thanks!


rickiepark:
Thanks. I’ll do my best! :-)

PR #273, opened 2019-01-03
tf.DType api external doc link

0101011:


lamberta:
Thanks, but this isn't necessary. When the docs are imported the backtick style for reference symbols ( \`tf.symbol\`) is automatically converted to a reference link. You can see this one here: https://www.tensorflow.org/guide/tensors#data_types

But, we should update the [style guide](https://www.tensorflow.org/community/documentation#links_in_markdown) to note this.

PR #275, opened 2019-01-04
Remove enable eager from the text

hgadig:
Since TF 2.0 does not require enabling eager, this sentence can be removed.
"and enable eager execution for this program"


hgadig:
I signed the CLA. But it still shows as not signed.

PR #277, opened 2019-01-04
fixed the broken links

rthadur:
fixed the broken links

### DIFF
--- a/site/en/2/tutorials/keras/index.md
+++ b/site/en/2/tutorials/keras/index.md
@@ -4,7 +4,7 @@ This notebook collection is inspired by the book
 *[Deep Learning with Python](https://books.google.com/books?id=Yo3CAQAACAAJ)*.
 These tutorials use `tf.keras`, TensorFlow's high-level Python API for building
 and training deep learning models. To learn more about using Keras with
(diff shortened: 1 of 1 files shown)


lamberta:
Thanks! Can you please sign the CLA?

site/en/2/tutorials/keras/index.md
```diff
@@ -4,7 +4,7 @@ This notebook collection is inspired by the book
 *[Deep Learning with Python](https://books.google.com/books?id=Yo3CAQAACAAJ)*.
 These tutorials use `tf.keras`, TensorFlow's high-level Python API for building
 and training deep learning models. To learn more about using Keras with
-TensorFlow, see the [TensorFlow Keras Guide](../../guide/keras).
+TensorFlow, see the [TensorFlow Keras Guide](../../guide/keras.ipynb).
```
lamberta: Note: Since we're in the /2 directory this should be "../../..", but I think the idea is to make these the default so leave as-is for now.

MarkDaoust: >leave as-is for now.

+1 any links crossing in or out of the 2/ directory will broken in the future. So it's best not to add them in the first place.

PR #281, opened 2019-01-04
typo fixes removed "and enable eager execution"

rthadur:
since tf 2.0 does not require enable eager execution


rthadur:
@lamberta i signed the CLA but it is not reflecting here , is there any way to resolve this ?

lamberta:
Thanks!

PR #285, opened 2019-01-05
Remove hardcoded 'start_string' in text generation tutorial

Andy-Wilkinson:
When following the 'Text generation using a RNN with eager execution' tutorial the 'generate_text' function takes a start_string as an argument, however, this is always overwritten with a hard-coded value. The hardcoded value has been removed.


Andy-Wilkinson:
I signed it!

PR #287, opened 2019-01-08
fix links for tf2 guides and tutorials

VikramTiwari:


MarkDaoust:
My comment on #277 was meant to be more specific than this.

MarkDaoust:
My comment on #277 was meat to be more specific than this.
What I meant was we shouldn't use relative links from inside the 2/ directory to outside of it, or vice-versa.

I see no downside to fixing all these buttons.

MarkDaoust:
In fact, I should add a test to ensure these buttons are correct.

PR #293, opened 2019-01-11
Fix heading issue in markdown

amitness:
Due to updates in the GitHub markdown processor, the markdown for H2 headings is only rendered when there is space between `##` and the `title`. I have updated the markdown with space so that the header is rendered properly.

Previously:
image

Current:
image


amitness:
I signed it!

PR #294, opened 2019-01-11
Code refactor and updat doc

AakashKumarNain:
The code in the notebook was a little messy. Refactored the code to make it more modular and readable. Also, the text cells in the notebook have been updated to include more clear documentation.

### DIFF
--- a/site/en/r2/tutorials/keras/overfit_and_underfit.ipynb
+++ b/site/en/r2/tutorials/keras/overfit_and_underfit.ipynb
@@ -1,25 +1,42 @@
 {
+  "nbformat": 4,
+  "nbformat_minor": 0,
+  "metadata": {
+    "colab": {
+      "name": "overfit-and-underfit.ipynb",
+      "version": "0.3.2",
+      "provenance": [],
+      "private_outputs": true,
+      "collapsed_sections": [
+        "fTFj8ft5dlbS"
+      ],
+      "toc_visible": true
+    },
+    "kernelspec": {
+      "display_name": "Python 3",
+      "name": "python3"
+    },
+    "accelerator": "GPU"
+  },
   "cells": [
     {
-      "cell_type": "markdown",
       "metadata": {
         "colab_type": "text",
         "id": "fTFj8ft5dlbS"
       },
+      "cell_type": "markdown",
       "source": [
         "##### Copyright 2018 The TensorFlow Authors."
       ]
     },
     {
-      "cell_type": "code",
-      "execution_count": 0,
       "metadata": {
         "cellView": "form",
-        "colab": {},
         "colab_type": "code",
-        "id": "lzyBOpYMdp3F"
+        "id": "lzyBOpYMdp3F",
+        "colab": {}
       },
-      "outputs": [],
+      "cell_type": "code",
       "source": [
         "#@title Licensed under the Apache License, Version 2.0 (the \"License\");\n",
         "# you may not use this file except in compliance with the License.\n",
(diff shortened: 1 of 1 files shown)


site/en/r2/tutorials/keras/overfit_and_underfit.ipynb
```diff
@@ -461,193 +572,263 @@
...
     {
-      "cell_type": "markdown",
       "metadata": {
         "colab_type": "text",
         "id": "ASdv7nsgEFhx"
       },
+      "cell_type": "markdown",
       "source": [
-        "## Strategies"
+        "## Strategies to fight overfitting "
```
0101011: "## Strategies to prevent overfitting "

AakashKumarNain:
> Thanks for the submission, view here: https://colab.research.google.com/github/AakashKumarNain/docs/blob/master/site/en/r2/tutorials/keras/overfit_and_underfit.ipynb
> 
> This collection of tf.keras notebooks is based on [the notebooks](https://github.com/fchollet/deep-learning-with-python-notebooks) from the book _Deep Learning with Python_—updated and expanded where necessary. Because this is introductory material, while modularity is nice, I'd prefer to be explicit for these guides.
> 
> @MarkDaoust Can you take a look?

@lamberta Thank you for the response. I have read that book. And I agree that the guide should be implicit at this point, this is why I did't create all the models from a single function call. Keras API is lovely. It is the cleanest API to the date. Hence I only made the necessary changes to make sure that it is on the lines of Keras

site/en/r2/tutorials/keras/overfit_and_underfit.ipynb
```diff
@@ -222,225 +267,291 @@
...
-        "baseline_model = keras.Sequential([\n",
-        "    # `input_shape` is only required here so that `.summary` works. \n",
-        "    keras.layers.Dense(16, activation=tf.nn.relu, input_shape=(NUM_WORDS,)),\n",
-        "    keras.layers.Dense(16, activation=tf.nn.relu),\n",
-        "    keras.layers.Dense(1, activation=tf.nn.sigmoid)\n",
-        "])\n",
-        "\n",
-        "baseline_model.compile(optimizer='adam',\n",
+        "def create_baseline_model():\n",
+        "  baseline_model = Sequential()\n",
```
yashk2810: Hey,

.add() just adds more code when its not necessary. I think we should stick to the original version of creating the models here and everywhere below.

site/en/r2/tutorials/keras/overfit_and_underfit.ipynb
```diff
@@ -461,193 +572,263 @@
...
+        "1.3, 0, 1.1]`. The \"dropout rate\" is the fraction of the features that are being zeroed-out; it is usually set between 0.2 and 0.5. At test time, no units are dropped out, and instead the layer's output values are scaled down by a factor equal to the dropout rate, so as to balance for the fact that more units are active than at training time.\n",
+        "\n",
+        "In pure numpy code, it can be implemented like this:\n",
+        "```\n",
+        "a = np.random.rand(5,1)\n",
+        "drop_rate = 0.2\n",
+        "mask = np.random.rand(*a.shape) > 0.2\n",
+        "a_with_drop = a*mask\n",
+        "a_with_drop = (1/(1-drop_rate))*a_with_drop\n",
```
yashk2810: Maybe you could link to the implementation in tf.keras?

site/en/r2/tutorials/keras/overfit_and_underfit.ipynb
AakashKumarNain: Hey,

Thanks for reviewing it. It certainly adds few words to the code but then there is a difference between `ugly` code and `good` code. After using `Keras` for a long time, I think this is the actual way all the people(especially everyone I know on `Kaggle`) use it. Regarding the `dropout`, I think I would remove it and restate it in the original state. Showing the internals wasn't necessary for the notebook, I certainly overthought it.

site/en/r2/tutorials/keras/overfit_and_underfit.ipynb
AakashKumarNain: Thanks. I will change that

lamberta:
@AakashKumarNain I was thinking about this on the commute home, if you have time at some point, could you look over the more advanced notebooks? The beginner-focused notebooks get reviewed a lot, but it's much more difficult to find contributors who can explain beyond-the-basics.

We're currently linking off to a bunch of work-in-progress notebooks on https://www.tensorflow.org/tutorials/ ---click through the "Generative models" section, "Images", etc. We're starting to move these over to our [TF 2.0 notebooks](https://github.com/tensorflow/docs/tree/master/site/en/r2/tutorials), but there is lots to be done there. Let me know if any of these seem interesting to you.

Also, consider joining the new [docs@tensorflow.org list](https://groups.google.com/a/tensorflow.org/forum/#!forum/docs).

AakashKumarNain:
Makes sense totally. I will look into it over the weekend. Thanks
PS: I have already joined that list

PR #296, opened 2019-01-11
Estimators: comment fix + better name proposed

0101011:
1. `my_training_set` is not a function (see line 115).

2. Propose changing ‘input_fn’ to `data_input` in the whole block of text as a more explicit function name and to improve readability.

Input function traditionally expects some input from a user whereas in our context data loading and manipulation might be more relevant.

### DIFF
--- a/site/en/guide/estimators.md
+++ b/site/en/guide/estimators.md
@@ -82,7 +82,7 @@ of the following four steps:
     For example, the following code illustrates the basic skeleton for
     an input function:
 
-        def data_input(dataset):
+        def input_fn(dataset):
            ...  # manipulate dataset, extracting the feature dict and the label
            return feature_dict, label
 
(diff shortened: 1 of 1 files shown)


site/en/guide/estimators.md
```diff
@@ -81,7 +82,7 @@ of the following four steps:
     For example, the following code illustrates the basic skeleton for
     an input function:
 
-        def input_fn(dataset):
+        def data_input(dataset):
```
lamberta: `input_fn` is used all over the Estimator docs so I'd resist changing here.

site/en/guide/estimators.md
```diff
@@ -112,8 +113,8 @@ of the following four steps:
 4.  **Call a training, evaluation, or inference method.**
     For example, all Estimators provide a `train` method, which trains a model.
 
-        # my_training_set is the function created in Step 1
-        estimator.train(input_fn=my_training_set, steps=2000)
+        # `data_input` is the function created in Step 1
+        estimator.train(data_input=my_training_set, steps=2000)
```
lamberta: `input_fn` is the param name for `.train`: https://github.com/tensorflow/tensorflow/blob/r1.12/tensorflow/python/estimator/estimator.py#L280

0101011:
Exactly, I found out 2 minutes later that the naming convention already established in the codebase, change would be unreasonable.

Rolling those back.

PR #304, opened 2019-01-20
Remove unnecessary variables in text classification tutorial

gud2great:
1. _history_dict_ variable was already declared, so there is no reason to use _history.history_ again.
2. _acc_values_ and _val_acc_values_ are not used because they are equal to _acc_ and _val_acc_.


yashk2810:
LGTM

PR #305, opened 2019-01-21
Fix : Ko_Basic_regression

namepen:
In Ko basic regression, 'plot_history' funtion is old version than en_basic_regression.
And add sublots it looks great.

I add some line in plot_history function like that:

```
def plot_history(history):
  hist = pd.DataFrame(history.history)
  hist['epoch'] = history.epoch
  
  plt.figure(figsize=(8,12))
  
  plt.subplot(2,1,1)
  plt.xlabel('Epoch')
  plt.ylabel('Mean Abs Error [MPG]')
  plt.plot(hist['epoch'], hist['mean_absolute_error'],
           label='Train Error')
  plt.plot(hist['epoch'], hist['val_mean_absolute_error'],
           label = 'Val Error')
  plt.ylim([0,5])
  plt.legend()
  
  
  plt.subplot(2,1,2)
  plt.xlabel('Epoch')
  plt.ylabel('Mean Square Error [$MPG^2$]')
  plt.plot(hist['epoch'], hist['mean_squared_error'],
           label='Train Error')
  plt.plot(hist['epoch'], hist['val_mean_squared_error'],
           label = 'Val Error')
  plt.ylim([0,20])
  plt.legend()
  
  plt.show()
```


lamberta:
Thanks

Staged: https://colab.research.google.com/github/namepen/docs/blob/Fix-ko_basic_regression/site/ko/tutorials/keras/basic_regression.ipynb

It looks like this code differs a bit from the [English version](https://github.com/tensorflow/docs/blob/master/site/en/tutorials/keras/basic_regression.ipynb), yes?

If that's the case, please update this PR to include that change. We don't want to be in a situation where we're maintaining separate notebook code for each language. English should be the source of truth for these tutorials and the other languages should follow.

namepen:
lamberta 

Sorry. I explain in detail.

First, There is difference in 'plot_history' function.
think en_basic_regression is latest version, the below line is not included in ko_basic_regression.
(It's also [Tensorflow.org/TUTORIALS](https://www.tensorflow.org/tutorials/keras/basic_regression?hl=ko)  and RU_basic_regression.)

```
  hist = pd.DataFrame(history.history)
  hist['epoch'] = history.epoch
```
So. I add this line in ko_basic_regession. 

Second, I think that adding a subplots in plot_history looks good. 
added line is that '#Add':
```
def plot_history(history):
  hist = pd.DataFrame(history.history)
  hist['epoch'] = history.epoch
  
  plt.figure(figsize=(8,12))          #Add
  
  plt.subplot(2,1,1)                   #Add
  plt.xlabel('Epoch')
  plt.ylabel('Mean Abs Error [MPG]')
  plt.plot(hist['epoch'], hist['mean_absolute_error'],
           label='Train Error')
  plt.plot(hist['epoch'], hist['val_mean_absolute_error'],
           label = 'Val Error')
  plt.ylim([0,5])
  plt.legend()
  
  
  plt.subplot(2,1,2)                   #Add
  plt.xlabel('Epoch')
  plt.ylabel('Mean Square Error [$MPG^2$]')
  plt.plot(hist['epoch'], hist['mean_squared_error'],
           label='Train Error')
  plt.plot(hist['epoch'], hist['val_mean_squared_error'],
           label = 'Val Error')
  plt.ylim([0,20])
  plt.legend()
  
  plt.show()
```
org plot image is:
<div>
</img>
</div>

and adding subplot image is:
<div>
</img>
</div>


If you like it, I also edit 'en_basic_regression'.
otherwise I fix this PR about first thing including RU_basic_regression.

comment me. thanks

lamberta:
Thanks for the explanation. It looks good to me but I'll let @MarkDaoust or @yashk2810 take a look.

One of the challenges of community translations is keeping things in sync. From my perspective, en/ is the source of truth and the translation directories flow from that. Though how fast that happens is totally up to the community. If there's a code change here, then it needs to get "upstreamed" to the en/ counterpart for it to stick. So if this change looks good, then we should make the same change in en/ (and ru/ if possible).

yashk2810:
Adding subplots is better, I agree. Can you please make changes everywhere as suggested by @lamberta

namepen:
Thanks. I change en, ko and ru regression.

namepen:
@lamberta 
Sorry. I pulled the file in the desktop after worked it in colab. it seems to have a conflict issue in this. 

Thanks for reformatted.

PR #306, opened 2019-01-22
Fix keras optimizer warning

adammichaelwood:
closes b/122067269

Changes Keras optimizer to TF optimizer to avoid warnings,
and adds explanation of same.

https://colab.research.google.com/github/adammichaelwood/docs/blob/fix-keras-optimizer-warning/site/en/tutorials/keras/save_and_restore_models.ipynb


lamberta:
Editorial is ok.
@wolffg @MarkDaoust @yashk2810 for review

Here's the warning:
> WARNING:tensorflow:This model was compiled with a Keras optimizer (<tensorflow.python.keras.optimizers.Adam object at 0x7fbd5eb232e8>) but is being saved in TensorFlow format with `save_weights`. The model's weights will be saved, but unlike with TensorFlow optimizers in the TensorFlow format the optimizer's state will not be saved.

It seems like we should point users forward with`tf.keras.optimizers.Adam` and maybe the warnings are a feature here. ALternatively, is there something we can do on the model saving side so it doesn't complain?

lamberta:
Hey @yashk2810, can you research this? `tf.train.AdamOptimizer` vs `tf.keras.optimizers.Adam` vs `'adam'`, which is preferred for the examples?

yashk2810:
I would prefer using tf.keras.optimizers.Adam() wherever possible. If the loss used is not string, then the optimizer shouldn't be too. If both of them are string, then I am fine with that too.

adammichaelwood:
fixed
I think this is good now.

lamberta:
Now this example uses both `tf.train.AdamOptimizer` and `tf.keras.optimizers.Adam`. Is that the intent? Seems confusing.

Switching `.compile` to `tf.keras.optimizers.Adam` is what prints the warning referred to in the bug:

> WARNING:tensorflow:This model was compiled with a Keras optimizer (<tensorflow.python.keras.optimizers.Adam object at 0x7fe470480710>) but is being saved in TensorFlow format with `save_weights`. The model's weights will be saved, but unlike with TensorFlow optimizers in the TensorFlow format the optimizer's state will not be saved.

So it seems there are a couple of options:
1. Use a mix of `train.AdamOptimizer` and `keras.optimizers.Adam` to avoid the warning (but not provide direction to the user why you should use one instead of another).
2. Switch all to `tf.keras.optimizers.Adam` and accept this warning (will it eventually go away?) and mark the bug WAI. (This is also the behavior when using `'adam'`)
3. Use `train.AdamOptimizer` to get rid of the warnings (as in Adam's initial change), knowing this is not desired and will eventually have to change it to keras.

Thoughts?

adammichaelwood:
The mixed-use approach was directed by @wolffg

lamberta:
We could use some guidance from product on this. Looping in @fchollet and @karmel

karmel:
IIUC, this should not be a problem with the new optimizers, which are in the nightlies now, and will go out with the next release. Can we tolerate the warning for now, rather than switch here and switch back after? @tanzhenyu just in case I am incorrect.

adammichaelwood:
I can put a note in about the warnings, so at least they don't make people think something has gone wrong.

tanzhenyu:
@adammichaelwood sounds good to me!

lamberta:
Thanks @karmel 
Then we should go with 2. Switch all to `tf.keras.optimizers.Adam` and accept this warning which it should go away in the next release. I don't think it's worth adding a note about the warning since we'll have to just update it again soon. If you do, then please create a tracking bug so we don't forget about it.

site/en/tutorials/keras/save_and_restore_models.ipynb
```diff
@@ -258,7 +258,7 @@
         "    keras.layers.Dense(10, activation=tf.nn.softmax)\n",
         "  ])\n",
         "  \n",
-        "  model.compile(optimizer='adam', \n",
+        "  model.compile(optimizer=tf.train.AdamOptimizer(),\n",
```
yashk2810: tf.keras.optimizers.Adam()

site/en/tutorials/keras/save_and_restore_models.ipynb
```diff
@@ -810,11 +813,14 @@
       },
       "cell_type": "code",
       "source": [
-        "# The optimizer was not restored, re-attach a new one.\n",
-        "new_model.compile(optimizer='adam', \n",
+        "# The model has to be compiled with with a loss function before evaluating.\n",
+        "# This step is not required if the model is being deployed \n",
+        "# without further training or evaluation.\n",
+        "new_model.compile(optimizer=tf.train.AdamOptimizer(), \n",
```
yashk2810: tf.keras.optimizers.Adam()

lamberta:
Thanks, Adam!

PR #308, opened 2019-01-23
Convert Estimators CNN example to notebook

lamberta:
This is a straight conversion of the [Estimators CNN tutorial](https://www.tensorflow.org/tutorials/estimators/cnn) to a notebook, with the following changes:

* Changed training steps from 20,000 to 1,000. Results in significantly decreased accuracy but runs in a reasonable amount of time in Colab (as noted in the prior example, as written, this notebook takes about an hour to train). I've edited the text to reflect this.
* Removed the final "Run the model" section since that's evident in Colab.

Stage: https://colab.research.google.com/github/lamberta/docs/blob/cnn-estimator-notebook/site/en/tutorials/estimators/cnn.ipynb


yashk2810:
Instead of the split parameter in tfds, this of loading both the train and test is convenient.

```
datasets, ds_info = tfds.load(name="mnist", with_info=True, as_supervised=True)
mnist_train, mnist_test = datasets['train'], datasets['test']
```

site/en/tutorials/estimators/cnn.ipynb
```diff
@@ -0,0 +1,1014 @@
...
+      "source": [
+        "# Train the model\n",
+        "train_input_fn = tf.estimator.inputs.numpy_input_fn(\n",
+        "    x={\"x\": train_data},\n",
+        "    y=train_labels,\n",
+        "    batch_size=100,\n",
+        "    num_epochs=None,\n",
+        "    shuffle=True)\n",
+        "\n",
+        "mnist_classifier.train(\n",
```
MarkDaoust: Can we just demonstrate the logging_hook with steps=1, then turn it off for the main training? 

```
mnist_classifier.train(
    input_fn=train_input_fn,
    steps=1,
    hooks=[logging_hook])
```

```
mnist_classifier.train(
    input_fn=train_input_fn,
    steps=1000)
```

That way we avoid thousands of lines of output in the rendered page.

site/en/tutorials/estimators/cnn.ipynb
```diff
@@ -0,0 +1,1014 @@
...
+    {
+      "metadata": {
+        "id": "ccobb0qETV-S",
+        "colab_type": "code",
+        "colab": {}
+      },
+      "cell_type": "code",
+      "source": [
+        "# Load training and eval data\n",
+        "mnist = tf.contrib.learn.datasets.load_dataset(\"mnist\")\n",
```
MarkDaoust:
```
# Load training and eval data
((train_data, train_labels),
 (eval_data, eval_labels)) = tf.keras.datasets.mnist.load_data()

train_data = (train_data/255.0).astype(np.float32)
train_labels = train_labels.astype(np.int32)

eval_data = (eval_data/255.0).astype(np.float32)
eval_labels = eval_labels.astype(np.int32)
```

Gets us numpy arrays, without contrib.
That way there's no work to do rewording the text.

site/en/tutorials/estimators/cnn.ipynb
lamberta: Done. Moved second training run to separate code cell to improve output. Also reworked the text around this.

site/en/tutorials/estimators/cnn.ipynb
lamberta: Done, thanks

yashk2810: I think we should convert to float32 first and then divide. The numpy array is float64 which is a costlier division than float32. Not that it matters here but its a good practice.

lamberta: Update to Mark's snippet ...
@yashk2810 Can you add a patch to this PR or paste a snippet for your suggestion?

yashk2810: # Load training and eval data
((train_data, train_labels),
 (eval_data, eval_labels)) = tf.keras.datasets.mnist.load_data()

train_data = train_data/np.float32(255)

# not required
train_labels = train_labels.astype(np.int32)

eval_data = eval_data/np.float32(255)

# not required
eval_labels = eval_labels.astype(np.int32)

lamberta:
Updated to @yashk2810 's suggestion of converting to float32 first. PTAL

yashk2810:
Is it possible to use keras layers?

lamberta:
Possibly, but they don't look like a straight drop in replacement. That's beyond the scope I've my port to notebook and dataset fix, but you are welcome to get it to work :)

yashk2810:
Sounds good.

PR #311, opened 2019-01-24
Fix load images notebook to render for site

lamberta:


lamberta:
stage: https://colab.research.google.com/github/lamberta/docs/blob/fix-images-nb/site/en/tutorials/load_data/images.ipynb

lamberta:
Thanks. Removed backticks in title causing render issue on tf.org

PR #312, opened 2019-01-24
minor edits on keras guide

adammichaelwood:
Just a couple tiny text edits.


site/en/r2/guide/keras.ipynb
```diff
@@ -1203,18 +1219,16 @@
         "The following example distributes a `tf.keras.Model` across multiple GPUs on a\n",
```
MarkDaoust: Could use a link to guide/distribute_strategy.ipynb right here "See the full guide"

adammichaelwood: added

PR #318, opened 2019-02-01
Resolves Attribute error on colab

hamk3010:
Using **tf.random.categorical** gives the following error:

`AttributeError: module 'tensorflow._api.v1.random' has no attribute 'categorical'`

Changing it to **tf.random. multinomial** with the same parameters seems to resolve the issue and works on Colab notebooks.

Here is the [solution](https://stackoverflow.com/questions/54210128/tensorflow-text-generation).

### DIFF
--- a/site/en/tutorials/sequences/text_generation.ipynb
+++ b/site/en/tutorials/sequences/text_generation.ipynb
@@ -745,7 +745,7 @@
       },
       "cell_type": "code",
       "source": [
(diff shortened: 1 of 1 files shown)


lamberta:
Thanks, Hamza!
Please sign the CLA when you get a chance.

Stage: https://colab.research.google.com/github/hamk3010/docs/blob/patch-1/site/en/tutorials/sequences/text_generation.ipynb

PR #321, opened 2019-02-03
Fix minor inaccuracy in russian translation.

catinwarmhands:
"Test dataset" ("проверочный датасет") are sometimes called "training dataset" ("тренировочный датасет"), which can lead to confusion among new users.

### DIFF
--- a/site/en/tutorials/keras/basic_classification.ipynb
+++ b/site/en/tutorials/keras/basic_classification.ipynb
@@ -1,989 +1,1040 @@
 {
-  "nbformat": 4,
-  "nbformat_minor": 0,
-  "metadata": {
-    "colab": {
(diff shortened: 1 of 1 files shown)


catinwarmhands:
I signed cla!

lamberta:
Thanks! I think you meant to edit [this file](https://github.com/tensorflow/docs/blob/master/site/ru/tutorials/keras/basic_classification.ipynb), though :)

We're still figuring out how community translations will work, but if you're interested, consider signing up here: http://bit.ly/tf-translate

Also can use the [docs@tensorflow.org list](https://groups.google.com/a/tensorflow.org/forum/#!forum/docs) to coordinate.

catinwarmhands:
Oops, my bad :) 
I'll fix it now

PR #322, opened 2019-02-04
Fix minor inaccuracy in russian translation.

catinwarmhands:
"Test dataset" ("проверочный датасет") are sometimes called "training dataset" ("тренировочный датасет"), which can lead to confusion among new users.


lamberta:
I can't speak to the translation, but there is a distinction between training and test data. Does this match the [English version](https://github.com/tensorflow/docs/blob/master/site/en/tutorials/keras/basic_classification.ipynb)? If the English version is confusing, then perhaps we should fix it upstream there first and then update the translations

0101011:
Got to check this one too, there might be a mistake indeed, thanks!

Will be back in 24 hours.

catinwarmhands:
@lamberta This problem only appears in russian translation, english version is fine.

lamberta:
Hi @ezhulenev, when you get a chance, can you sign off on this? Thanks

0101011:
LGTM, missed this one indeed

PR #325, opened 2019-02-08
Add disclaimer to Russian community translations

lamberta:


lamberta:
Example stage: https://colab.research.google.com/github/lamberta/tf-docs/blob/ru-disclaimer/site/ru/guide/keras.ipynb

"Note" will become a callout box when imported into the site.

PR #331, opened 2019-02-09
jupyter notebook rendering of matplotlib through: %matplotlib notebook

SamuelMarks:


lamberta:
Thanks, Samuel. A few notes:
1. We should probably consolidate imports at the top of notebook. I see `import matplotlib.pyplot as plt` uses a couple of times in this notebook.
2. I'm a little unclear of the behavior differences between `%matplotlib inline` vs `%matplotlib notebook` but this doc is ultimately rendered [here on tensorflow.org](https://www.tensorflow.org/tutorials/load_data/images). From earlier discussions, it's a valid use case that a user can copy-paste the code snippets on that page into their Python REPL and things work. Does switching this to `%matplotlib notebook` affect that?

lamberta:
Thanks, but closing this since we have to think about users that *only* use the web page to copy/paste code snippets. While we develop and test with notebooks, we would prefer to not use the magic commands unless necessary.

PR #335, opened 2019-02-13
Add ko/r2/tutorials/keras notebooks

rickiepark:
This notebooks are mostly identical to ko/tutorials/keras.
But It is updated with changes in en/r2/tutorials/keras.
Running with Colab is OK.
Thanks.

@lamberta , I will continue translating `en/r2/tutorials/beginner`. :)


rickiepark:
Thanks for your advice

PR #336, opened 2019-02-15
improved example, basic_classification

vruge:
use  `plt.show()` for ploting example, like in chapter "Preprocess the data"

PR #337, opened 2019-02-15
update ko and ru with changes from #336

vruge:


rickiepark:
I receive all notifications of the repository. If necessary, I will update KO files. :)

PR #338, opened 2019-02-16
force eol for .ipynb

vruge:
The motivation was  [PR337](https://github.com/tensorflow/docs/pull/337), because it change end of file.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.ipynb text eol=lf


lamberta:
Seems right—here's the [best doc](https://help.github.com/articles/dealing-with-line-endings/) I could find.

@MarkDaoust WDYT?

lamberta:
*breaks tensorflow*

PR #340, opened 2019-02-17
Added an example replacing if with `tf.where`

odysseus0:
The original docs mentioned after an example using Python's `if` statement that:

"Typical ML code involves batches; in those cases you should consider using the faster and vectorized `tf.where` if possible."

I added a sample code replacing `if` with `tf.where` in the original the example.

Note: 
I followed the [CONTRIBUTING.md](https://github.com/odysseus0/docs-1/blob/master/CONTRIBUTING.md) and edited the notebook using Colab. I did realize that there is a lot of noise resulting from Jupyter Notebook edit in Colab, even after resetting all the output. Since the [CONTRIBUTING.md](https://github.com/odysseus0/docs-1/blob/master/CONTRIBUTING.md) did not mention what I should do, I did not know what to do.

### DIFF
--- a/site/en/r2/guide/autograph.ipynb
+++ b/site/en/r2/guide/autograph.ipynb
@@ -1,11 +1,29 @@
 {
+  "nbformat": 4,
+  "nbformat_minor": 0,
+  "metadata": {
+    "colab": {
+      "name": "TF2.0 Functions and Autograph",
+      "version": "0.3.2",
+      "provenance": [],
+      "private_outputs": true,
+      "collapsed_sections": [
+        "Jxv6goXm7oGF"
+      ],
+      "toc_visible": true
+    },
+    "kernelspec": {
+      "display_name": "Python 3",
+      "name": "python3"
+    }
+  },
   "cells": [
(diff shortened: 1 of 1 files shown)


brilee:
There is already an example of when `tf.where` is appropriate near the end of the colab. Maybe this is an indication that the flow of the colab isn't as logical as can be?

odysseus0:
> There is already an example of when `tf.where` is appropriate near the end

Duh. I did not even realize that it was all the way at the very bottom. This probably suggests that we need to reorganize the examples a bit.

> Maybe this is an indication that the flow of the colab isn't as logical as can be?

I am with you. We should probably make some adjustment to the [CONTRIBUTING.md](https://github.com/odysseus0/docs-1/blob/master/CONTRIBUTING.md) a bit.

lamberta:
Thanks. Updated the note to link to the "Batching" section. Since it's more advanced, I think it's appropriate that it appears later in the doc. Removed the explicit reference to `tf.where` since that adds some unnecessary confusion.

PR #360, opened 2019-03-03
pip install setuptools is needed when building from source

kpe:
adding `setuptools` to `pip install` instruction otherwise building the wheel package would fail with:
```
 $ bazel-bin/tensorflow/tools/pip_package/build_pip_package /tmp/tensorflow_pkg
Sun Mar 3 22:24:31 CET 2019 : === Preparing sources in dir: /tmp/tmp.lOEexzCqAk
~/proj/local/tensorflow ~/proj/local/tensorflow
~/proj/local/tensorflow
Sun Mar 3 22:24:36 CET 2019 : === Building wheel
  File "setup.py", line 37, in <module>
    from setuptools import Command
ModuleNotFoundError: No module named 'setuptools'
```

### DIFF
--- a/site/en/install/source.md
+++ b/site/en/install/source.md
@@ -37,7 +37,7 @@ Install the TensorFlow *pip* package dependencies (if using a virtual environmen
 omit the `--user` argument):
 
 <pre class="prettyprint lang-bsh">
-<code class="devsite-terminal">pip install -U --user pip six numpy wheel mock</code>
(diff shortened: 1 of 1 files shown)


lamberta:
@gunan PTAL when you get a chance, here's the [build from source](https://www.tensorflow.org/install/source) guide. Can you confirm `setuptools` is required?

PR #361, opened 2019-03-03
First pass at updating the community section.

dynamicwebpaige:


ewilderj:
I'm really confused as to what this page is for. It has spurious amounts of detail, and some copypasta from the RFC docs in the middle without context that are confusing. Can you take another look at this as a standalone page to see if it makes sense to you?

A few other oddnesses, there are subheads "Forums" and "Forums and user groups". The text that says the primary discussion for contributing to TF is on GitHub needs developers@ adding to it. Also, GitHub is inconsistently rendered as "Github" in places.

This sentence is awkward "If you submit an algorithm or feature request, verify the algorithm has been vetted by an RFC" - the RFC link should go to the RFC process doc, vs the directory, but also the sentence doesn't really make sense. Should be something like "If you want to work on significant new functionality, please discuss on developers@tensorflow.org first and propose an RFC before submitting code."

MarkDaoust:
I rolled-this back in  commit e71b380c7d43.

PR #362, opened 2019-03-04
Added installation instructions for the TF 2.0 alpha.

dynamicwebpaige:


lamberta:
After discussing with @MarkDaoust, we're not comfortable updating this package name until we are certain it's there.
We can make follow-up changes when we know it works.

PR #372, opened 2019-03-07
added import to optional tqdm instructions

DecentGradient:
added import statement to markdown


DecentGradient:
something went wrong, that's all we know! redoing

PR #374, opened 2019-03-07
Minor fixes in /tensorflow/docs/blob/master/site/en/r2/tutorials/sequences/image_captioning.ipynb

DecentGradient:
added import instructions for tqdm


site/en/r2/tutorials/sequences/image_captioning.ipynb
```diff
@@ -572,7 +572,7 @@
       "source": [
         "# loading the numpy files \n",
         "def map_func(img_name, cap):\n",
-        "  img_tensor = np.load(img_name.decode('utf-8')+'.npy')\n",
+        "  img_tensor = np.load(img_name.numpy().decode('utf-8')+'.npy')\n",
```
MarkDaoust: I don't believe this change is necessary.

The deprecation notice for numpy_function is here:

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/script_ops.py#L463

It explains that the difference between `numpy_funciton` and `py_function` is that numpy_function recieves np.arrays as arguments.

numpy arrays do not have a `.numpy` method.

 `np.array(image_name)` would work on either a tf.tensor or an `np.array`.

PR #381, opened 2019-03-09
Add print method

asadman1523:


asadman1523:
I signed it!

lamberta:
Thank you for the submission.
These print calls are not required since the notebooks with display the last returned value of the cell, and this is what we use to generate the html here: https://www.tensorflow.org/tutorials/keras/basic_classification
I think adding an extra function nesting detracts from the overall readability of the notebook.

PR #392, opened 2019-03-17
Add community translation note/readme for Portuguese.

lamberta:
Added note to README and notebook.

stage: https://colab.research.google.com/github/lamberta/tf-docs/blob/pt-note/site/pt/tutorials/keras/basic_classification.ipynb

### DIFF
--- a/site/pt/README.md
+++ b/site/pt/README.md
@@ -1,0 +1,30 @@
+# Community translations
+
+A nossa comunidade TensorFlow traduziu estes documentos. Como as traduções da
+comunidade são *o melhor esforço*, não há garantias de que sejam uma reflexão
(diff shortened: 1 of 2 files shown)


lamberta:
Hi @catiafortunato,
Thank you for translating the note, I've added it to the notebook and created an initial `pt/README.md` file.
Another request, I need to translate the header "Community translations", can you help? :)
Also, in the readme I started a *do not translate* list that was based on the current `ru/README.md`. Let me know if this is a good starting point (we can always add more).
Thank you!

catiafortunato:
Hi, the header translation is "Traduções da Comunidade".
I think the list makes sense and is a good starting point, yes! :)

PR #393, opened 2019-03-17
fixing typos in the migration doc

bayethiernodiop:


bayethiernodiop:
One question, since this is merged why the modifications are not yet visible on the tensorflow website. thx

lamberta:
There is a separate publish step. We usually batch these together every week or so.

bayethiernodiop:
make sense. thx

PR #395, opened 2019-03-18
Fixed pip installation in GPU example.

breuera:


breuera:
CLA is signed; although the change is trivial.

PR #418, opened 2019-03-24
Updates PTB link

MaximumEndurance:


lamberta:
Thanks, but we need to link to the official page here

PR #423, opened 2019-03-25
RU: Readme.md - Added underscore docs notice

0101011:
1. Better wording 😊

2. Putting a warning sign about underscore docs being still in development and won't be pulled yet


0101011:
Perfectly right, you’ll be understood in any of 10-13 countries around here 😃
(See Slavic branch for reference)

PR #433, opened 2019-03-28
Add comments about  enable_eager_execution no longer being necessary

JohnAllen:
Just went to work with this tutorial and the error message when you run this using TF version 2.0 is `AttributeError: module 'tensorflow' has no attribute 'enable_eager_execution'` which led me to believe I had a version problem, not that I could remove this line of code altogether.  

So this comment is to help those that are using version 2.0 know that they can just remove this line of code.

### DIFF
--- a/site/en/tutorials/load_data/images.ipynb
+++ b/site/en/tutorials/load_data/images.ipynb
@@ -103,7 +103,7 @@
         "from __future__ import absolute_import, division, print_function\n",
         "\n",
         "import tensorflow as tf\n",
-        "tf.enable_eager_execution()\n",
+        "tf.enable_eager_execution() # Only necessary in Tensorflow versions prior to 2.0\n",
         "tf.__version__"
       ]
     },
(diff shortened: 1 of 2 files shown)


lamberta:
Thanks! But I think we want to keep this as-is since this is still part of the stable docs (currently, r1.13) and we'd be in trouble if we have to keep calling out differences between TF 1.x and TF 2 :)

TF 2.0 Alpha docs are here on the site: https://www.tensorflow.org/alpha/tutorials/load_data/images

JohnAllen:
Those Alpha docs link to this notebook though (from the Github link at top).

That's how I ended up in this notebook using 2.0.  Is there a separate 2.0 images notebook that we could link to?

lamberta:
Ah. Yeah, those notebooks are over here under the `r2/` directory: https://github.com/tensorflow/docs/blob/master/site/en/r2/tutorials/load_data/images.ipynb

The notebook buttons look updated, but we may need to regenerate that section of the site.

PR #434, opened 2019-03-28
Changing tf.cond by if statement in custom_layers_and_models.ipynb

LastRemote:
In TensorFlow 2.0, we could use simple if statement instead of tf.cond for better understanding. It also gets rid of those lambda functions.


LastRemote:
I signed it!

yashk2810:
Thanks for the PR!

PR #388, opened 2019-03-13
Fix: save_to_tsv in word embedding tutorial

dreamnew:
#### Original code:
out_m.write(word + "\n")
out_v.write('\t'.join([str(x) for x in embeddings]) + "\n")
#### New Code:
out_m.write(word + "\n".encode('utf-8'))
out_v.write('\t'.join([str(x) for x in embeddings]).encode('utf-8') + "\n".encode('utf-8'))

PR #446, opened 2019-04-01
Fix concat str to bytes TypeError in word_embeddings.ipynb

maziyarpanahi:
Hi,

There is already a pull request to [fix this issue](https://github.com/tensorflow/docs/pull/388), however, it has 4 `.encode` to fix a simple problem. All we need to do is to remove the redundant `.encode()` from the end of this line.

```python
 word = reverse_word_index[word_num].encode('utf-8')
```

old:

```python
out_v = open('vecs.tsv', 'w')
out_m = open('meta.tsv', 'w')
for word_num in range(vocab_size):
  word = reverse_word_index[word_num].encode('utf-8')
  embeddings = weights[word_num]
  out_m.write(word + "\n")
  out_v.write('\t'.join([str(x) for x in embeddings]) + "\n")
out_v.close()
out_m.close()
```

new:

```python
out_v = open('vecs.tsv', 'w')
out_m = open('meta.tsv', 'w')
for word_num in range(vocab_size):
  word = reverse_word_index[word_num]
  embeddings = weights[word_num]
  out_m.write(word + "\n")
  out_v.write('\t'.join([str(x) for x in embeddings]) + "\n")
out_v.close()
out_m.close()
```
This works without any problem while keeping it simple as it is already.

### DIFF
--- a/site/en/r2/tutorials/sequences/word_embeddings.ipynb
+++ b/site/en/r2/tutorials/sequences/word_embeddings.ipynb
@@ -1,584 +1,597 @@
 {
-  "cells": [
-    {
-      "cell_type": "markdown",
-      "metadata": {
-        "colab_type": "text",
-        "id": "GE91qWZkm8ZQ"
-      },
-      "source": [
-        "##### Copyright 2019 The TensorFlow Authors."
-      ]
-    },
-    {
-      "cell_type": "code",
-      "execution_count": 0,
-      "metadata": {
-        "cellView": "form",
-        "colab": {},
-        "colab_type": "code",
-        "id": "YS3NA-i6nAFC"
-      },
-      "outputs": [],
-      "source": [
-        "#@title Licensed under the Apache License, Version 2.0 (the \"License\");\n",
-        "# you may not use this file except in compliance with the License.\n",
-        "# You may obtain a copy of the License at\n",
-        "#\n",
-        "# https://www.apache.org/licenses/LICENSE-2.0\n",
-        "#\n",
-        "# Unless required by applicable law or agreed to in writing, software\n",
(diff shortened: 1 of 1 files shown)


yashk2810:
We need the .encode('utf-8') to test in python2.

maziyarpanahi:
I understand, but Python 2 EOL is January 2020. Do we really need to make the code look that complicated for a version that’ll be deprecated by the end of this year?

yashk2810:
We test this notebook in python 2. That is the reason we need the encode('utf-8'). This should be removed in the near future though :)

maziyarpanahi:
> We test this notebook in python 2. That is the reason we need the encode('utf-8'). This should be removed in the near future though :)

I totally understand. Since the original pull request did actually solve this for both Py2 and Py3 but it was closed for styling reason, would you like me to commit another time and change it to the same?

```python
out_m.write(word + "\n".encode('utf-8'))
out_v.write('\t'.join([str(x) for x in embeddings]).encode('utf-8') + "\n".encode('utf-8'))
```

lamberta:
Here's the TensorFlow statement for Py 2 support: https://groups.google.com/a/tensorflow.org/d/msg/developers/aHZFv09z4Xw/jPAL8ndmAgAJ
Until then, notebooks should run in both Python 2.7 and Python 3. After that, we can remove Python 2 code to support the latest release going forward.

maziyarpanahi:
Thanks, but this notebook doesn’t run in Python 3.

lamberta:
Just verified that the current notebook executes using the Python 3 runtime with GPU: https://colab.research.google.com/github/tensorflow/docs/blob/master/site/en/r2/tutorials/text/word_embeddings.ipynb

PR #455, opened 2019-04-05
Add matplotlib plot inline command

MostafaGazar:
Add matplotlib plot inline command to tutorials/eager/custom_training.ipynb


lamberta:
Thanks, but I don't think this is necessary since the notebook default behavior is to inline.

MostafaGazar:
Sure, for some reason though if you run it outside of Colab in a normal Jupyter notebook. It does not plot anything the first time you call `plot.show()` if `%matplotlib inline`  was not set.

PR #458, opened 2019-04-05
Update test script for pip install

yselivonchyk:
Remove tf.enable_eager_execution() from installation verification scripts. Tensorflow now has eager execution enabled by default.


yselivonchyk:
I signed it!

lamberta:
Thanks, but we're not at 2.0 stable, yet. We still need to enable eager for this to work.

PR #459, opened 2019-04-05
Delete unnecessary semicolons

siyavash:
Found unnecessary semicolons in ipynb files.


yashk2810:
If semicolon is not included, then this text gets printed "Text(0, 0.5, 'Sepal length')".

So, I think we need the semicolon.

siyavash:
Oh, I didn't notice using semicolon in jupyter notebooks can suppress output of last line.
So, what's your suggestion ? 
Add plt.show() or close this PR ?
IMO, it's better to add plt.show(), because in other parts of tutorial plt.show() is used.

yashk2810:
Adding plt.show() would work.

lamberta:
Merged here: https://github.com/tensorflow/docs/commit/18dec2e29531b9fef26de3db8458af3ac8fe66ba

siyavash:
I think this commit [eed8604](https://github.com/tensorflow/docs/pull/459/commits/eed8604f7240924c65aa9c978419f83efcbff382) is necessary, because in boosted_tree.ipnyb, plt.show() is used before importing.

lamberta:
Yes, I see.
Once the 'ready to pull' label is applied the merge process has started for the PR. I'll create a separate PR for that last commit.

PR #465, opened 2019-04-06
Added bazel version specific message

biltongza:
Fixes https://github.com/tensorflow/tensorflow/issues/27575


av8ramit:
We should also update "Tested build configurations".

PR #467, opened 2019-04-07
corrected error in examples

VariableVasasMT:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/docs/467)


VariableVasasMT:
<p>ran with tensorflow 2.0.0-alpha</p>

lamberta:
Thanks!
Can you `git mv` this over to the [/r2/tutorials/images/](https://github.com/tensorflow/docs/blob/master/site/en/r2/tutorials/images/) directory?
Then install TF 2 in the notebook:

```
!pip install tensorflow-gpu==2.0.0-alpha0
import tensorflow as tf
```

We're currently not publishing this on the site (it starts with an underscore), but we should try and get it in shape for the TF 2 release.

PR #475, opened 2019-04-10
KO: Update to sync with en

rickiepark:
Update `ko/alpha/guide/migration_guide.ipynb` and `ko/alpha/tutorials/keras/index.md`.
Thanks!

### DIFF
--- a/site/ko/alpha/guide/migration_guide.ipynb
+++ b/site/ko/alpha/guide/migration_guide.ipynb
@@ -759,7 +759,7 @@
       },
       "cell_type": "code",
       "source": [
-        "datasets, ds_info = tfds.load(name='mnist', with_info=True, as_supervised=True)\n",
+        "datasets, info = tfds.load(name='mnist', with_info=True, as_supervised=True)\n",
         "mnist_train, mnist_test = datasets['train'], datasets['test']"
       ],
       "execution_count": 0,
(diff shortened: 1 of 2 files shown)


lamberta:
Hey Haesun, I made a [little script](https://github.com/tensorflow/docs/blob/master/tools/nb_code_sync.py) to keep the code cells of translated notebooks in sync, feel free to try it out.

Usage: `./tools/nb_code_sync.py site/ko/notebook.ipynb`

It only checks code blocks (and requires they have the same amount in each notebook). Since it overwrites comments in changed cells, you probably want to use `git add --patch` to selectively add hunks of the file. Lemme know how it goes.

rickiepark:
Hi Billy,
Thank you, nb_code_sync.py is useful. I use --src option, because ko/alpha's src is en/r2.
But I still have to use git diff, because markdown changes is not detected. :(

lamberta:
Thanks for checking, Haesun.
Yes, it does not compare Markdown cells. That decision was deliberate since I've found that devs will tolerate out-of-date text more than out-of-date code. And it's easy for me to compare code cells and difficult for me to compare translated text.
I'm open to ideas, though. We can't rely on cell IDs since we may lose them if notebooks get rendered out from Colab or Jupyter, depending on the contributor's preference. I would be nice to keep track of the commit hash where the notebook was lasted translated and then diff from that point. But I'm not sure if that would be helpful to your workflow.

What do you think? It would be a nice to come up with an easy workflow to point contributors to and encourage keeping things up to date. Have you found a good way to do this?

PR #476, opened 2019-04-10
Name scope style instruction update

armando-fandango:


ewilderj:
Needs engineer or DevRel to approve code related change.

PR #485, opened 2019-04-12
JA: added translated files in site/ja/r2/tutorials/keras

masa-ita:
I've translated files (a markdown file and notebooks) in site/ja/r2/tutorials/keras.


lamberta:
Looping in some potential reviewers: @ohtaman, @sfujiwara

PR #486, opened 2019-04-12
us the fc variable declared to remove tf.feature_column redundancy

bayethiernodiop:


yashk2810:
As @lamberta said, its nice to have the full imports since we need it to parse the notebook.

lamberta:
Thanks for the submission, but decided to keep it as-is

PR #490, opened 2019-04-15
correct wrong dimensions

LevineHuang:


yashk2810:
Can you make the change without running the notebook so that the diff remains small? Right now its hard to find the change that you made.

lamberta:
Thanks. Unfortunately, unable to diff and getting a merge conflict. Closing this pull request, but let me know if the problem hasn't been fixed in another PR.

PR #494, opened 2019-04-16
Added commands to install Nvidia driver for CUDA 9

rogerthat94:
The Ubuntu 16.04 (CUDA 9.0 for TensorFlow < 1.13.0) install instructions did not include commands for installing an Nvidia driver

### DIFF
--- a/site/en/install/gpu.md
+++ b/site/en/install/gpu.md
@@ -157,6 +157,12 @@ complicates installation of the NVIDIA driver and is beyond the scope of these i
 <code class="devsite-terminal">sudo apt install ./nvidia-machine-learning-repo-ubuntu1604_1.0.0-1_amd64.deb</code>
 <code class="devsite-terminal">sudo apt update</code>
 
+# Install NVIDIA Driver
+# Issue with driver install requires creating /usr/lib/nvidia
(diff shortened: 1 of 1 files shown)


tfboyd:
Looking, I do not believe that is the correct driver for 9.0.  It will work, but it newer than you need.  Checking a few other items.  

What are your thoughts. @rogerthat94   I think CUDA 9.0 was 384, newer drivers are fine.  I am looking at the docs to ensure we are not staying you need 410.

tfboyd:
Changed my mind, no reason not to just use 410 so they can then upgrade to 1.14 or 2.0 more easily.  We make it clear you do not need 410 earlier in the document for 9.0

tfboyd:
@MarkDaoust  back to you

lamberta:
Thanks, @tfboyd!

rogerthat94:
There might be someone using an old card that supports 384 and not 410? I'm not sure if Nvidia stopped supporting any cards in between 384 and 410. However, this is probably a very small number of people, and they would probably realize they should install the latest version that works for their card if 410 doesn't work. 

For people with cards that support 410, I completely agree that there is no reason not to just use 410.

PR #499, opened 2019-04-18
enable v2 behavior is not available

abbazs:
enable_v2_behavior is not available in the latest builds.
Calling it produces:
AttributeError: module 'tensorflow' has no attribute 'enable_v2_behavior'


abbazs:
I signed it!

PR #501, opened 2019-04-19
Update pip install to Py 3.7

lamberta:
Confirmed in https://github.com/tensorflow/tensorflow/issues/17022


PathToLife:
https://github.com/tensorflow/tensorflow/issues/17022#issuecomment-486493494
As of now (25 April 2019), pip on 3.7 does not work.

PR #504, opened 2019-04-20
Enable use of PNG input images.

ilopezfr:
Replacing `decode_jpeg` by the generic `decode_image` and setting up `channels=3` allows the use of this util function to work with PNG images, which come with 4 channels.


ilopezfr:
I signed it!

PR #522, opened 2019-04-24
JA: add js/tutorials/training

technohippy:
I translated below into Japanese.
https://github.com/tensorflow/tfjs-website/blob/devsite/docs/tutorials/training


technohippy:
I signed with another email.

PR #524, opened 2019-04-24
use the variable kernel_size in all conv2d

bayethiernodiop:
use the variable kernel_size in all conv2d  instead of (1,1) in the first and last layer


bayethiernodiop:
one question : at the end of the notebook what is the notebook that is referred pls

site/en/r2/tutorials/eager/custom_layers.ipynb
```diff
@@ -276,13 +276,13 @@
         "    super(ResnetIdentityBlock, self).__init__(name='')\n",
         "    filters1, filters2, filters3 = filters\n",
         "\n",
-        "    self.conv2a = tf.keras.layers.Conv2D(filters1, (1, 1))\n",
+        "    self.conv2a = tf.keras.layers.Conv2D(filters1, kernel_size)\n",
```
yashk2810: This is supposed to be 1x1 because its a 3 layer resnet identity block according to the resnet architecture.

yashk2810: Please see the diagram here (https://www.researchgate.net/figure/Bottleneck-Blocks-for-ResNet-50-left-identity-shortcut-right-projection-shortcut_fig2_321347448) or the architecture in the paper.

Thanks but closing this PR :)

bayethiernodiop: yes, i agree but it's already the case when we use the variable kernel_size which is equals to 1

yashk2810: Yes, but the kernel_size can be changed to anything, but the first and the last layer in the block should still be 1x1.

bayethiernodiop: ah, ok. you right, thanks.

lamberta: This might make a good code comment :)

yashk2810: Yeah, also you can change the kernel_size parameter to 3

PR #535, opened 2019-04-25
[Experimental] Add Text lint with RedPen

chie8842:
In Japan, We are working on translation work with more than one person. I suggest automatic text lint so that it doesn't occur differences in writing format between translators.


lamberta:
This is an interesting idea but we're not prepared to add it to the repo now.
Could you propose the idea to the docs@tensorflow.org for evaluation? Thanks

PR #536, opened 2019-04-25
Update images.ipynb

wenqinlong:
"ve" should be "be".

### DIFF
--- a/site/en/tutorials/load_data/images.ipynb
+++ b/site/en/tutorials/load_data/images.ipynb
@@ -1,4 +1,4 @@
- {
+{
   "cells": [
     {
       "cell_type": "markdown",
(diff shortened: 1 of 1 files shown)


site/en/tutorials/load_data/images.ipynb
```diff
@@ -1,4 +1,4 @@
-{
+v{
```
lamberta: Looks like this parse error sneaked in

wenqinlong: Sorry for that. I also doubt that this explanation about `repeat` and `shuffle` has a conflict with this [document](https://www.tensorflow.org/guide/performance/datasets).
The detail is seen in the [stackoverflow question](https://stackoverflow.com/questions/55851440/is-there-a-conflict-in-these-two-tensorflow-documents-about-repeat-and-shuffle) I asked.

PR #537, opened 2019-04-25
Fix api_generator.parser._gen_pairs to work in Python 3.7

StanislawSwierc:
In Python 3.7 recommendation from PEP 479 was enabled. In the new
behavior StopIteration exceptions raised directly or indirectly in coroutines
and generators are transformed into RuntimeError exceptions.

https://www.python.org/dev/peps/pep-0479/

### DIFF
--- a/tools/tensorflow_docs/api_generator/parser.py
+++ b/tools/tensorflow_docs/api_generator/parser.py
@@ -386,9 +386,7 @@ def _gen_pairs(items):
     The original items, in pairs
   """
   assert len(items) % 2 == 0
-  items = iter(items)
-  while True:
-    yield next(items), next(items)
+  return zip(items[::2], items[1::2])
 
 
 class _FunctionDetail(


StanislawSwierc:
I've just signed the Contributor License Agreement.

lamberta:
Thanks, Stanislaw.
I looks like you signed the CLA using a different email address than the one used for this commit. Can you sign the CLA with this email or re-submit with the other?

tools/tensorflow_docs/api_generator/parser.py
```diff
@@ -386,6 +386,8 @@ def _gen_pairs(items):
     The original items, in pairs
   """
   assert len(items) % 2 == 0
+  if not items:
+    return
```
yashk2810: How about this instead to iter?

@MarkDaoust 
```
zip(items[::2], items[1::2])
```

StanislawSwierc: This looks much cleaner! Thx.

PR #542, opened 2019-04-27
Fix customized training in RNN text generation tutorial

qiubit:
The network in "Customized training" section of "Text generation" RNN tutorial for TensorFlow 2.0 doesn't train properly due to logits being passed to loss function that expects probability distributions and not reseting state per each epoch. This is fixed with this pull request.


yashk2810:
Thank you :)

PR #547, opened 2019-04-30
Update tf2 advanced tutorial to use plural for batches

Tenoke:
Use plural for images and labels to make clear that we
are working with a batch and not a single instance


Tenoke:
I signed it!

PR #548, opened 2019-04-30
minor missing text

DecentGradient:
assumes the feature vector model from hub needs to be tf2 compatible

### DIFF
--- a/site/en/tutorials/images/hub_with_keras.ipynb
+++ b/site/en/tutorials/images/hub_with_keras.ipynb
@@ -541,7 +541,7 @@
         "\n",
         "TensorFlow Hub also distributes models without the top classification layer. These can be used to easily do transfer learning.\n",
         "\n",
-        "Any [image feature vector url from tfhub.dev](https://tfhub.dev/s?module-type=image-feature-vector\u0026q=tf2) will work here."
+        "Any [tf2 compatible image feature vector url from tfhub.dev](https://tfhub.dev/s?module-type=image-feature-vector\u0026q=tf2) will work here."
       ]
     },
     {


yashk2810:
The directory you made change to contains notebooks that run TF 1.x. For TF2, please see site/r2/en/...

DecentGradient:
Are you sure? The link is a query for the term tf2 and I got to it from the GitHub link in the tf2 tutorial and the earlier link to hub mentions tf2 and this is now a run on sentence

lamberta:
The fix should be made to this file: https://github.com/tensorflow/docs/blob/master/site/en/tutorials/images/hub_with_keras.ipynb

And it looks like the button links also need to be fixed in that notebook since the page does link to the TF1 notebook: https://www.tensorflow.org/alpha/tutorials/images/hub_with_keras

@DecentGradient can you also fix those links while you're in there?

DecentGradient:
@lamberta Sure, just to clarify, was I in the right file or not?

DecentGradient:
@lamberta sure, as far as i can tell there are two links to hub in this tutorial. Currently, in both versions, there is one link (the first)  that searches for tf2 models and the second searches for tf2 but doesnt mention it in the link text.
I will:

1. fix the link text in the second link in r2
2. fix both links in r1 to not have tf2 in the query or link text
3. check the button links in r2 and update them (i think the r1s are ok)

lamberta:
Sorry, I'll clarify. You can revert this file and only update the notebook under r2/

DecentGradient:
Done
[https://github.com/tensorflow/docs/pull/549](https://github.com/tensorflow/docs/pull/549)
[https://github.com/tensorflow/docs/pull/550](https://github.com/tensorflow/docs/pull/550)

DecentGradient:
@lamberta I did it in separate PRs. (:face_with_head_bandage: gitfu through github) 
thanks for the clarification

PR #549, opened 2019-05-01
Hub w/ Keras r1 -> r2

DecentGradient:
updated button links and link to hub


DecentGradient:
@lamberta how long should it take from when a pull request is merged until it shows up on tf.org?

lamberta:
@DecentGradient There's a manual step to convert the notebooks and update the site. Generally, we batch a bunch of changes together. But there's a site update in the works so should be up tonight

DecentGradient:
@lamberta thanks

PR #551, opened 2019-05-02
Fix sample code in Japanese tutorial

maeda-kazuya:
- Add missing code; calling "plt.show()"
- Remove unnecessary code


maeda-kazuya:
Hi, I'm trying the tutorial of TensorFlow and noticed some sample code doesn't work.
It looks like missing "plt.show()" to show the image, so I added it.
Please point out if I'm missing anything, thanks.

site/ja/tutorials/keras/basic_classification.ipynb
```diff
@@ -487,17 +488,15 @@
       },
       "cell_type": "code",
       "source": [
-        "import matplotlib.pyplot as plt\n",
-        "%matplotlib inline\n",
-        "\n",
```
maeda-kazuya: matplotlib.pyplot is already imported above, so I think this part is not necessary.

site/ja/tutorials/keras/basic_classification.ipynb
```diff
@@ -797,7 +796,7 @@
         "    thisplot = plt.bar(range(10), predictions_array, color=\"#777777\")\n",
         "    plt.ylim([0, 1])\n",
         "    predicted_label = np.argmax(predictions_array)\n",
-        "\n",
```
maeda-kazuya: I also noticed that this sample code doesn't work if I just copy & paste because of unaligned indent.

site/ja/tutorials/keras/basic_classification.ipynb
```diff
@@ -797,7 +796,7 @@
...
+        "    \n",
```
maeda-kazuya: I also noticed that this sample code doesn't work if I just copy & paste because of unaligned indent.

masa-ita: Did you paste these code on python command prompt?

These example codes are intend to be executed on jupyter notebooks.
Blank lines are ignored before execution except in python interpreter command prompt.

Although lint program like pycodestyle/pep8 warns with "W293 blank line contains whitespace", it seems good for compatibility to add indent on blank lines.

I'm not sure which is better.  There are argument on this topic.
https://stackoverflow.com/questions/2727988/python-indentation-in-empty-lines

site/ja/tutorials/keras/basic_classification.ipynb
masa-ita: Thank you for your PR.
matplotlib.pyplot is imported in first code cell, and the magic command "%matplotlib inline" is omitted in other notebooks, so I think these lines are OK to be removed.

site/ja/tutorials/keras/basic_classification.ipynb
MarkDaoust: I'm in the  "if the tooling doesn't do it automatically, don't spend too much time on it" camp.

PR #552, opened 2019-05-03
zh-CN: add Chinese translation for 2 files

kuri-leo:
zh-ch: add Chinese translation of `site\zh-cn\tutorials\keras\index.md` && `site\zh-cn\tutorials\keras\basic_classification.ipynb`

### DIFF
--- a/site/zh-cn/tutorials/keras/basic_classification.ipynb
+++ b/site/zh-cn/tutorials/keras/basic_classification.ipynb
@@ -1,0 +1,994 @@
+{
+  "nbformat": 4,
+  "nbformat_minor": 0,
+  "metadata": {
+    "colab": {
+      "name": "basic_classification.ipynb",
+      "version": "0.3.2",
+      "provenance": [],
+      "private_outputs": true,
+      "collapsed_sections": [],
+      "toc_visible": true
+    },
+    "kernelspec": {
+      "name": "python3",
+      "display_name": "Python 3"
+    }
+  },
+  "cells": [
+    {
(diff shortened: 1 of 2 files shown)


kuri-leo:
~~Looks good for me(of course~~, But I strongly suggest that someone else reviews my translation.

There are also some questions:

1. About URL. Should the URL be translated? For example: `www.tensorflow.org` --> `tensorflow.google.cn`, the second one is much friendly with Chinese network ;-) or if there are official tools can do this word automatically? (free us from this trivial work
2. About glossary. there're few non-translate words listed in `README.md` though, how about to use [official glossary](https://developers.google.cn/machine-learning/glossary/) as the complete list?
3. Text in Jupyter Notebook is very complex to locate and translate...

duanw0916:
LGTM

byronyi:
It is hard to translate and review the Jupyter notebooks. Shall we use markdown for the docs instead?

kuri-leo:
Hi @byronyi, thanks for your advice, but I'm sorry that I may not agree with your idea. Docs we translate should be in completely same formate with the original ones, for easier building and publishing.

I translated it with opening it via VSCode, not Jupyter Notebook, as I'm afraid Jupyter and plugins may create more files like cache. This can bring unexpected changes or files and finally "pollute" the repo.

Personally, I strongly suggest that we, the translators, **SHOULD NOT** run any code in Jupyter locally. Run it in Colab (read-only mode) after submitting commits to GitHub is a much better choice for testing and validation.

PR #554, opened 2019-05-03
[KO] update site/ko/alpha/tutorials/keras

rickiepark:
update from 09ee44669153ca2964a806e9064f3057e79eb7a7
(Because the problem when load pickling object in numpy 1.16.3)


rickiepark:
Hi @lamberta , I just change it because these two notebooks are omitted in https://github.com/tensorflow/docs/commit/09ee44669153ca2964a806e9064f3057e79eb7a7 :)

PR #558, opened 2019-05-07
Mention TF_FORCE_GPU_ALLOW_GROWTH in using_gpu.md

lmartak:
As mentioned in [this comment](https://github.com/tensorflow/tensorflow/commit/5d2047029a77545c97c0fdf74d9c03c92d1dcb88#commitcomment-33445269) to the commit of discussed functionality, I'd like to have it mentioned here, as it is the most relevant place for it to be mentioned that I could find.

Since I failed to find info about whether this feature is intended to be kept around and integrated into TF2 in the future development, or its future is not that certain, I'd like to at least start the discussion through this PR and maybe advocate for it, as I recently found it most useful and would really appreciate for it to be standardized/propagated into future TF2 stable releases.

Thanks for any reply!

### DIFF
--- a/site/en/guide/using_gpu.md
+++ b/site/en/guide/using_gpu.md
@@ -104,10 +104,8 @@ config.gpu_options.allow_growth = True
 session = tf.Session(config=config, ...)
 ```
 
-To enable this option by default, you can also instead just set the [environment 
-variable](https://en.wikipedia.org/wiki/Environment_variable) 
-`TF_FORCE_GPU_ALLOW_GROWTH` to `true`. Note, that this is a platform specific 
-setup and is done outside of the Python environment.
+To enable this option by default, set the environmental variable
+`TF_FORCE_GPU_ALLOW_GROWTH` to `true`. This configuration is platform specific.
 
 The second method is the `per_process_gpu_memory_fraction` option, which
 determines the fraction of the overall amount of memory that each visible GPU
--- a/site/en/guide/using_gpu.md
+++ b/site/en/guide/using_gpu.md
@@ -104,7 +104,7 @@ config.gpu_options.allow_growth = True
 session = tf.Session(config=config, ...)
 ```
 
-To enable this option by default, set the environmental variable
+Another way to enable this option is to set the environmental variable
 `TF_FORCE_GPU_ALLOW_GROWTH` to `true`. This configuration is platform specific.
 
 The second method is the `per_process_gpu_memory_fraction` option, which
--- a/site/en/guide/using_gpu.md
+++ b/site/en/guide/using_gpu.md
@@ -104,6 +104,11 @@ config.gpu_options.allow_growth = True
 session = tf.Session(config=config, ...)
 ```
 
+To enable this option by default, you can also instead just set the [environment 
+variable](https://en.wikipedia.org/wiki/Environment_variable) 
+`TF_FORCE_GPU_ALLOW_GROWTH` to `true`. Note, that this is a platform specific 
+setup and is done outside of the Python environment.
+
 The second method is the `per_process_gpu_memory_fraction` option, which
 determines the fraction of the overall amount of memory that each visible GPU
 should be allocated. For example, you can tell TensorFlow to only allocate 40%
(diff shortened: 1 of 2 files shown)


lmartak:
I signed it!

lamberta:
Thanks, Lukas! Looks handy but I'm wondering if the platform-specific nature of this is why we're not surfacing it in the docs. Any insights into this @tfboyd?

As for inquiring about its status in TF 2, that would be a better question for either the [testing@tensorflow.org](https://groups.google.com/a/tensorflow.org/forum/#!forum/testing) or [developers@tensorflow.org](https://groups.google.com/a/tensorflow.org/forum/#!forum/developers) groups.

tfboyd:
I may totally miss the point, so please correct me.  I do not think we want to share another way of turning on the GPU growth feature.  the config proto is what we want people to use not the ENV VAR.  If there is some advantage to the ENV_VAR other than someone liking to use them, please let me know and we can sort this out.

lmartak:
@lamberta Thanks for the links, I'll ask about the future plans there.

@tfboyd Thanks for the comment. I understand your concerns, let me clarify this. + Apologies for the size and many thanks for your time if you're going to read this!

I am not sure about the design decisions that led to implementation of this feature, so I want to clarify how much one can rely on it in the future. But since it is there, useful to at least system admins (e.g. people configuring runtimes that are served through [Colab](https://colab.research.google.com/) or any other multi-user or multi-instance system), it might as well be documented somewhere, as other people administrating multi-user systems might want to make use of it in the future.

> I do not think we want to share another way of turning on the GPU growth feature.

I'd disagree that this is just another way of doing the same thing. It is rather the only way to persistently set a default availability of the feature. This is a different thing to do, conceptually, and is especially useful when high-level libraries such as `tf.keras` or `tf.estimator` start meddling with the existing user-configured `session` under the hood in undocumented ways, screwing the config up along the way (own experience, separate issue material).

> some advantage to the ENV_VAR

In a single GPU multi-user system, one can use this feature to set a system-wide default instead of forcing every user to include a code snippet that changes the TF-enforced default (disabled) to the one that is actually preferred (enabled) and ensure it is present all of their projects on every spot where default / active `session` is being manipulated.

I can imagine that this is not of much relevance in industry, where there is plenty of hardware for every practitioner, but in academia, the story is quite opposite and scarcity of hardware sometimes implies multiple practitioners working on a single piece of CUDA-capable hardware in parallel. To give some examples from experience:
 * [JupyterHub](https://jupyter.org/hub) for class/lab: multiple students running their python jobs, each using only a fraction of GPU space-time, so the shared GPU resource gets utilized by those jobs much better, in parallel
 * Multiple users prototyping their models while measuring it's "GPU memory hunger" (as determined by TF) or running long training jobs in parallel.

I understand that this is not an intended use of `allow_growth`, but most novice practitioners don't get anywhere near 100% GPU Utilization even when employing `tf.data` input pipeline optimizations. Therefore, aforementioned use-cases stay relevant. Also, as PyTorch works in `allow_growth=True` by default (not sure if it has a `False` option though), this enables to PyTorch and TensorFlow jobs to utilize GPU resources in parallel.

Being able to impose own preference over this default on the scope of whole maintained system is a great feature to have (even though quite use-case specific) and I just wished for it to be advertised [a bit better than just this](https://stackoverflow.com/a/54927335/11465567).

I hope this clarifies my viewpoint enough.

Thanks for your time and for any reply!

tfboyd:
More than enough to justify. You had me at notebook and needing to set this outside the user code.  I would like the updated wording to say something like this is another way to set the value.  I hope to come back to this later today and hopefully a fast back and forth then submit before the week is over at the latest.  After all your typing and reading I don't want this lost.

lamberta:
Thanks @tfboyd and @lmartak 
Reworded text a bit, but we're back at Lukas's question about TF2 ... can the TF2 gpu guide also be updated with this? [r2/guide/using_gpu.md](https://github.com/tensorflow/docs/blob/master/site/en/r2/guide/using_gpu.md)

tfboyd:
This feature is supposed to work with TF 2.0, so Yes.  If it does not work, that is a bug.

lamberta:
Thanks. Also added to r2 guide since it *should* work :)

lmartak:
Beautiful. Both in single PR. Thanks for the closure on intended future support! Also it's nice to see the effort getting some utility as you guys work so fast! :) 

Long live TensorFlow!

PR #563, opened 2019-05-12
JA : translation r2/guide/autograph.ipynb

AseiSugiyama:
Translate https://github.com/tensorflow/docs/blob/master/site/en/r2/guide/autograph.ipynb


AseiSugiyama:
@googlebot I signed it 👍

PR #568, opened 2019-05-15
Suggestion to update keras.ipynb by adding tensorflow_datasets installation

ymodak:
Since tensorflow_datasets is a standalone module and users are required to install it separately in their local session (not using colab) the current notebook will throw error complaining tensorflow_datasets module not found.
By adding tensorflow_datasets installation to existing notebook users will be able to execute the code as is without additional module installation.


lamberta:
Thanks!

PR #570, opened 2019-05-16
#28762 [tensorflow/docs]Translate variables.md from English to Chinese.

Mr-Nineteen:


kuri-leo:
To correct my mistake: path should be `docs/site/zh-cn/r2/guide/variables.md` , also updated to suggested changes.

Mr-Nineteen:
@kuri-leo
OK，thx.

PR #595, opened 2019-05-23
Fix MirroredStrategy reduce bug

fsx950223:
MirroredStrategy reduce doesn't have axis parameter


lamberta:
Thanks!
Please also update [site/en/r2/tutorials/distribute/training_loops.ipynb](https://github.com/tensorflow/docs/blob/master/site/en/r2/tutorials/distribute/training_loops.ipynb), if applicable.

fsx950223:
> Thanks for the PR. the axis argument has actually been added after 1.13, so it's present in the nightly and we wanted to illustrate the latest API.
> Can we instead modify the tutorial to install tf-nightly instead?
> 
> `!pip install tf_nightly`

changed

PR #496, opened 2019-04-17
#27435 Fix Verify installation scripts for eager execution and aliases

timoveldt:
See [#27435](https://github.com/tensorflow/tensorflow/issues/27435). Removed tf.enable_eager_execution(). Change tf.random_normal to tf.random.normal

PR #596, opened 2019-05-24
Fix verification code

Rholais:
Module `tensorflow` has no attribute `random_normal`


lamberta:
Thank you, Rholais.
This is a duplicate of https://github.com/tensorflow/docs/pull/496 and we're waiting on TF 2.0 before we merge and publish.

PR #598, opened 2019-05-25
Fix typo in _image_classification.ipynb

katoosky:
I found typo and fix it when I translate this document to Japanese.


lamberta:
Thank you, @katoosky.
But please be aware that notebooks that start with an underscore are *not* published on tensorflow.org and considered a work-in-progress. Something to consider if you're translating documents.

katoosky:
Thank you.
Oh, Okay.
I will translate other documents first and wait for this document is finished.

PR #608, opened 2019-05-30
Dense note paragraph --> H4 section

thedriftofwords:
My attempt at parsing the note might not be fully accurate.

### DIFF
--- a/site/en/r2/tutorials/distribute/training_loops.ipynb
+++ b/site/en/r2/tutorials/distribute/training_loops.ipynb
@@ -168,7 +168,7 @@
         "id": "5mVuLZhbem8d"
       },
       "source": [
-        "How does `tf.distribute.MirroredStrategy` strategy work? \n",
+        "How does `tf.distribute.MirroredStrategy` strategy work?\n",
         "\n",
         "*   All the variables and the model graph is replicated on the replicas.\n",
         "*   Input is evenly distributed across the replicas.\n",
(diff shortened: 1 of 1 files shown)


lamberta:
Thanks, Crystal! Looping in Priya to confirm

site/en/r2/tutorials/distribute/training_loops.ipynb
```diff
@@ -702,7 +702,22 @@
...
+        "### Tracking training loss across replicas\n",
+        "\n",
+        "As a general rule, you should use `tf.keras.Metrics` to track per-sample values and avoid values that have been aggregated within a replica. Loss scaling with `tf.keras.Metrics` accounts for the number of replicas in sync.\n",
```
yashk2810: Maybe put this in a "Note: As a general rule, you should use `tf.keras.Metrics` to track per-sample values and avoid values that have been aggregated within a replica. Loss scaling with `tf.keras.Metrics` accounts for the number of replicas in sync.". Rest of it can be outside the note.

guptapriya:
@anj-s can you review this change as your recently added this description? thanks

site/en/r2/tutorials/distribute/training_loops.ipynb
anj-s: we can remove this line "Loss scaling with `tf.keras.Metrics` accounts for the number of replicas in sync" since it makes the note confusing to the reader.

site/en/r2/tutorials/distribute/training_loops.ipynb
```diff
@@ -702,7 +702,22 @@
...
+        "\n",
+        "We do *not* recommend using `tf.metrics.Mean` to track the training loss across different replicas, because it does not account for the number of replicas in\n",
```
anj-s: "because it does not account for the number of replicas" can be replaced with "because of the loss scaling computation that we carry out" or something to that effect.

site/en/r2/tutorials/distribute/training_loops.ipynb
```diff
@@ -702,7 +702,22 @@
...
+        "sync.\n",
+        "\n",
+        "For example, if you run a training job with the following characteristics:\n",
+        "* Two replicas\n",
+        "* Two samples are processed on each replica\n",
+        "* Resulting loss values: [2,  3] and [4,  5] on each replica\n",
+        "* Global batch size = 4\n",
+        "\n",
+        "Loss calculated with `tf.keras.Metrics` is scaled by an additional factor that is equal to the number of replicas in sync. With loss scaling, you calculate the per-sample value of loss on each replica by adding the loss values, and then dividing by the global batch size. In this case: `(2 + 3) / 4 = 1.25` and `(4 + 5) / 4 = 2.25`. \n",
```
anj-s: It makes more sense for "Loss calculated with `tf.keras.Metrics` is scaled by an additional factor that is equal to the number of replicas in sync." to be the last line in this section.

yashk2810:
I am pulling this in and will make the changes, since this seems like an important change for readability.

thedriftofwords:
@yashk2810 Thanks for taking care of that!

PR #631, opened 2019-06-04
broken link fix

anestisdotpy:
Fix to [29352](https://github.com/tensorflow/tensorflow/issues/29352).

Seems like in TF version 2 the link is not clickable.


anestisdotpy:
No problem. Thanks for the approval!

PR #637, opened 2019-06-04
KO: translation for ko/alpha/tutorials/text/unicode.ipynb ..

CamilaKim:
KO: translation for ko/alpha/tutorials/text/unicode.ipynb ..


lamberta:
Thanks. It looks like you're still using the ".local" email address for your pull request commit, which is the reason the CLA bot isn't recognizing you.

PR #638, opened 2019-06-05
Add options for CPU usage in the build process

I-Hong:
This pr is to add an option to choose the usage of CPU.


gunan:
While this may be a useful piece of information, I feel like this is simply a bazel flag. I am not sure this is the right place to mention it.
Maybe we can reference bazel command line reference for different options?
https://docs.bazel.build/versions/master/command-line-reference.html

I-Hong:
There is `memory option` added here, and both of memory and CPU are related to hardware resource.  I am not sure. What do you think?

PR #641, opened 2019-06-05
Update GPU installation instructions to reflect the fact that cuda-10…

MadcowD:
… apt package installs a different driver

I've updated GPU installation instructions to reflect the fact that cuda-10 apt package installs a different driver (418) than the one specified in the tutorial (410). This can lead to damaged configurations with two drivers installed simultaneously :O


MadcowD:
I signed it! @googlebot

tfboyd:
This seems fine, e.g. LGTM from me.  When I did the instructions it likely worked because only 410 was out.  If cuda10.0 install installs the driver then maybe we don't even need the driver install step.  That can be a future PR if some wants to test. I recall having issues, but things are changing very quickly.

PR #648, opened 2019-06-07
Update code snippets in ja load_data/images from en version

lamberta:
`./tools/nb_code_sync.py site/ja/tutorials/load_data/images.ipynb`


lamberta:
Need to debug this but will do it another time :)

PR #649, opened 2019-06-08
edited regex to ignore code comments in auto-linking

anestisdotpy:
This is a fix to [29339](https://github.com/tensorflow/tensorflow/issues/29339).

My tests are highlighted in the following screenshot:

Screenshot 2019-06-08 at 01 41 52

### DIFF
--- a/tools/tensorflow_docs/api_generator/parser.py
+++ b/tools/tensorflow_docs/api_generator/parser.py
@@ -223,9 +223,8 @@ class ReferenceResolver(object):
 
     for line in string.splitlines():
       if not line.startswith('#'):
-        line_to_replace = line
-        line = re.sub(AUTO_REFERENCE_RE, sloppy_one_ref, line)
-        string = string.replace(line_to_replace, line)
+        changed_line = re.sub(AUTO_REFERENCE_RE, sloppy_one_ref, line)
+        string = string.replace(line, changed_line)
 
     return string
 
--- a/tools/tensorflow_docs/api_generator/parser.py
+++ b/tools/tensorflow_docs/api_generator/parser.py
@@ -115,7 +115,7 @@ def _get_raw_docstring(py_object):
   else:
     return ''
 
-AUTO_REFERENCE_RE = re.compile(r'^(?!#.*$).*`([a-zA-Z0-9_.]+?)`')
+AUTO_REFERENCE_RE = re.compile(r'(^(?!#.*$).*)`([a-zA-Z0-9_.]+?)`', re.MULTILINE)
 
 
 class ReferenceResolver(object):
@@ -307,12 +307,12 @@ class ReferenceResolver(object):
 
   def _one_ref(self, match, relative_path_to_root):
     """Return a link for a single "`tf.symbol`" reference."""
-    string = match.group(1)
+    string = match.group(2)
     link_text = string
 
     if string.startswith('tensorflow::'):
       # C++ symbol
-      return self._cc_link(string, link_text, relative_path_to_root)
+      return match.group(1) + self._cc_link(string, link_text, relative_path_to_root)
 
     is_python = False
     for py_module_name in self._py_module_names:
@@ -321,7 +321,7 @@ class ReferenceResolver(object):
         break
 
     if is_python:  # Python symbol
-      return self.python_link(
+      return match.group(1) + self.python_link(
           link_text, string, relative_path_to_root, code_ref=True)
 
     # Error!
--- a/tools/tensorflow_docs/api_generator/parser.py
+++ b/tools/tensorflow_docs/api_generator/parser.py
@@ -115,7 +115,7 @@ def _get_raw_docstring(py_object):
   else:
     return ''
 
-AUTO_REFERENCE_RE = re.compile(r'`([a-zA-Z0-9_.]+?)`')
+AUTO_REFERENCE_RE = re.compile(r'^(?!#.*$).*`([a-zA-Z0-9_.]+?)`')
 
 
 class ReferenceResolver(object):
--- a/tools/tensorflow_docs/api_generator/parser.py
+++ b/tools/tensorflow_docs/api_generator/parser.py
@@ -115,7 +115,7 @@ def _get_raw_docstring(py_object):
   else:
     return ''
 
-AUTO_REFERENCE_RE = re.compile(r'(^(?!#.*$).*)`([a-zA-Z0-9_.]+?)`', re.MULTILINE)
+AUTO_REFERENCE_RE = re.compile(r'`([a-zA-Z0-9_.]+?)`')
 
 
 class ReferenceResolver(object):
@@ -221,7 +221,11 @@ class ReferenceResolver(object):
       except TFDocsError:
         return match.group(0)
 
-    string = re.sub(AUTO_REFERENCE_RE, sloppy_one_ref, string)
+    for line in string.splitlines():
+      if not line.startswith('#'):
+        line_to_replace = line
+        line = re.sub(AUTO_REFERENCE_RE, sloppy_one_ref, line)
+        string = string.replace(line_to_replace, line)
 
     return string
 
@@ -307,12 +311,12 @@ class ReferenceResolver(object):
 
   def _one_ref(self, match, relative_path_to_root):
     """Return a link for a single "`tf.symbol`" reference."""
-    string = match.group(2)
+    string = match.group(1)
     link_text = string
 
     if string.startswith('tensorflow::'):
       # C++ symbol
-      return match.group(1) + self._cc_link(string, link_text, relative_path_to_root)
+      return self._cc_link(string, link_text, relative_path_to_root)
 
     is_python = False
     for py_module_name in self._py_module_names:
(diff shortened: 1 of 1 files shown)


tools/tensorflow_docs/api_generator/parser.py
```diff
@@ -115,7 +115,7 @@ def _get_raw_docstring(py_object):
   else:
     return ''
 
-AUTO_REFERENCE_RE = re.compile(r'`([a-zA-Z0-9_.]+?)`')
+AUTO_REFERENCE_RE = re.compile(r'^(?!#.*$).*`([a-zA-Z0-9_.]+?)`')
```
yashk2810: This regex pattern will match only the first line and if the docstring is multiple lines, then we will need re.MULTILINE.

Also, it will match the entire line in the docstring if it finds what it is looking for. So we need to catch that because of the re.sub(...) in `replace_references` below.

These code changes should fix it.

AUTO_REFERENCE_RE = re.compile(r'(^(?!#.*$).*)`([a-zA-Z0-9_.]+?)`', re.MULTILINE)

Then,

On line 310: do `string = match.group(2)`

On line 315: do `return match.group(1) + self._cc_link(string, link_text, relative_path_to_root)`

On line 324: do `return match.group(1) + self.python_link(link_text, string, relative_path_to_root, code_ref=True)`

yashk2810:
Please test the changes you have made by running generate2.py here (https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/docs/generate2.py). The instructions on how to run it are in the file.

tools/tensorflow_docs/api_generator/parser.py
anestisdotpy: Thanks for the correction. I am quite new to this so I did not take these into account 😆 
However, now it does not link the `tf symbols` in the actual text.
I will try to fix it!

anestisdotpy:
> Please test the changes you have made by running generate2.py here (https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/docs/generate2.py). The instructions on how to run it are in the file.

That's very useful. Thank you!

tools/tensorflow_docs/api_generator/parser.py
yashk2810: In the regex the backticks got highlighted. The change there is to use multiline and group the first match too.

anestisdotpy: Oh I can see now why this happened, I didn’t notice the highlighted bit. I will change my commit once I get access to my computer. 
Thanks for the review!

anestisdotpy: Hi @yashk2810 . Your regex is not working for me, I don't know if I am doing something wrong.
I used `AUTO_REFERENCE_RE = re.compile(r'(^(?!#.$).)'([a-zA-Z0-9_.]+?)'', re.MULTILINE)` (I am putting the back-ticked ones as normal so the whole code line is highlighted) but when I run `generate2.py` and consider `TextLineDataset.md` then the `tf.symbols` outside the code blocks are not linked. I made the changes on lines 310, 315 & 324. Did you test it?


What do you think of the following regex? I tested it and it looks okay, with the changes you suggested on lines 310, 315 & 324:

`AUTO_REFERENCE_RE = re.compile(r'(^(?!#).*)'([a-zA-Z0-9_.]+?)'', re.MULTILINE)`

anestisdotpy: Hi @yashk2810. Your suggested fixes still don't work for me when I test it with `generate2.py`.  I used AUTO_REFERENCE_RE = re.compile(r'(^(?!#.$).)'([a-zA-Z0-9_.]+?)'', re.MULTILINE) and the suggested changes on lines 310,315 & 324 (I put the backticked ones as normal in the regex so the whole code line is highlighted) but they do not put ref links in the `tf.symbols` outside the code blocks. Did you test it, and if so, does it work for you?

I believe the problem is that in the original way of how this is done, the `tf.symbols` get matched without any other text before or after. In my commit, the regex matches the whole line and if 2 `tf.symbols` are in a line, only the latter will be changed. For this reason, I came up with a way that puts a "guard" condition, and then it does the changes in the original way if the condition is satisfied. However, I am not sure how good that is in terms of efficiency. I tested it with `generate2.py` and it looks good. What is your opinion?

```python
# Without any changes on lines 310,315 & 324
AUTO_REFERENCE_RE = re.compile(r'`([a-zA-Z0-9_.]+?)`')
AUTO_REFERENCE_RE_GUARD = re.compile(r'^(?!#.*$).*')
...
...
...

# In `def replace_references(...)` add the following right below `def sloppy_one_ref(...)`
for line in string.splitlines():
  if AUTO_REFERENCE_RE_GUARD.match(line):
    line_to_replace = line
    line = re.sub(AUTO_REFERENCE_RE, sloppy_one_ref, line)
    string = string.replace(line_to_replace, line)
			
return string

```

yashk2810: The regex I suggested was:
```python
AUTO_REFERENCE_RE = re.compile(r'(^(?!#.*$).*)`([a-zA-Z0-9_.]+?)`', re.MULTILINE)
```
Sorry, if it messed up due to the markdown highlighting.

This will match the entire line but with 2 groups. match.group(1) will be the text and match.group(2) will match the tf.symbol.


See:

```python
In [2]: import re

In [3]: AUTO_REFERENCE_RE = re.compile(r'(^(?!#.*$).*)`([a-zA-Z0-9_.]+?)`', re.MULTILINE)

In [4]: with open('temp.txt', 'r') as f:
   ...:     content = f.read()

In [5]: content
Out[5]: 'Maps `map_func` across the elements of this dataset.\n\nThis transformation applies `map_func` to each element of this dataset, and\nreturns a new dataset containing the transformed elements, in the same\norder as they appeared in the input.\n\nFor example:\n\n```python\na = Dataset.range(1, 6)  # ==> [ 1, 2, 3, 4, 5 ]\n\na.map(lambda x: x + 1)  # ==> [ 2, 3, 4, 5, 6 ]\n```\n\nThe input signature of `map_func` is determined by the structure of each\nelement in this dataset. For example:\n\n```python\n# NOTE: The following examples use `{ ... }` to represent the\n# contents of a dataset.\n# Each element is a `tf.Tensor` object.\na = { 1, 2, 3, 4, 5 }\n# `map_func` takes a single argument of type `tf.Tensor` with the same\n# shape and dtype.\nresult = a.map(lambda x: ...)\n\n# Each element is a tuple containing two `tf.Tensor` objects.\nb = { (1, "foo"), (2, "bar"), (3, "baz") }\n# `map_func` takes two arguments of type `tf.Tensor`.\nresult = b.map(lambda x_int, y_str: ...)\n\n# Each element is a dictionary mapping strings to `tf.Tensor` objects.\nc = { {"a": 1, "b": "foo"}, {"a": 2, "b": "bar"}, {"a": 3, "b": "baz"} }\n# `map_func` takes a single argument of type `dict` with the same keys as\n# the elements.\nresult = c.map(lambda d: ...)\n```\n\nThe value or values returned by `map_func` determine the structure of each\nelement in the returned dataset.\n'

In [6]: list(AUTO_REFERENCE_RE.finditer(content))
Out[6]:
[<_sre.SRE_Match object; span=(0, 15), match='Maps `map_func`'>,
 <_sre.SRE_Match object; span=(54, 92), match='This transformation applies `map_func`'>,
 <_sre.SRE_Match object; span=(366, 399), match='The input signature of `map_func`'>,
 <_sre.SRE_Match object; span=(1228, 1270), match='The value or values returned by `map_func`'>]

In [8]: [i.group(1) for i in list(AUTO_REFERENCE_RE.finditer(content))]
Out[8]:
['Maps ',
 'This transformation applies ',
 'The input signature of ',
 'The value or values returned by ']

In [9]: [i.group(2) for i in list(AUTO_REFERENCE_RE.finditer(content))]
Out[9]: ['map_func', 'map_func', 'map_func', 'map_func']
```

anestisdotpy: Still doesn't work for me however, I am adding a new commit with your suggestions.
Thank you!

yashk2810: Seems to be working for me. I just ran the script with your changes. Where is it not working?

anestisdotpy: As I mentioned above, only the last `tf.symbol` match in a line gets linked.
An example from `TextLineDataset.md`:


Screenshot 2019-06-10 at 00 19 03

yashk2810: Right, I didn't notice that. The AUTO_REFERENCE_RE_GUARD change looks good then.

anestisdotpy: You are right, that would definitely work. I didn't think about it.

yashk2810: Awesome, send in the revised commit and I'll approve it :)

anestisdotpy: Done. Do I have to update my branch as well? It is a couple of commits behind the docs master. As I mentioned earlier, I am quite new to all this. 
Thank you!

yashk2810: > Do I have to update my branch as well?
Nope.

yashk2810:
Thanks!

anestisdotpy:
> Thanks!

No problem! Thanks for your help and the review.

PR #658, opened 2019-06-10
Update transformer.ipynb

ppriyank:
original multiplication of mask with negative infinity will zero out the valid tokens (mask equals one for valid tokens, or True) and invalid tokens (whose masks = 0, or False), will be saved from negative infinity.

### DIFF
--- a/site/en/r2/tutorials/text/transformer.ipynb
+++ b/site/en/r2/tutorials/text/transformer.ipynb
@@ -704,7 +704,7 @@
         "\n",
         "For example, consider that `Q` and `K` have a mean of 0 and variance of 1. Their matrix multiplication will have a mean of 0 and variance of `dk`. Hence, *square root of `dk`* is used for scaling (and not any other number) because the matmul of `Q` and `K` should have a mean of 0 and variance of 1, so that we get a gentler softmax.\n",
         "\n",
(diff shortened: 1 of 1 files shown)


ppriyank:
I signed it!

yashk2810:
> valid tokens (mask equals one for valid tokens, or True) and invalid tokens (whose masks = 0, or False), will be saved from negative infinity.

That's the whole point. We need to mask the future values (which is 1) and not mask the 0 values (which are valid).

If there is any more need to clarify that, please add some text so that it becomes more clear :)

ppriyank:
> > valid tokens (mask equals one for valid tokens, or True) and invalid tokens (whose masks = 0, or False), will be saved from negative infinity.
> 
> That's the whole point. We need to mask the future values (which is 1) and not mask the 0 values (which are valid).
> 
> If there is any more need to clarify that, please add some text so that it becomes more clear :)

As seen in the code, mask=1 refers to the valid token I  guess,  can you verify this once?

yashk2810:
Yes, if you see the example masks that are shown (https://www.tensorflow.org/beta/tutorials/text/transformer#masking), you will see that 1 (is invalid) and 0's are valid.

ppriyank:
Understood, thanks. 
Apologies for the confusion.

PR #664, opened 2019-06-11
possible typo in guide/data.md

jamiekang:
This PR is to report a possible typo or confusing sentence.
In guide/data.md, I found this:
"A data source constructs a Dataset from data stored in memory in one ore more files."

'ore ': should be 'or'
files : traditionally files reside in disk. So does this mean the memory has a file system like a RAM disk?


rickiepark:
@jamiekang Thank you for contributing. 

Dataset class can read bunch of files into memory(tensor).
I guess the author try to explain data flow(files->memory->dataset object).
They are not related with ram disk. 

Please make PR in site/ko with translation. :)
Thanks.

PR #672, opened 2019-06-13
Rename variable 'steps' to 'steps_per_epoch'

texmex76:
The variable 'steps' should be called 'steps_per_epoch'. Probably overlooked when running the notebook since 'steps_per_epoch' had been defined before.


texmex76:
I signed it!

PR #693, opened 2019-06-17
fix typo

gkutiel:


rickiepark:
Hi @gkutiel 

Are you using Python 2.7?
In Python 3.x, `/` is true division. So `3/2` ouputs `1.5`.

In site/en/tutorials/sequences/text_generation.ipynb,
`/` and `//` makes same results. But for Python 3 users, `//` is better.

In site/en/r2/tutorials/text/text_generation.ipynb and site/ko/beta/tutorials/text/text_generation.ipynb, 
`examples_per_epoch` is not used.
You can remove the line `examples_per_epoch = len(text)//seq_length`

Thanks. :)

PR #715, opened 2019-06-23
fixed a broken link

therahulkumar:


therahulkumar:
I signed it!

lamberta:
Thanks, @therahulkumar. Your link is correct but this section looks out-of-date and should really point to https://www.tensorflow.org/resources
I'll make the update here.

PR #719, opened 2019-06-27
Correct tf.read_file() to tf.io.read_file()

djones:
Otherwise you will get this error when running the examples:

> module 'tensorflow' has no attribute 'read_file'


lamberta:
Thanks, David!

PR #739, opened 2019-07-02
Added Houston TensorFlow & Applied AI/ML {by HCCUG.org™}

val-moreign:
Per Billy's instructions, have added Houston TensorFlow & Applied AI/ML {by HCCUG.org™} to this page


val-moreign:
I signed it!

val-moreign:
Thanks for taking care of this so quickly, Billy!

PR #740, opened 2019-07-02
the first reviewer

diogoarm:
I added my username to the list of reviewers. @lamberta I suggest renaming the folder "pt" to "pt-br" or even create a new folder to Brazilian Portuguese translations.

### DIFF
--- a/site/pt/REVIEWERS
+++ b/site/pt/REVIEWERS
@@ -1,3 +1,4 @@
 # Add your GitHub username to get comment-tagged on all pull requests to
 # the site/pt directory. Remove your username whenever you'd like.
 #
+diogoarm


lamberta:
Regarding the nuances of pt-BR vs pt-PT, this is something I now know more than I ever wanted 😉

My *sense* is that once the Brazilian community gets steam, it will dominate—but we haven't really seen that, yet. And, in fact, some of the initial notebooks were contributed by someone in Portugal.

My understanding is that for *written* technical docs, 90%-100% is intelligible to the other side. Given the numbers of developers in Brazil, I'm leaning toward a switch to the `pt-br/` directory (and list), but haven't had the community participation to justify it, yet. But feel free to press the issue to me or the [docs@tensorflow.org](https://groups.google.com/a/tensorflow.org/forum/#!forum/docs) if this is something you want to run with.

We are planning some TensorFlow events in Brazil later this year and it would be a good opportunity for a translation sprint.

cc: @alextp

PR #754, opened 2019-07-05
Change of the deprecated syntax 'tf.VERSION' for 'tf.version.VERSION'

Vincentdecursay:
When starting the test after the installation of kera, this error show up: change of the deprecated syntax 'tf.VERSION' for 'tf.version.VERSION'


lamberta:
Thanks!

PR #761, opened 2019-07-07
Update images.ipynb

fabiofumarola:
fixed Attribute error in `tf.read_file`. It is `tf.io.read_file`


lamberta:
Confirmed CLA for username.

PR #769, opened 2019-07-09
Fixed minor typo

zanedurante:
Changed loss function described in text cell to match one used in code cell


zanedurante:
@lamberta Good catch! TF2 notebook is now updated also.

PR #781, opened 2019-07-11
Use softmax as activation for the dense layer

Terahezi:
It is necessary to use softmax as activation to convert raw prediction values to probability distribution.

### DIFF
--- a/site/en/tutorials/load_data/images.ipynb
+++ b/site/en/tutorials/load_data/images.ipynb
@@ -1012,7 +1012,7 @@
         "model = tf.keras.Sequential([\n",
         "  mobile_net,\n",
         "  tf.keras.layers.GlobalAveragePooling2D(),\n",
-        "  tf.keras.layers.Dense(len(label_names))])"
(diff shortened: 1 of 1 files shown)


Terahezi:
> https://cla.developers.google.com/

I signed it!

bhanu-sharma:
> https://cla.developers.google.com/

I signed it!

lamberta:
Can you also update the TF2 notebook: https://github.com/tensorflow/docs/blob/master/site/en/r2/tutorials/load_data/images.ipynb

yashk2810:
Also, can you update the loss in compile to use string losses?

Instead of `tf.keras.losses.sparse_categorical_crossentropy` use `'sparse_categorical_crossentropy'`

Terahezi:
> Also, can you update the loss in compile to use string losses?
> 
> Instead of `tf.keras.losses.sparse_categorical_crossentropy` use `'sparse_categorical_crossentropy'`

But what is the diff?

yashk2810:
There's no difference, its just a style preference 😀

PR #787, opened 2019-07-11
add tfx forum to forums page

adammichaelwood:


ewilderj:
Maybe for consistency use the same convention as the list below, ie. alpha order, and the list name as the link, so:

```
*   [tfx](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfx) -
    Discussion and collaboration around [TensorFlow Extended (TFX)](https://www.tensorflow.org/tfx/).
```

and put it in the list in alphabetical order?

PR #789, opened 2019-07-12
Generate test images properly

zyh911:
In every epoch the modified script can gererate different test images.


yashk2810:
Not generating new images is intentional so that you can clearly see the progression of the model.

Closing this for now.

zyh911:
But I have seen
# shuffling so that for every epoch a different image is generated
# to predict and display the progress of our model.
in the cript. Did I misunderstand?

zyh911:
But I have seen
```
shuffling so that for every epoch a different image is generated
to predict and display the progress of our model.
```
in the cript. Did I misunderstand?

PR #794, opened 2019-07-12
RU: basic_text_classification_with_tfhub - Minor fixes

0101011:


stabuev:
Thanks for fixes all corrections are reasonable.

lamberta:
Thanks for the fix, Andrew, and the review 🚀

For reviews, if you add a comment with "LGTM" or use the GitHub approval button, then tfdocsbot can automatically apply the `lgtm` label (which is easier for me to see on the pull request list).

Thanks!

PR #797, opened 2019-07-13
Add x-axis label for value plot

NicoleEO:
Suggestion to add class name index to x-axis

### DIFF
--- a/site/en/r2/tutorials/keras/basic_classification.ipynb
+++ b/site/en/r2/tutorials/keras/basic_classification.ipynb
@@ -772,7 +772,7 @@
         "def plot_value_array(i, predictions_array, true_label):\n",
         "  predictions_array, true_label = predictions_array[i], true_label[i]\n",
         "  plt.grid(False)\n",
-        "  plt.xticks([])\n",
+        "  plt.xticks(range(10))\n",
         "  plt.yticks([])\n",
         "  thisplot = plt.bar(range(10), predictions_array, color=\"#777777\")\n",
         "  plt.ylim([0, 1])\n",
--- a/site/en/r2/tutorials/keras/basic_classification.ipynb
+++ b/site/en/r2/tutorials/keras/basic_classification.ipynb
@@ -855,12 +855,13 @@
         "num_rows = 5\n",
         "num_cols = 3\n",
         "num_images = num_rows*num_cols\n",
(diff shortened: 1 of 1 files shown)


dynamicwebpaige:
This is excellent! Thanks for the contribution. 🙂 

cc: @lamberta

site/en/r2/tutorials/keras/basic_classification.ipynb
```diff
@@ -855,7 +855,7 @@
         "num_rows = 5\n",
         "num_cols = 3\n",
         "num_images = num_rows*num_cols\n",
-        "plt.figure(figsize=(2*2*num_cols, 2*num_rows))\n",
+        "plt.figure(figsize=(2*2*num_cols+1, 2*num_rows))\n",
```
yashk2810: Is there a reason to increase the width by 1?

NicoleEO: @yashk2810 One of the plot labels in the grid figure were getting slightly covered by the x tick labels added. I added a little room to make this label completely visible.

yashk2810: Can you try using this `plt.tight_layout()` before plt.show() and see if it improves the display?

NicoleEO: @yashk2810 so, I retested this and here are the results. I think I went with the manual adjustment since it spreads quite a bit with tight_layout. Let me know which you prefer and I can do another PR.
tight_layout:
image
manual:
image

yashk2810: The thing with doing it manually is, if you change something, you will need to update the figsize too. plt.tighy_layout() does that for you, so it would be nice to use that :)

>  I can do another PR.
You can modify it in this PR also.

NicoleEO: Good point. I will update on my PR at lunch. Thanks!

NicoleEO: @yashk2810 This has been updated.

yashk2810: Sweet thanks.

LGTM

PR #799, opened 2019-07-13
Test - Print index and name label w/

TahiriNadia:


lamberta:
Out of sync. Thanks, though

PR #800, opened 2019-07-13
x label in plot prediction

TahiriNadia:
Visualize clearly the score prediction in correlation w/ image (in left)


lamberta:
Out of sync. Thanks, though

PR #802, opened 2019-07-13
:heavy_minus_sign: rm__future__ no needed

TahiriNadia:
rm `from __future__ import absolute_import, division, print_function, unicode_literals`


lamberta:
Thanks, but we still use these to run and test the notebooks on Python2---we're not done with it, yet 😃

PR #806, opened 2019-07-14
Corrected typo and ommission

Silas-Asamoah:
In the "Overview" section, "about" was missing in describing the data content.
In the "Converting Data into Tensors", object was wrongly spelt


Silas-Asamoah:
I signed it!

PR #825, opened 2019-07-16
Add `save_weights` in HDF5 to save_and_restore_models.ipynb and other

8bitmp3:
Suggestions for the Keras Save and Restore Models tutorial (TensorFlow 2.0 Beta1):

- Add a `save_weights` **in HDF5** section: the TensorFlow default `.ckpt` vs Keras `.h5` weights files were discussed during one of the TF2 community testing calls, sometimes it may create confusion, so maybe we should emphasize how to switch between the two more often in various tutorials, just in case
- Add other minor fixes, suggestions, UX improvements in line with the Google Docs style guide

@lamberta @MarkDaoust @yashk2810 this is the first time I used Colab to make changes, hopefully it didn't break anything important, as Jsons are very delicate

### DIFF
--- a/site/en/r2/tutorials/keras/save_and_restore_models.ipynb
+++ b/site/en/r2/tutorials/keras/save_and_restore_models.ipynb
@@ -562,9 +562,7 @@
       "source": [
         "## Manually save weights\n",
         "\n",
-        "Above you saw how to load the weights into a model.\n",
-        "\n",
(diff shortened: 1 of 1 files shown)


lamberta:
Thanks, Olzhas.

There's some nice changes here, but my initial feeling is that we shouldn't add the HDF5 section to the tutorial, but belongs in the guide: https://www.tensorflow.org/beta/guide/keras/saving_and_serializing#whole-model_saving

Using the default seems like the way to go, unless there's a reason to save out to h5 (but this doesn't say why you would). But it would be fine to say that h5 is also an option and then link off to the guide.

8bitmp3:
@lamberta thanks Billy. Done as you asked + a few minor changes. Downloaded the notebook from Colab and replaced the older version, so hopefully it didn't break anything.

PR #833, opened 2019-07-19
Function tf.random_uniform not available

abviv:
Found more instances of un-available function. Revert to tf.random.uniform.


lamberta:
Hi Abishek,

There's a bunch in the TF1 docs and not worth the churn for translations in changing them. In TF1, `tf.random_uniform` is an alias of [tf.random.uniform](https://www.tensorflow.org/api_docs/python/tf/random/uniform) and it's fine to use. But if you see this in the [TF2 beta docs](https://www.tensorflow.org/beta), that's an error.

I've filed an internal bug to fix the redirects since the `tf.random_uniform` page should redirect to `tf.random.uniform`.

Thanks

PR #838, opened 2019-07-19
Fixes link in tf.keras.callbacks

brunapinos:
Solves part of the issue [#29950](https://github.com/tensorflow/tensorflow/issues/29950), that raises some not found links in tf.keras callbacks.


lamberta:
Thanks, but these docs are generated and tensorflow.org isn't published from this branch.

@MarkDaoust will need to take a look at the API generator.

PR #848, opened 2019-07-22
style mis-match

bhanu-sharma:
use variable instead of the instance variable.


bhanu-sharma:
I would like to understand the idea of using class variables instead of the instance variables while writing the model subclasses. Would it not be better if instance variables are used inside the class definition instead of class variables? 

If so, i will make these changes in this PR.

Thanks.

PR #858, opened 2019-07-23
make inline comment match running text

patzm:
The ordering of the spatial dimensions is mismatching in the running text and the inline python comment. Which one is correct? I think the most common convention is `batch_size x height x width x channels` if `channels_first=False`. See the [TensorFlow Hub documentation](https://www.tensorflow.org/hub/common_signatures/images).


lamberta:
Thanks!

PR #871, opened 2019-07-26
Decorate with @tf.function

Ryandry1st:
Implement graph-based acceleration for improved performance. Using google Colab this improved epoch times from ~ 30s / epoch to ~3.6s/ epoch. See associated issue : https://github.com/tensorflow/tensorflow/issues/31013#issue-472671053

### DIFF
--- a/site/en/r2/tutorials/generative/cvae.ipynb
+++ b/site/en/r2/tutorials/generative/cvae.ipynb
@@ -304,17 +304,14 @@
         "      eps = tf.random.normal(shape=(100, self.latent_dim))\n",
         "    return self.decode(eps, apply_sigmoid=True)\n",
         "\n",
-        "  @tf.function \n"
         "  def encode(self, x):\n",
         "    mean, logvar = tf.split(self.inference_net(x), num_or_size_splits=2, axis=1)\n",
         "    return mean, logvar\n",
         "\n",
-        "  @tf.function \n"
         "  def reparameterize(self, mean, logvar):\n",
         "    eps = tf.random.normal(shape=mean.shape)\n",
         "    return eps * tf.exp(logvar * .5) + mean\n",
         "\n",
-        "  @tf.function \n"
         "  def decode(self, z, apply_sigmoid=False):\n",
         "    logits = self.generative_net(z)\n",
         "    if apply_sigmoid:\n",
@@ -357,7 +354,6 @@
       "source": [
         "optimizer = tf.keras.optimizers.Adam(1e-4)\n",
         "\n",
-        "@tf.function \n"
         "def log_normal_pdf(sample, mean, logvar, raxis=1):\n",
         "  log2pi = tf.math.log(2. * np.pi)\n",
         "  return tf.reduce_sum(\n",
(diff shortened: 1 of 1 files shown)


Ryandry1st:
I signed it!

yashk2810:
As `tf.function` is recursively applied, you only need it at the top-level function.

Ryandry1st:
> As `tf.function` is recursively applied, you only need it at the top-level function.

I would agree but there is not a top-level function being made/called here because each function is being used at different times. Perhaps a number of the functions could be put under a single one, but there are at least three different function calls (compute loss, compute gradients, apply gradients), so each of them must have the decorator. 

I can update to have the minimum, but that does not really lend itself to readability, especially considering the embedded functions still need to be created so that they could be used as a graph function (as opposed to expecting eager execution/native python).

site/en/r2/tutorials/generative/cvae.ipynb
```diff
@@ -370,11 +376,13 @@
         "  logqz_x = log_normal_pdf(z, mean, logvar)\n",
         "  return -tf.reduce_mean(logpx_z + logpz - logqz_x)\n",
         "\n",
+        "@tf.function \n"
         "def compute_gradients(model, x):\n",
         "  with tf.GradientTape() as tape:\n",
         "    loss = compute_loss(model, x)\n",
         "  return tape.gradient(loss, model.trainable_variables), loss\n",
         "\n",
+        "@tf.function \n"
```
yashk2810: tf.function on compute_gradients is all that is necessary to provide the speedup.

Also, can you merge apply_gradients with compute_gradients so that there is only one tf.function required?

Ryandry1st: So compute_loss() does not have any performance gain from @tf.function? because it is called separately in the training process.

I tested how the recursiveness of autograph works at this colab https://colab.research.google.com/drive/16zrshCFCVB-b4zIF_Z1GtN8YahuSbAib 

and it can be seen that the speed up only occurs if the called function is decorated. So by leaving the decorator off of the loss calculation, it will reduce performance.

I further tested the actual tutorial code and found that leaving @tf.function off from the compute_loss() slows the time by about 0.1s/epoch. Perhaps not the most important, but still useful, especially if the network/loss function were expanded or slower hardware being used.

Finally, I tested it where only the loss, compute gradient, and apply gradient functions were decorated. This increased per epoch training time by 0.05s /epoch.

All together the proposed singular decorator led to training time of 3.88s/epoch in colab, which is about 7% decreased performance. 

As to combining the functions, this actually improved performance dramatically, cutting off about 0.4s/epoch so I will include this change

yashk2810: I just ran it with only one tf.function decorator on top of compute_and_apply_gradients, and it took 3.2 seconds per epoch. For the first epoch it takes longer because autograph traces the code to create the graph and also other initialization stuff happens. But after that its pretty fast.

Is this different from what you observed?

Ryandry1st: Yes, I usually disregard the first epoch or two to ensure consistency. If you only use the one tf.function I get about 3.22s/epoch for the first 10 epochs (it increases after that). If I include the other decorators I average about 3.1s/epoch in the first 10 epochs. I think this is just due to the compute_loss function call that is being run each epoch. Not necessarily important, but it isn't negligible either, especially if the loss function were more intensive.

yashk2810: SGTM

PR #872, opened 2019-07-26
fixed pickle error in the data loading code

VishwaPrabhakar:
the error due to pickling of object solved now in the basic text classification example by keras.

### DIFF
--- a/site/en/tutorials/keras/basic_text_classification.ipynb
+++ b/site/en/tutorials/keras/basic_text_classification.ipynb
@@ -1,728 +1,1028 @@
 {
-  "cells": [
-    {
-      "cell_type": "markdown",
-      "metadata": {
-        "colab_type": "text",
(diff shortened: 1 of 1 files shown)


VishwaPrabhakar:
I signed it! the CLA.

lamberta:
Your diff was hard to read. If you save from Colab it's easier for us to see your change.

Also, we're about the switch the site over to TF2-first so the focus is now on docs in the `r2/` directory.

yashk2810:
Isn't this fixed in tf-nightly?

There are instructions at the top of the notebook.

```
# keras.datasets.imdb is broken in 1.13 and 1.14, by np 1.16.3
!pip install tf_nightly
```

MarkDaoust:
Thanks @VishwaPrabhakar, but it looks like the instructions  on the notebook are pretty clear.

We don't try to support old versions on master.

Thanks.

PR #879, opened 2019-07-28
Add Python to markdown

8bitmp3:
Great guide! To enhance UX, maybe we can have Python identifiers in markdown for syntax highlighting in code blocks. Cheers


MarkDaoust:
These `python` have no effect on the actual site output, but it doesn't hurt to have them, and they may help in come viewers.

> @MarkDaoust WDYT?

+1

But not blocking for this change.

PR #891, opened 2019-07-31
Revise wrong concatenation at the positional encoding

juheeuu:
I think the positional encoding concatenation part is wrong, because it makes the positional encoding vector as like this, 
PE_pos = [ sin(*) ,...,sin(*),cos(*),...cos(*)] 
But It should be looks like this.
image

Thus, I changed the code. 
Please take care :)

### DIFF
--- a/site/en/r2/tutorials/text/transformer.ipynb
+++ b/site/en/r2/tutorials/text/transformer.ipynb
@@ -454,7 +454,10 @@
         "                          np.arange(d_model)[np.newaxis, :],\n",
         "                          d_model)\n",
         "  \n",
+        "  # apply sin to even indices in the array; 2i\n",
         "  angle_rads[:, 0::2] = np.sin(angle_rads[:, 0::2])\n",
+        "  \n",
+        "  # apply cos to odd indices in the array; 2i+1\n",
         "  angle_rads[:, 1::2] = np.cos(angle_rads[:, 1::2])\n",
         "    \n",
         "  pos_encoding = angle_rads[np.newaxis, ...]\n",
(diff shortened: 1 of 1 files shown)


site/en/r2/tutorials/text/transformer.ipynb
```diff
@@ -778,7 +805,7 @@
         "id": "JPmbr6F1C-v_"
       },
       "source": [
-        "Each multi-head attention block gets three inputs; Q (query), K (key), V (value). These are put through linear (Dense) layers and split up into multiple heads. \n",
+        "![대체 텍스트](https://)Each multi-head attention block gets three inputs; Q (query), K (key), V (value). These are put through linear (Dense) layers and split up into multiple heads. \n",
```
lamberta: remove

site/en/r2/tutorials/text/transformer.ipynb
```diff
@@ -396,56 +428,49 @@
...
-        "  # apply sin to even indices in the array; 2i\n",
-        "  sines = np.sin(angle_rads[:, 0::2])\n",
-        "  \n",
-        "  # apply cos to odd indices in the array; 2i+1\n",
-        "  cosines = np.cos(angle_rads[:, 1::2])\n",
-        "  \n",
-        "  pos_encoding = np.concatenate([sines, cosines], axis=-1)\n",
-        "  \n",
-        "  pos_encoding = pos_encoding[np.newaxis, ...]\n",
+        "  angle_rads[:, 0::2] = np.sin(angle_rads[:, 0::2])\n",
```
lamberta: keep comments

yashk2810:
I don't think it really matters. Can you share the accuracy, loss and prediction changes that you got after doing this?

site/en/r2/tutorials/text/transformer.ipynb
juheeuu: I've removed it!

site/en/r2/tutorials/text/transformer.ipynb
juheeuu: Okay.

juheeuu:
@yashk2810 
I agree with you, It doesn't affect the result. 
The Loss & Accuracy is.. 
`original : Epoch 20 Loss 0.5687 Accuracy 0.3424` 
` changed: Epoch 20 Loss 0.5577 Accuracy 0.3422` 

But I still think it has to be changed because the code is different with description. 
It can be confusing for first-time students like me. 
Therefore, I think the code should be modified or a more detailed description added.

yashk2810:
LGTM

lamberta:
Thanks!

PR #894, opened 2019-08-01
Update groups.md: Added Tensorflow Community Pune

JBAhire:


JBAhire:
@googlebot I signed it!

lamberta:
Thank you for submission.
Unfortunately, we can't add this group because it does not include 'TensorFlow' in the group name. This is the minimum requirement to ensure it's a TensorFlow user group and not a generic AI/ML community group.

PR #898, opened 2019-08-02
update plot methods in Make predictions section

harshildarji:
The current implementation of methods `plot_image` and `plot_value_array` works well when the size of  `predictions_array` is greater than `1`. But this implementation gives error while performing prediction on a single image which is not at `0`<sup>th</sup> position in `test_images`.

---
For example, the current implementation of single image prediction uses the first image from the `test_images`,
```python
# Grab an image from the test dataset
img = test_images[0]

print(img.shape)
```
which works well while plotting the value array graph (_the first argument of the `plot_value_array` method is the index of the image being used for the test purpose, which in this case is `0`_),
```python
plot_value_array(0, predictions_single, test_labels)
plt.xticks(range(10), class_names, rotation=45)
plt.show()
```
But, if we use an image other than the first one (let's say the second image from the `test_image`),
```python
# Grab an image from the test dataset
img = test_images[1]

print(img.shape)
```
and try to plot the value array graph using,
```python
plot_value_array(1, predictions_single, test_labels)
plt.xticks(range(10), class_names, rotation=45)
plt.show()
```
it will throw following error:
```python
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-57-358696683377> in <module>
----> 1 plot_value_array(1, predictions_single, test_labels)
      2 plt.xticks(range(10), class_names, rotation=45)
      3 plt.show()

<ipython-input-48-290e8ffd30e7> in plot_value_array(i, predictions_array, true_label)
     19 
     20 def plot_value_array(i, predictions_array, true_label):
---> 21   predictions_array, true_label = predictions_array[i], true_label[i]
     22   plt.grid(False)
     23   plt.xticks([])

IndexError: index 1 is out of bounds for axis 0 with size 1
```
---
In this PR, I update both the methods and also its references to work perfectly even in case of using any test image to perform single image prediction.

Now, if you perform single image prediction (_in this case, the second image from `test_images`_), it will give you the following graph instead of throwing an error.

image

Let me know if further explanation (_or proof_) is required.
Thank you.

Best regards,
Harshil Darji

### DIFF
--- a/site/en/r2/tutorials/keras/basic_classification.ipynb
+++ b/site/en/r2/tutorials/keras/basic_classification.ipynb
@@ -124,7 +124,7 @@
         "id": "FbVhjPpzn6BM"
       },
       "source": [
-        "This guide trains a neural network model to classify images of clothing, like sneakers and shirts. It's okay if you don't understand all the details; this is a fast-paced overview of a complete TensorFlow program with the details explained as we go.\n",
+        "This guide trains a neural network model to classify images of clothing, like sneakers and shirts. It's okay if you don't understand all the details; this is a fast-paced overview of a complete TensorFlow program with the details explained as you go.\n",
         "\n",
         "This guide uses [tf.keras](https://www.tensorflow.org/guide/keras), a high-level API to build and train models in TensorFlow."
       ]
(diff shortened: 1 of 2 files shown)


harshildarji:
@googlebot I signed it!

lamberta:
Hi Harshil,

With TF2 around the corner, we're no longer working TF1 docs. Can you see how this applies to here: https://github.com/tensorflow/docs/blob/master/site/en/r2/tutorials/keras/basic_classification.ipynb

Also, try downloading the ipynb notebook from Colab to make the diff easier to review.
Thanks!

harshildarji:
Hi @lamberta,
I checked the TF2 version of the `basic_classification.ipynb` and found that it also had the same problem. So I have also updated that file to work well with any image from the `test_images` to perform single image prediction without any errors. 
Thank you.

PR #909, opened 2019-08-07
Fix quickstart tutorial to reset metrics every epoch

ljdang:
Update the quickstart advanced tutorial to reset the loss and accuracy metrics every epoch.


ljdang:
@googlebot I signed it!

PR #912, opened 2019-08-08
Update transformer.ipynb

alphamuth:
This error is coming as eager mode is needed to use numpy with tensor data
`Tensor' object has no attribute 'numpy'`
so to solve this error we need to execute this below line 
tf.enable_eager_execution()


lamberta:
TF 2 is eager by default so this isn't needed. The notebook runs fine in Colab: https://colab.research.google.com/github/tensorflow/docs/blob/master/site/en/r2/tutorials/text/transformer.ipynb
Maybe something wrong with your setup?

PR #913, opened 2019-08-08
Update distribute_strategy.ipynb

romeokienzler:
added double-spaces at the end in enumeration in order to trigger line breakes when rendered on https://www.tensorflow.org/guide/distribute_strategy


lamberta:
I don't think that's the issue. It needs a newline before the list and looks like it's already been fixed in the source.

We're focused on TF2 docs since the site is about to switch over. Check out the TF2 version of this guide: https://www.tensorflow.org/beta/guide/distribute_strategy

romeokienzler:
Ok, thanks a lot for taking time to review my PR. No worries. TF2 version of that document looks really great, btw...

PR #914, opened 2019-08-09
Fix link to keras guide

lamberta:


lamberta:
Fixes https://github.com/tensorflow/tensorflow/issues/31225

PR #919, opened 2019-08-11
Added Neural Machine Translation Tutorial

rish-16:
The tutorial covers tokenization, cleaning, pre-processing, and configuring TPUs using a parallelly-distributed training `strategy`. The whole tutorial is written in `tf.keras`.


rish-16:
@googlebot I signed it!

lamberta:
Thanks for the submission!
We're only working on TF2 docs now so this would need some changes before consideration into the official docs: switch to TF2 (eager by default, no Sessions, etc). And, as a tutorial, it needs more explanations before code cells.
After switching to TF2, you may consider submitting it to the [community](https://github.com/tensorflow/examples/tree/master/community/en) section of the tensorflow/examples repo.

PR #922, opened 2019-08-12
Fix typo in site/en/r2/guide/keras/rnn.ipynb 

alvaroabascar:
"The returned states can be used resume the RNN execution" -> "The returned states can be used **to** resume the RNN execution"


alvaroabascar:
@googlebot I signed it!

PR #932, opened 2019-08-14
Fix python3 numpy format bug

fsx950223:


yashk2810:
I think this has been fixed recently.

PR #935, opened 2019-08-15
fix typo in keras functional API notebook

dellison:
"None that the batch size..." --> "Note that the batch size..."


dellison:
I've signed the CLA now.

dellison:
@googlebot I signed it!

PR #936, opened 2019-08-15
pix2pix: Always same test_dataset picture 

crypdra:
- The shuffle operation had been applied on the train_dataset for a second time.
- The .take(1) operation will always return the same element, as it does not iterate the dataset

I fixed the dataset shuffle operation and replaced the .take(1) problem by defining an iterator. In order for the iterator to work for more than 107 epochs I defined a repeat operation.

### DIFF
--- a/site/en/r2/tutorials/generative/pix2pix.ipynb
+++ b/site/en/r2/tutorials/generative/pix2pix.ipynb
@@ -403,10 +403,10 @@
         "# shuffling so that for every epoch a different image is generated\n",
         "# to predict and display the progress of our model.\n",
         "test_dataset = test_dataset.shuffle(BUFFER_SIZE)\n",
-		  "test_dataset = test_dataset.repeat()\n",
+        "test_dataset = test_dataset.repeat()\n",
         "test_dataset = test_dataset.map(load_image_test)\n",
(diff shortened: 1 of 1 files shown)


crypdra:
@googlebot I signed it!

yashk2810:
The image is kept same because it allows us to see the growth of the model. You can compare the picture from epoch 30 to epoch 15 and see the progress. If the image keeps changing, its hard to see the progress.

crypdra:
> The image is kept same because it allows us to see the growth of the model. You can compare the picture from epoch 30 to epoch 15 and see the progress. If the image keeps changing, its hard to see the progress.

But there's a comment in the code which explicitly says, that the code is supposed to change the test picture on every epoch:
```
# shuffling so that for every epoch a different image is generated
# to predict and display the progress of our model.
```

yashk2810:
> But there's a comment in the code which explicitly says, that the code is supposed to change the test image on every epoch:

Sorry about that. I fixed it here (https://github.com/tensorflow/docs/commit/3b597c2492e626cf5c79ba4a97f19e0fb1ff0383)

lamberta:
Sounds like this is resolved. Thanks, @crypdra, for raising the issue, and thanks, @yashk2810, for the fix.

PR #938, opened 2019-08-17
KO: update ko/beta/tutorials/keras/feature_columns.ipynb

rickiepark:
Sync with en file.
Thanks.


site/ko/beta/tutorials/keras/feature_columns.ipynb
```diff
@@ -701,7 +701,8 @@
         "\n",
         "model.compile(optimizer='adam',\n",
         "              loss='binary_crossentropy',\n",
-        "              metrics=['accuracy'])\n",
+        "              metrics=['accuracy'],\n",
+        "              run_eagerly=True)\n",
```
lamberta: @yashk2810 Why does this need to be set in TF2? What does it mean if it's eager by default?

lamberta:
Thanks, Haesun, for the update.
Now trying to understand the reason we need `run_eagerly=True` here

site/ko/beta/tutorials/keras/feature_columns.ipynb
yashk2810: compile and fit runs in function mode by default for better performance out of the box. 

Eager by default means that outside fit and compile you can still debug and print stuff in eager mode.

lamberta: Thanks, Yash.
I get the performance benefits, but the notebook breaks without this—is that WAI?

yashk2810: Yup, I have a bug filed against that and it was suggested to use run_eagerly there.

lamberta: This has been fixed in `tf-nightly` and en/ notebook updated in https://github.com/tensorflow/docs/commit/ac65fe769fedf44a3afe4592f56182831cc72913

@rickiepark Can you update this notebook to use `tf-nightly-2.0-preview` and remove `run_eagerly=True`. Thanks!

rickiepark: @lamberta I just pushed a commit for it. Thanks!

PR #943, opened 2019-08-19
Use r2.0rc

lamberta:
Change links to github master branch to 2.0rc:
```
$ find . -type f -name "*.ipynb" -print0 | xargs -0 sed -i '' -e 's|tensorflow/docs/blob/master|tensorflow/docs/blob/r2.0rc|g'
```

Also rebased from master.


yashk2810:
Approving it, but its difficult to check in github. Let's bring it in and then check :)

lamberta:
CLAs verified

yashk2810:
Let the chaos begin

PR #946, opened 2019-08-20
Update time_series.ipynb - scaling methods

rverschuren:
**Page concerned :** [Time series forecasting guide](https://www.tensorflow.org/beta/tutorials/text/time_series)

I noticed that the name of the method used for scaling data did not match its description.

#### Here are two methods for scaling data : 

- **Normalization** typically means rescales the values into a range of [0,1]. 
- **Standardization** typically means rescales data to have a mean of 0 and a standard deviation of 1 (unit variance).

This can be confusing when you observe data and notice that the values are not between [0,1]. 

[Source](https://towardsdatascience.com/normalization-vs-standardization-quantitative-analysis-a91e8a79cebf)

### DIFF
--- a/site/en/tutorials/structured_data/time_series.ipynb
+++ b/site/en/tutorials/structured_data/time_series.ipynb
@@ -328,7 +328,8 @@
         "id": "-eFckdUUHWmT"
       },
       "source": [
-        "It is important to normalize features before training a neural network. A common way to do so is by subtracting the mean and dividing by the standard deviation of each feature."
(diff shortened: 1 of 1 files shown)


MarkDaoust:
Hi @rverschuren 

Sorry about the delay here, It looks like you got a merge conflict.

These can be messy to resolve in notebooks. it might be easier to resolve it by just reapplying your change to master, and creating a new PR.

Thanks.

rverschuren:
Moved to #1252

PR #950, opened 2019-08-21
TF in Spanish /es folder

crispindev:
Adding additional folders and files.


crispindev:
> Is this necessary without content, yet?
> Uppercase filename to `README.md`

I'll include some translated files in the folders and the corrections. The idea is to give some type of structure for the event assistants so they are not lost on how and where to create the content.

PR #970, opened 2019-08-30
Fix wording in Load images tutorial

rogerthat94:


lamberta:
Thanks. The r1/ docs have been archived since we're switching the site over to TF2. The TF2 version of this file is here: https://github.com/tensorflow/docs/blob/master/site/en/tutorials/load_data/images.ipynb

If you make this change to the TF2 doc, please align with the Google dev docs style guide: https://developers.google.com/style/person

PR #982, opened 2019-09-06
Updated typo

jvishnuvardhan:
When we change input `test_images` from the currently selected image to other in the [tutorial](https://www.tensorflow.org/tutorials/keras/basic_classification), then the prediction will change and reports correctly. But, the static text at the end of the tutorial `And the model predicts a label of 2.` will not change.

So I changed the static text from `And the model predicts a label of 2.` to `And the model predicts a label as expected.` 
Thanks

### DIFF
--- a/site/en/tutorials/keras/basic_classification.ipynb
+++ b/site/en/tutorials/keras/basic_classification.ipynb
@@ -987,7 +987,7 @@
         "colab_type": "text"
       },
       "source": [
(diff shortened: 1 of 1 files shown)


jvishnuvardhan:
@lamberta I have updated the file in master as you suggested. https://github.com/tensorflow/docs/pull/987

lamberta:
Thank you! Closing this PR for that one

PR #984, opened 2019-09-08
removed duplicate plt.axis basic_regression.ipynb

minoring:
Remove duplicate plt.axis at basic_regression.ipynb


minoring:
@lamberta I made new PR #988 
Thank you!

PR #992, opened 2019-09-10
model inconsistency

minoring:
The model is not consistent with the comment above.
I was wondering which one was incorrect; between the comment and the model.
As my assumption, I changed the comment.
(I changed and tested the model as the comment mentioned, and got about 50% accuracy on the validation set.)

### DIFF
--- a/site/en/tutorials/images/classification.ipynb
+++ b/site/en/tutorials/images/classification.ipynb
@@ -1167,7 +1167,7 @@
         "id": "1Ba2LjtkVrXy"
       },
       "source": [
-        "Here, you apply dropout to first and last max pool layers and to a fully connected layer that has 512 output units. 30% of the first and last max pool layer, and 10% of fully connected layer output units, are randomly set to zero during each training epoch."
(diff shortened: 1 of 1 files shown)


site/en/tutorials/images/classification.ipynb
```diff
@@ -1167,7 +1167,7 @@
         "id": "1Ba2LjtkVrXy"
```
yashk2810: <p>How about this?</p><p> </p><p>"Applying dropout will randomly set 20% of the neurons to zero during each training epoch. This helps to avoid overfitting on the training dataset."</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/tensorflow/docs/pull/992/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='86'/>

minoring: Thank you for your kind review :)
I updated the comment as your advice which I think is more natural.

PR #996, opened 2019-09-12
KO : edit ko/guide/variables.md

dogemad:
Corrected to the word "레이어" instead of the word "층".

I think it will convey better meaning when you use the word "레이어" rather than the word "층".

It is true that "층" is "레이어" in Korean, but I think it is more natural to use "레이어" considering context and word specificity.

thanks!


rickiepark:
Thanks @dogemad 
But we translate 'layer' as '층'. Please refer http://bit.ly/tf-docs-translation-glossary.

dogemad:
@JKIsaacLee @Choiuijin1125 @rickiepark thanks for your review! i close it!

PR #998, opened 2019-09-13
Update linear.ipynb

thMike:
Make sure it has the file same structure as https://github.com/tensorflow/models/tree/master/official/r1/wide_deep/ 
https://github.com/tensorflow/tensorflow/issues/32497


thMike:
@googlebot I signed it!

PR #1021, opened 2019-09-17
Fix bug: call tf.random.set_random_seed instead of tf.random.set_seed

jontingvold:
tf.random.set_seed(13) produces 

AttributeError: module 'tensorflow._api.v1.random' has no attribute 'set_seed'

In new versions of Tensorflow this is renamed to

tf.random.set_random_seed(13)

### DIFF
--- a/site/en/r2/tutorials/text/time_series.ipynb
+++ b/site/en/r2/tutorials/text/time_series.ipynb
@@ -254,7 +254,7 @@
       },
       "outputs": [],
       "source": [
-        "tf.random.set_seed(13)"
+        "tf.random.set_random_seed(13)"
       ]
     },
     {


lamberta:
Hi Jon,
We're currently switching the site over to TF2-first and this file is now here: https://github.com/tensorflow/docs/blob/master/site/en/tutorials/structured_data/time_series.ipynb
Can take a look and make your change to the `master` branch (if needed)?
Thanks

jontingvold:
Looks like it changed the other way around. It was already changed so it is compatible with 2.0rc, while tf.random.set_random_seed is correct in v1.

My mistake. Issue closed.

But it is confusing when the notebooks are using v2.0, when it is only a release candidate.

PR #1041, opened 2019-09-24
Fix FGSM tutorial

Catminusminus:
## Overview
This PR will fix https://github.com/tensorflow/tensorflow/issues/32715.

## Page concerned
https://www.tensorflow.org/beta/tutorials/generative/adversarial_fgsm

## Description
The implementaion of FGSM in the doc is incorrect, because the `create_adversarial_pattern` function, which must take the image and the label of the image, takes the image and the predicted probability of the image.
Thus, by this PR, the `create_adversarial_pattern` function is going to take `label` instead of `image_probs` (`== model.predict(image)`), where the `label` is the one-hot vector of the label of the image.

### DIFF
--- a/site/en/tutorials/generative/adversarial_fgsm.ipynb
+++ b/site/en/tutorials/generative/adversarial_fgsm.ipynb
@@ -289,7 +289,7 @@
       "source": [
         "# Get the input label of the image.\n",
         "label = np.zeros(image_probs.shape)\n",
-        "label[0][np.argmax(image_probs)] = 1\n",
+        "label[0][208] = 1\n",
         "\n",
         "perturbations = create_adversarial_pattern(image, label)\n",
         "plt.imshow(perturbations[0])"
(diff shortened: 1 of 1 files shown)


Catminusminus:
@googlebot I signed it!

piyush-kgp:
@Catminusminus , it still looks broken to me. You should not be using `image_probs` to create `label`. You should be using the ground truth.

Catminusminus:
@piyush-kgp Thank you for your reviewing! You are right. I used `image_probs` to create `label` assuming that the predicted label of model is correct (== Labrador_retriever), because there is not the `encode_label` function which transforms a string label to `label` vector.
Now, I use the ground truth `208` by eye-grep of `https://storage.googleapis.com/download.tensorflow.org/data/imagenet_class_index.json`. Should I download it again in the notebook by using `urllib`? What would you recommend I do?

PR #1044, opened 2019-09-25
Revise tf.summary section to eager.ipynb with tf.compat.v2.summary

yunabe:
This pull request revives "Summaries and TensorBoard" section to https://www.tensorflow.org/guide/eager with `tf.compat.v2.summary`.

See https://github.com/tensorflow/tensorflow/issues/32588 for the background.

### DIFF
--- a/site/en/guide/eager.ipynb
+++ b/site/en/guide/eager.ipynb
@@ -935,7 +935,8 @@
         "understanding, debugging and optimizing the model training process. It uses\n",
         "summary events that are written while executing the program.\n",
         "\n",
-        "You can use `tf.summary` of Tensorflow 2.x (`tf.compat.v2.summary` in Tensorflow 1.x) to record summaries of variable in eager execution. For example, to record summaries of `loss` once every 100 global steps:"
+        "You can use `tf.summary` to record summaries of variable in eager execution.\n",
+        "For example, to record summaries of `loss` once every 100 training steps:"
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


yunabe:
> We try to keep a clean separation between tf1 docs and tf2 docs.
> So I'm going to merge this without the if-else's because this is the file we're going to publish when tf2 launches. The tf1 docs have been banished to the r1 directory.

I see. That means people using tf1.15 won't be able to find tf1 version of https://www.tensorflow.org/guide after tf2.0 release. It's good to know. I thought tf1 and tf2 versions of guides will be maintained while people are migrating to tf2.

Anyway, I think the current live doc (https://www.tensorflow.org/guide/eager#summaries_and_tensorboard) is not very good as an example of execution because `get_or_create_global_step` and `record_summaries_every_n_global_steps` are something we will deprecate in eager execution era. Can I send a PR to `r2.0rc` to fix the live one?

MarkDaoust:
>  That means people using tf1.15 won't be able to find tf1 version of https://www.tensorflow.org/guide after tf2.0 release. 

That's true. There are arguments for both sides. 

> Can I send a PR to r2.0rc to fix the live one?

No. The update in this PR will go live on the site when TensorFlow2.0 launches. The site is frozen till then.

yunabe:
Thank you for the review and the comment, Mark.

My original intent of this PR was fixing the current doc for tf1 users, especially TF beginners like me, who is trying to learn eager mode (#32588). But it seems like we do not maintain the current live version until TF2 release. It was great to know. Thank you for the advices.

I will read not only for the live site but also `master` ipynb files until TF2 release.

PR #1047, opened 2019-09-26
Fix img-url in checkpoints.md

FKgk:
Fix img-url in checkpoints.md

I changed tensorflow-link-url, but reverted it


lamberta:
Thanks. I wouldn't spend too much effort in the `r1/` directory since it's in maintenance mode, but this is a worthwhile fix.

PR #1048, opened 2019-09-26
added tensorflow lagos to community groups

keleog:
I added the TensorFlow Lagos group information to the community groups.


keleog:
@googlebot I signed it!

PR #1054, opened 2019-09-30
Numeric error correction

greentec:
Test accuracy is 70.74%, not over 99%.


lamberta:
Thanks. You are correct, but I'm wondering if this is a regression or we should aim a little higher :) @yashk2810 @MarkDaoust

MarkDaoust:
Thanks, It looks like the update that switched it to use cifar10 forgot to update that line.

PR #1055, opened 2019-09-30
fix code

elizabethsiegle:
acc = history.history['acc'] val_acc = history.history['val_acc']

### DIFF
--- a/site/en/tutorials/images/classification.ipynb
+++ b/site/en/tutorials/images/classification.ipynb
@@ -678,8 +678,8 @@
       },
       "outputs": [],
       "source": [
-        "acc = history.history['accuracy']\n",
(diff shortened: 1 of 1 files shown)


lamberta:
Thanks, but this doesn't run in the TF2 version (see tfdocsbot link to your notebook in Colab).
TF2 version: https://colab.sandbox.google.com/github/tensorflow/docs/blob/master/site/en/tutorials/images/classification.ipynb

Seems similiar https://github.com/tensorflow/docs/pull/1056, which makes me wonder if there's a bad link somewhere? We recently switched the site over, though the tensorflow.org page seems to have the correct links: https://www.tensorflow.org/tutorials/images/classification

PR #1056, opened 2019-09-30
modify pyplot keys

LittileZiggs:
the keys of the plt are wrong, they should not be history.history['accuracy'] and history.history['val_accuracy'].  They should be history.history['acc'] and history.history['val_acc'] instead.

### DIFF
--- a/site/en/tutorials/images/cnn.ipynb
+++ b/site/en/tutorials/images/cnn.ipynb
@@ -364,8 +364,8 @@
       },
       "outputs": [],
       "source": [
-        "plt.plot(history.history['accuracy'], label='accuracy')\n",
-        "plt.plot(history.history['val_accuracy'], label = 'val_accuracy')\n",
+        "plt.plot(history.history['acc'], label='accuracy')\n",
+        "plt.plot(history.history['val_acc'], label = 'val_accuracy')\n",
         "plt.xlabel('Epoch')\n",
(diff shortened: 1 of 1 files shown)


LittileZiggs:
> @googlebot I signed it!

lamberta:
Thanks, but your notebook breaks in Colab (see link from tfdocsbot above). Make sure you are using TF2.
The current TF2 notebook works: https://colab.research.google.com/github/tensorflow/docs/blob/master/site/en/tutorials/images/cnn.ipynb

LittileZiggs:
> Thanks, but your notebook breaks in Colab (see link from tfdocsbot above). Make sure you are using TF2.
> The current TF2 notebook works: https://colab.research.google.com/github/tensorflow/docs/blob/master/site/en/tutorials/images/cnn.ipynb

Thank you. I previously run the codes in TF1.14, keys should be 'acc' and 'val_acc'. After I upgraded to TF2, keys should be altered to 'accuracy' and 'val_accuracy', or it would crash. It seems that the docs are based on TF2 instead of TF1.14.

lamberta:
Thanks. Yes, we recently switched everything over to TF2-first 😀

TF 1.x docs are archived in the `r1/` directory (GitHub only): https://github.com/tensorflow/docs/tree/master/site/en/r1

PR #1058, opened 2019-09-30
Fix issues in text_classification keras tutorial

ab-10:
- `  print('Label:', train_label.numpy())` was causing `AttributeError: 'Tensor' object has no attribute 'numpy'`, setting `tf.enable_eager_execution()` fixes this.

- `train_data.output_shapes` was causing deprecation warning:
```
WARNING:tensorflow:From <ipython-input-13-5710cd302ddc>:6: DatasetV1.output_shapes (from tensorflow.python.data.ops.dataset_ops) is deprecated and will be removed in a future version.
Instructions for updating:
Use `tf.compat.v1.data.get_output_shapes(dataset)`.
```
I've updated uses of `output_shapes` as suggested by the warning.

### DIFF
--- a/site/en/tutorials/keras/text_classification.ipynb
+++ b/site/en/tutorials/keras/text_classification.ipynb
@@ -154,7 +154,8 @@
         "  %tensorflow_version 2.x\n",
         "except Exception:\n",
         "  pass\n",
-        "import tensorflow as tf"
+        "import tensorflow as tf",
+        "tf.enable_eager_execution()"
       ]
     },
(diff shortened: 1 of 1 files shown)


lamberta:
Thanks, but I think you may be running this in TF1 instead of TF2. This works: https://colab.research.google.com/github/tensorflow/docs/blob/master/site/en/tutorials/keras/text_classification.ipynb
(And eager is enabled by default in TF2 so no need to enable it.)

PR #1064, opened 2019-10-01
Fix broken Estimators link in site/en/guide/keras/overview

ahlaw:
Fixed broken estimators link on this page: [https://www.tensorflow.org/guide/keras/overview](https://www.tensorflow.org/guide/keras/overview)


ahlaw:
@googlebot I signed it!

PR #1067, opened 2019-10-01
Update time_series.ipynb

christophmeyer:
Minor correction in the text: target_size is the size of the label not the label itself.


MarkDaoust:
Or maybe upgrade it to use [`TimeSeriesGenerator`](https://www.tensorflow.org/api_docs/python/tf/keras/preprocessing/sequence/TimeseriesGenerator)

lamberta:

PR #1068, opened 2019-10-01
A typo

kernelizd:
Says the NN acieved 99%, actually its only 70%.


lamberta:
Thanks. Duplicate of https://github.com/tensorflow/docs/pull/1054

PR #1077, opened 2019-10-04
Stop inheriting tf.keras.Model just to compose layers

yunabe:
This PR fix rewrite unnecessary `tf.keras.Model` with `tf.keras.layers.Layer` in 4 tutorials.

https://github.com/tensorflow/tensorflow/issues/33006


lamberta:
Thanks, @yunabe.
Some of these seem reasonable, some seem like a style issue. Agree that we could explain why you would choose one over the other, but I'm concerned that offering more choice is adding complexity for users. What's the benefit of switching the example in the checkpoint.ipynb guide? Because it doesn't use `.fit`?

yashk2810:
From what I understand, if you inherit from `tf.keras.layers.Layer` then you can use the custom layer inside a Model.

But if you are creating a custom model, then you need to inherit from `tf.keras.Model` so that the intent is clear. It doesn't matter if you don't use `.fit` or other functionality associated with it.

So, in this case in `nmt_with_attention` for example, `Encoder` and `Decoder` are Models, but BahadanuAttention is a `Layer`.

Does that make this clear?

yunabe:
> So, in this case in `nmt_with_attention` for example, `Encoder` and `Decoder` are Models, but BahadanuAttention is a `Layer`.

I think this is a great example of what I want to be clarified in TF's tutorials (https://github.com/tensorflow/tensorflow/issues/33006). Does this mean `Encoder` and `Decoder` should inherit from `Model`, not `layers.Layer`, just for documentation purpuse?

yashk2810:
> just for documentation purpuse?

No, it should be always like that. Anything that is a model can be serialized and saved as a SavedModel and other advantages over a Layer. So the final thing that you train should be a Model so you can save the weights, etc.

yunabe:
First, what I want to clarify with https://github.com/tensorflow/tensorflow/issues/33006 is the scope of responsibilities of `keras.Model` and `keras.layers.Layer`. In other words, when we should inherit from `Model` and when we should inherit from `layers.Layer`.

>  So the final thing that you train should be a Model so you can save the weights, etc.

But we use `tf.train.Checkpoint` in the tutorial.
IMHO, we should keep tutorials as simple as possible (unless there is a rule like *always inherit tf.keras.Model*.)

If we want to implement `.save` on Encoder because we may use it in the future, why `BahadanuAttention` is a layer? Why don't you implement `.save` on `BahadanuAttention` too?

Also, `Encoder` and `Decoder` do not implement `.fit`, `.evaluate` and other utilities because it receives multiple inputs. It clearly violates [Liskov substitution principle
](https://en.wikipedia.org/wiki/Liskov_substitution_principle). Are they really subclasses of `Model`?

yunabe:
> Also, if you decide to train the model on multi-GPU's in the future, using distribution strategy with .fit and .compile is easier than doing it with a CTL.

I see. But `Encoder` and `Decoder` do not implement training utilities like `.fit`, `.evaluate`. Do we really want to recommend TF users to define these types of layers as subclasses of `tf.keras.Model`?

I understand there is a undocumented convention like

> subclass the final thing you want to train from a `tf.keras.Model`

in tutorials (regardless of whether they are trainable with `.fit` or not). I wonder TF team really recommends TF users to follow that convention.

MarkDaoust:
I think these changes are reasonable. I'll send a follow-up CL making the text clearer.

yunabe:
> One other feature provided by `keras.Model` (instead of `keras.layers.Layer`) is that in addition to tracking variables, a `keras.Model` also tracks its internal layers, making them easier to inspect.

https://github.com/keras-team/keras/releases/tag/2.3.0
keras says "Layers set as attributes of a Layer are now tracked".

Also, "Layers can be recursively nested" in custom_layers_and_models.ipynb sounds "layers are tracked" to a beginner of TF like me.

Thus, I still think I would feel some difficulty of understanding the difference and inconsistency in guidelines if I read them w/o prior knowledge.
But if TF team this is the best description, I'm happy with that.

Thank you for taking look into my feedback.

PR #1083, opened 2019-10-07
Update regression.ipynb

danshorstein:
add using get_dummies for turning categorical data into columns


danshorstein:
@googlebot I signed it!

lamberta:
Would rather not add more pandas, unless necessary.
Currently using it to read the CSV file---is this something we can use TFDS or TF-IO for?

PR #1086, opened 2019-10-07
Fix broken anchor link

jaden-young:
Internal link to `#Performance` should be `#performance`


jaden-young:
@googlebot I signed it!

PR #1090, opened 2019-10-08
JA: Translated site/ja/tutorials/text/word_embeddings.ipynb

masa-ita:


AseiSugiyama:
Sorry for the late review, however, this PR seems to be the translation of an outdated notebook. I guess the update of the original one makes itself better. So it would be great if we translate the notebook from upstream. How do you think?

masa-ita:
@AseiSugiyama san.  Thank you for your suggestion.
I didn't notice that the notebook has been revised after my PR.  I'm working on it.

masa-ita:
I made a mistake to rebase after the pull request, so I will close this and will make another PR.

PR #1109, opened 2019-10-16
Add pip uninstall before pip install

jlherren:
Otherwise an already installed version will not be installed over, causing
confusion, for example https://github.com/tensorflow/tensorflow/issues/7449

### DIFF
--- a/site/en/install/source.md
+++ b/site/en/install/source.md
@@ -296,8 +296,9 @@ these two configurations in the same source tree.
 The filename of the generated `.whl` file depends on the TensorFlow version and
 your platform. Use `pip install` to install the package, for example:
 
-<pre class="devsite-terminal prettyprint lang-bsh">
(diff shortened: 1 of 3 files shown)


jlherren:
@googlebot I signed it!

MarkDaoust:
Hi,

That's a really long, and old thread you linked to, can you summarize and explain the issue here a little more?

If you have a package installed, and install a different _version_ pip uninstalls the previous one.

It's only when you install a different package (live tensorflow vs tensorflow_gpu) that they can step on each-other, right? So I don't think this has the intended effect. With tensorflow 2.1+ there's no longer any need for the tensorflow_gpu package.

But feel free to reopen this if I have misunderstood something.

jlherren:
I think you misunderstood. My point was precisely when you reinstall the very *same* version again, which can be a very real scenario when someone builds tensorflow from source several times in a row, in an attempt to figure out which configuration works best. At least that's what happened to me and it wasn't obvious to me at first that what I just built wasn't being installed.

PR #1111, opened 2019-10-17
tf.data API performance optimization

AlexisBRENON:
Linked to tensorflow/tensorflow#33049
Last version of the new notebook can be seen here: [Colab notebook](https://colab.research.google.com/github/AlexisBRENON/docs/blob/AlexisBRENON/33049/input_pipeline/site/en/guide/data_performance.ipynb)

Tensorflowers:
- @MarkDaoust


jsimsa:
@lamberta could you please add @aaudiber as a reviewer

Andrew, this is a rewrite of the tf.data performance guide into a form that is colab runnable. Could you please do a thorough review? Thank you.

aaudiber:
Hi @AlexisBRENON,

The timing visualizations look really cool and I'm impressed by the code used to generate them! However, I'm concerned that they place some limitations on the doc:

1. They force us to use simulated data instead of more realistic examples, such as data from https://storage.googleapis.com/tf-datasets like we use in https://www.tensorflow.org/guide/data.
2. They require adding large blocks of non-trivial code which can distract users from the point of the doc.

I think writing code examples and giving timing information is enough to demonstrate that the optimizations really work. We could use diagrams similar to the visualizations to help explain topics, but those can be manually generated instead of interactively generated. That way, we can use realistic examples and avoid code complexity.

AlexisBRENON:
@aaudiber you're write, the "boilerplate" code to build this code (dataset generator, timing accumulation, py_function wrapper) become more clumsy that what I expected :roll_eyes: 

However, I fear that using realistic examples with, for example, a TFRecordDataset using data from `tf-dataset` will not allow to clearly exhibit execution time improvements...

So here is an idea:

- Make a first part of the guide with simpler boilerplate (just `time.sleep`, no measurements and graph plotting) to make sure that timing informations are coherent all along the guide (no network or CPU caching possible, no network latency or bandwidth involved, etc). We may still add some images (SVGs) generated with the current code to introduce the visualization. Moreover, these plots allow to understand some behaviors like the `interleave`, the `cache` or the vectorized `map`. 
- Keep the complicated code for plot generation and co. at the whole end of the guide, to just compare a "naive" pipeline with an optimized pipeline. This will allow anyone to re-use it to debug/monitor her own pipeline (in particular it exhibits some tricks to run in eager mode the data pipeline).

As I said to Mark in [the related issue](https://github.com/tensorflow/tensorflow/issues/33049), the "data" guide is already exhaustive about realistic examples. So I think that the "data performance" guide could be more technical.

Thanks for the feedback.

aaudiber:
Ok, I'm on board with your proposed changes. Lets keep the first parts of the guide as simple as possible code-wise, then have a section on visualizing timing at the end.

site/en/guide/data_performance.ipynb
```diff
@@ -0,0 +1,1142 @@
...
+    "As for the [naive approach](#The-naive-approach), here the times spent for loading, fetching, pre-processing (mapping) and training steps sum together for a single iteration."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "#### Parallel mapping\n",
+    "\n",
+    "Now, use the same pre-processing method but applying it in parallel on multiple samples."
```
aaudiber:
```suggestion
    "Now, use the same pre-processing function but apply it in parallel on multiple samples."
```

site/en/guide/data_performance.ipynb
```diff
@@ -0,0 +1,1142 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 14,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Execution time: 1.453609274001792\n"
```
aaudiber: when I run this, I get the same time for vectorized vs non-vectorized mapping. To show the performance, I think we need a non-artificial dataset with a large batch size and very cheap map function.

site/en/guide/data_performance.ipynb
```diff
@@ -0,0 +1,1142 @@
...
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "![Parallel mapping](./parallel_map.svg)\n",
+    "\n",
+    "Now, you can see on the plot that the pre-processing steps can overlap allowing to reduce the overall time for a single iteration."
```
aaudiber:
```suggestion
    "Now, you can see on the plot that the pre-processing steps overlap, reducing the overall time for a single iteration."
```

site/en/guide/data_performance.ipynb
```diff
@@ -0,0 +1,1142 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Caching\n",
+    "\n",
+    "The `tf.data.Dataset.cache` transformation can cache a dataset, either in memory or on local storage.\n",
+    "This will prevent some operations (like data loading and preparation) to be executed for each epoch."
```
aaudiber:
```suggestion
    "This will save some operations (like data loading and preparation) from being executed during each epoch."
```

site/en/guide/data_performance.ipynb
```diff
@@ -0,0 +1,1142 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "![Cached dataset](./cached_dataset.svg)\n",
+    "\n",
+    "When you cache a dataset, the transformations before the `cache` one (like the file loading and fetching) are executed only during the first epoch.\n",
+    "The next epochs will reuse the data persisted by the `cache` transformation.\n",
```
aaudiber:
```suggestion
    "The next epochs will reuse the data cached by the `cache` transformation.\n",
```

site/en/guide/data_performance.ipynb
```diff
@@ -0,0 +1,1142 @@
...
+    "maintain an internal buffer of elements. If the user-defined function passed\n",
+    "into the `map` transformation changes the size of the elements, then the\n",
+    "ordering of the map transformation and the transformations that buffer elements\n",
+    "affects the memory usage. In general, we recommend choosing the order that\n",
+    "results in lower memory footprint, unless different ordering is desirable for\n",
+    "performance.\n",
+    "\n",
+    "#### Caching partial computations\n",
+    "\n",
+    "As seen above, it is recommended to cache you dataset after the `map` transformation except if this transformation make the data too big to fit in memory.\n",
```
aaudiber:
```suggestion
    "It is recommended to cache the dataset after the `map` transformation except if this transformation makes the data too big to fit in memory.\n",
```

site/en/guide/data_performance.ipynb
```diff
@@ -0,0 +1,1142 @@
...
+     ]
+    }
+   ],
+   "source": [
+    "iterate(\n",
+    "    ArtificialDataset()\n",
+    "    .map(  # Apply time consuming operations before cache\n",
+    "        lambda x: factorial(8)\n",
+    "    ).cache(\n",
+    "    ).map(  # Apply memory consuming operations after cache\n",
```
aaudiber: since we already have an example of memory consuming vs time consuming operations below, I think we can simplify this by removing the second map.

site/en/guide/data_performance.ipynb
```diff
@@ -0,0 +1,1142 @@
...
+    "\n",
+    "As seen above, it is recommended to cache you dataset after the `map` transformation except if this transformation make the data too big to fit in memory.\n",
+    "A trade-off can be achieved if your mapped function can be split in two parts: a time consuming one and a memory consuming part.\n",
+    "In this case, you can chain your transformations like below:\n",
+    "\n",
+    "```python\n",
+    "dataset.map(time_consuming_mapping).cache().map(memory_consuming_mapping)\n",
+    "```\n",
+    "\n",
+    "This way, your time consuming part is only executed during the first epoch, while limiting your memory footprint avoiding big data chunks to be persisted in memory."
```
aaudiber:
```suggestion
    "This way, the time consuming part is only executed during the first epoch, and you avoid using too much cache space."
```

site/en/guide/data_performance.ipynb
```diff
@@ -0,0 +1,1142 @@
...
+    "Training time is simulated."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 4,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def iterate(ds, num_epochs=2):\n",
```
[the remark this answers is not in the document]
AlexisBRENON: `measure_training_time` sounds like the function will return the "training" time. Here the function does not return anything, and even do not "train" anything. That's why I choose a more generic term. Moreover, it is called in many places, and so a short name seems better (to not clutter simple calls).
What about:
- `run`: maybe to linked to the thread vocabulary
- `loop`: even more broad and unclear than `iterate`
- `benchmark`: linked to the performance evaluation vocabulary

site/en/guide/data_performance.ipynb
```diff
@@ -0,0 +1,1142 @@
...
+    "You can see that performing a training step require that:\n",
+    "- you load the data to the memory,\n",
+    "- you fetch a data entry,\n",
+    "- and then you use your data for training.\n",
+    "\n",
+    "However, in a naive synchronous implementation like here, while your pipeline is fetching the data, your model is sitting idle. \n",
+    "Conversely, while your model is training, the input pipeline is sitting idle.\n",
+    "The training step time is thus the sum of all, loading, fetching and training time.\n",
+    "\n",
+    "The next sections build on this input pipeline, illustrating best practices for designing performant TensorFlow input pipelines."
```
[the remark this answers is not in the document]
AlexisBRENON: I am not english native but it should be: "The next section**s** build" or "The next section build**s**" ?

site/en/guide/data_performance.ipynb
AlexisBRENON: Yes, I a little bit stuck with this part. My idea would be to divide all values of the tensor by a value which is long to compute (as a recurcive factorial or whatever). However, from the results here, I suppose that the value is kept in some kind of cache, hence not taking so much time to be computed for "scalar" version...
In my first experiment, I used a `sleep(1ms)` for the scalar case and a `sleep(batch_size ms)` for the vectorized one, which gave good result but not applicable in graph mode...
Maybe playing with loops can provide a acceptable result.

Why do you think that a non-artificial dataset would better show performance improvements?

AlexisBRENON:
Thanks for the review. I addressed most of your comments, asking some questions for others.
I am going to read it one more time, checking for consistency (open file, read data, etc.) all along the guide and fixing the images.
Is there a prefered way to manage the images. Currently they are saved in the `guide/` directory, without anything to avoid name clash with images of other guides. Should I make a dedicated directory for images? Add a prefix?

site/en/guide/data_performance.ipynb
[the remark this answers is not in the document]
AlexisBRENON: Just thought of `artificial_fit(...)`. It mirrors the `ArtificialDataset` naming scheme as well as linking to `tf.keras.Model.fit` which take similar arguments: the dataset, the number of epochs.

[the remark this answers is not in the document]
aaudiber: I like `benchmark`, since the purpose of the code is to benchmark the performance of the passed-in dataset

site/en/guide/data_performance.ipynb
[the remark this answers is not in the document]
aaudiber: Oh, it was correct to start with, disregard my comment

site/en/guide/data_performance.ipynb
aaudiber: Vectorization only has a noticeable impact when the overhead of scheduling map functions is large compared to the time it takes to run the map function. Scheduling a map function takes around 10 microseconds, so to see an impact we need to be iterating very fast, which isn't really possible when using the sleep-based artificial dataset.

We can do something like this:

```python
import tensorflow as tf
import time
tf.enable_eager_execution()

elements = 10000
batch_size = 1024
num_batches = elements/batch_size
ds = tf.data.Dataset.range(elements)

def f(x):
  return x+1

def vectorized_f(batch):
  return tf.map_fn(f, batch)

non_vectorized = ds.map(f).batch(batch_size)
vectorized = ds.batch(batch_size).map(vectorized_f)

def benchmark(dataset):
  start = time.time()
  # iterate using skip so that all iteration is done in c++.
  next(iter(dataset.skip(num_batches-1)))
  return time.time() - start

# warmup
benchmark(non_vectorized)
benchmark(vectorized)

print("Non-vectorized takes " + str(benchmark(non_vectorized)) + " seconds")
print("Vectorized takes " + str(benchmark(vectorized)) + " seconds")

# Non-vectorized takes 0.550505876541 seconds
# Vectorized takes 0.166946172714 seconds
```

aaudiber:
@AlexisBRENON I looked into it and the process for adding images to docs is somewhat complicated. Images need to be added to the tensorflow.org website first, then they can be linked from docs, (e.g. https://www.tensorflow.org/images/datasets_parallel_map.png). Once we've settled on the final images we want, I can add the images to the internal website. Then once the website re-publishes, we can merge this PR with links to those images.

@MarkDaoust can you confirm that this is the proper process?

site/en/guide/data_performance.ipynb
AlexisBRENON: Wow, this example exhibits an impresive 30% performance improvements! But it requires:
- to introduce a new dataset − even if easy to use, it creates a special case in the guide;
- to write a new benchmarking loop with special iteration mechanisms.

However, I did not manage to provide any example which work well with both, eager (`py_function`) and graph execution. I even try to use shorter `sleep` for dataset but, results are still unpredictable and not reproducible, as well as making the images less clear because "hidden" times (start of an epoch, various logic) are more/too visible compared to the measured steps...

So, I am a little bit stuck for the moment. If we choose an example similar to yours, maybe we have to introduce it more thoroughly, later in the guide (after "Caching").

A more general question. Is it required to use `tf.map_fn`? The `add` method already applies to all elements of the batch (thanks to broadcasting from what I understand).

AlexisBRENON:
I finally managed to have examples for scalar/vectorized. Improvement is not very big, but it is consistent accross runs (on my machine, scalar is always around 2.9s (±0.02s) while vectorized is always around 2.8s (±0.02s)).

I also split the guide, that I found a little bit long. This way there is the "Performance" guide, and the "Debug/monitor" guide. I don't know if it is really useful, but it reduces the clutter of the "Performance" one.

I think images are OK (clearly exhibiting the expected behavior).

I see that there is some conflicts. Should I rebase now or wait for approval and rebase just before merge?

site/en/guide/data_performance.ipynb
aaudiber: Here's a simplified example that shows over 10x speedup by using `tf.add` as you suggested. We see speedup even without introducing any special iteration mechanism. I think it's ok to introduce a new (very simple) dataset. I agree it should go at the end of the guide after we've gone over all the optimizations that apply to ArtificialDataset.

```
def benchmark(dataset):
  start = time.time()
  for elem in dataset:
    pass
  return time.time() - start

ds = tf.data.Dataset.range(10000)
non_vectorized = ds.map(lambda x: x+1).batch(256)
vectorized = ds.batch(256).map(lambda x: tf.add(x, 1))

print("Vectorized takes " + str(benchmark(vectorized)) + " seconds")
print("Non-vectorized takes " + str(benchmark(non_vectorized)) + " seconds")
```

AlexisBRENON: Updated with your example. I used `lambda x: x+1` for both examples, `tf.add` is not part of the TF2.0 API and make the syntax complicated.
Interestingly, adding the `unbatch` transformation afterward makes both datasets roughly on par.

MarkDaoust:
Hi,

What's the status of this PR?

Does it just need a final review and merge? Or are you still working on it?

AlexisBRENON:
I adressed the comments from Andrew. It needs a new (final?) review.
It still requires some work to integrate images to the main TF site, and thus update all image links to the new addresses.

site/en/guide/data_performance.ipynb
```diff
@@ -0,0 +1,766 @@
...
+    "colab_type": "text",
+    "id": "Tce3stUlHN0L"
+   },
+   "source": [
+    "##### Copyright 2019 The TensorFlow Authors."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
```
aaudiber: Use "Omit code cell output when saving this notebook". I think this should get all the execution counts down to 0.

site/en/guide/data_debug.ipynb
```diff
@@ -0,0 +1,923 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "colab_type": "text",
+    "id": "qFdPvlXBOdUN"
+   },
+   "source": [
+    "# Debug and monitor a tf.data.Dataset pipeline"
```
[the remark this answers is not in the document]
AlexisBRENON: Ok, no problem, I will move back the `ArtificialDataset` to the "Performance" page, as previously.

By the way, an advanced doc about using trace viewer would be nice! But it's for another PR.

site/en/guide/data_performance.ipynb
```diff
@@ -525,141 +525,165 @@
...
+   "cell_type": "code",
+   "execution_count": 19,
    "metadata": {},
    "outputs": [
     {
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "Execution time: 2.7948353739993763\n"
+      "WARNING:tensorflow:Entity <function <lambda> at 0x7fafec11b6a8> could not be transformed and will be executed as-is. Please report this to the AutoGraph team. When filing the bug, set the verbosity to 10 (on Linux, `export AUTOGRAPH_VERBOSITY=10`) and attach the full output. Cause: expected exactly one node node, found []\n",
```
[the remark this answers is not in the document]
AlexisBRENON: I ill investigate this.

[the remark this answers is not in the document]
AlexisBRENON: I think it is linked to the fact that cells are executed as interactive shell. Thus the lambda function `lambda x: x+1` does not expose its source code.

<details>
<summary>
Here is part of the traceback from Autograph (with verbosity set to 10):
</summary>

```
INFO:tensorflow:Converted call: <function <lambda> at 0x7fe65c0d09e0>
    args: (<tf.Tensor 'args_0:0' shape=() dtype=int64>,)
    kwargs: {}
[...]
INFO:tensorflow:Error transforming entity <function <lambda> at 0x7fe65c0d09e0>
  File "[...]/venv/lib/python3.7/site-packages/tensorflow_core/python/autograph/pyct/parser.py", line 53, in parse_entity
    original_source = inspect_utils.getimmediatesource(entity)
  File "[...]/venv/lib/python3.7/site-packages/tensorflow_core/python/autograph/pyct/inspect_utils.py", line 127, in getimmediatesource
    lines, lnum = inspect.findsource(obj)
  File "/usr/lib/python3.7/inspect.py", line 786, in findsource
    raise OSError('could not get source code')
OSError: could not get source code

During handling of the above exception, another exception occurred:

  File "[...]/venv/lib/python3.7/site-packages/tensorflow_core/python/autograph/impl/api.py", line 506, in converted_call
    converted_f = conversion.convert(target_entity, program_ctx)
  File "[...]/venv/lib/python3.7/site-packages/tensorflow_core/python/autograph/impl/conversion.py", line 322, in convert
    free_nonglobal_var_names)
  File "[...]/venv/lib/python3.7/site-packages/tensorflow_core/python/autograph/impl/conversion.py", line 240, in _convert_with_cache
    entity, program_ctx)
  File "[...]/venv/lib/python3.7/site-packages/tensorflow_core/python/autograph/impl/conversion.py", line 469, in convert_entity_to_ast
    nodes, name, entity_info = convert_func_to_ast(o, program_ctx)
  File "[...]/venv/lib/python3.7/site-packages/tensorflow_core/python/autograph/impl/conversion.py", line 630, in convert_func_to_ast
    node, source = parser.parse_entity(f, future_features=future_features)
  File "[...]/venv/lib/python3.7/site-packages/tensorflow_core/python/autograph/pyct/parser.py", line 61, in parse_entity
    ' @tf.autograph.do_not_convert. Original error: {}'.format(entity, e))
ValueError: Unable to locate the source code of <function <lambda> at 0x7fe65c0d09e0>. Note that functions defined in certain environments, like the interactive Python shell do not expose their source code. If that is the case, you should to define them in a .py source file. If you are certain the code is graph-compatible, wrap the call using @tf.autograph.do_not_convert. Original error: could not get source code

WARNING:tensorflow:Entity <function <lambda> at 0x7fe65c0d09e0> could not be transformed and will be executed as-is. Please report this to the AutoGraph team. When filing the bug, set the verbosity to 10 (on Linux, `export AUTOGRAPH_VERBOSITY=10`) and attach the full output. Cause: Unable to locate the source code of <function <lambda> at 0x7fe65c0d09e0>. Note that functions defined in certain environments, like the interactive Python shell do not expose their source code. If that is the case, you should to define them in a .py source file. If you are certain the code is graph-compatible, wrap the call using @tf.autograph.do_not_convert. Original error: could not get source code
```
</details>

Workaround is to declare an `increment` function.

site/en/guide/data_performance.ipynb
AlexisBRENON: This is a Colab feature, I use Jupyter. I cleared all cells output before saving, this should put `execution_count` to `null`, is it OK?

site/en/guide/_toc.yaml
```diff
@@ -52,7 +52,7 @@ toc:
 - heading: "Data input pipelines"
 - title: "tf.data"
   path: /guide/data
-- title: "Performance"
+- title: "Optimize pipeline performance"
```
lamberta: This has been renamed. Please rebase?

AlexisBRENON:
Rebased onto master. Keep the Markdown history.
Images are grouped inside a `guide/images/<guide_name>/` folder.
In Colab, I cannot tick the "Omit code cell output when saving this notebook" option (it is grayed out), but I saved the file before any execution, and with the left nav panel opened.

PR #1112, opened 2019-10-17
Minor fixes, as promised

thedriftofwords:


thedriftofwords:
The bulleted list renders correctly in Colab, but it currently looks like this in tensorflow.org:

> This dataset will have 4 features: * a boolean feature, False or True with equal probability * an integer feature uniformly randomly chosen from [0, 5] * a string feature generated from a string table by using the integer feature as an index * a float feature from a standard normal distribution

So the first commit simply adds a newline to fix that.

PR #1124, opened 2019-10-22
Fixed jupyter notebooks

npanpaliya:


site/en/tutorials/keras/text_classification_with_hub.ipynb
```diff
@@ -139,6 +139,7 @@
         "  pass\n",
         "import tensorflow as tf\n",
         "\n",
+        "!pip install tensorflow-hub\n",
```
wdirons: @npanpaliya , this example should pip install both tensorflow-datasets and tensorflow-hub as if they run this docker file first, they may not have either installed.

wdirons:
For reference @npanpaliya and I work on the same team. These issues were found using the TensorFlow supplied docker container with Jupyter installed. They don't have tensorflow-datasets or tensorflow-hub preinstalled. A similar issue was reported here: https://github.com/tensorflow/tensorflow/issues/33179

In the save_and_load notebook, `model.save('saved_model/my_model')` would fail because the saved_model directory did not exist.

site/en/tutorials/keras/text_classification_with_hub.ipynb
npanpaliya: Thanks @wdirons. I've added that now although when I ran this notebook it didn't complain for datasets here, it might have found it installed.

wdirons: Thank you @npanpaliya. If you created a new docker container and tried this notebook first, it would have complained about tensorflow-datasets not being installed, but if you ran the other notebook with your change to install tensorflow-datasets first, it would have already been installed.

PR #1129, opened 2019-10-24
Markdown list is not properly displayed

drozzy:
Added a newline before list so docs show up properly on 
https://www.tensorflow.org/guide/keras/masking_and_padding


drozzy:
@googlebot I signed it!

PR #1136, opened 2019-10-25
Fixed broken bookmarks

harmindergit:


lamberta:
Thanks. Where are you testing these? dashes-vs-underscores changes depending if you're viewing in GitHub, Colab, and tensorflow.org. We have some preprocessing in place to convert, but I can never remember which one is which. We should add some lint tests

PR #1140, opened 2019-10-26
Update transformer.ipynb

jilljenn:
Shouldn't `tape` be within the `with`?


jilljenn:
@googlebot I signed it!

yashk2810:
No, anything inside the with scope is tracked by the tape.

So after you exit it, then you need to calculate the gradients. If you put the calculation inside the tape, the tape will track the gradients too (which is not what we want).

lamberta:
Thanks!

PR #1143, opened 2019-10-28
Update model training chart with new history.history keys

thegrims:


MarkDaoust:
Hi @thegrims, 

Thanks for taking the time to make a PR.

But I think you're misunderstanding how this works.  'acc' and 'accuracy' are interchangeable. You just have to be consistent (pre TF2.0 keras was sometimes _inconsistent_  about this ). You've only changed 2 of the 3 uses so this PR doesn't run as is. 

So I'm closing this one (also I like "accuracy" as it's more descriptive), but thanks for trying.

thegrims:
Thanks for the update! I didn't realize this

PR #1144, opened 2019-10-28
use the BATCH_SIZE variable

samuel500:
BATCH_SIZE is never used.


samuel500:
@googlebot I signed it!

PR #1169, opened 2019-11-05
fix(js): correct typos.

evolify:


evolify:
@googlebot I signed it!

PR #1173, opened 2019-11-07
Added a missing Dense layer in the model.

masa-ita:
There should be a Dense layer with 16 units in the model, according to the description of layers.

### DIFF
--- a/site/en/tutorials/text/word_embeddings.ipynb
+++ b/site/en/tutorials/text/word_embeddings.ipynb
@@ -1,682 +1,694 @@
 {
-  "cells": [
-    {
-      "cell_type": "markdown",
-      "metadata": {
-        "colab_type": "text",
-        "id": "GE91qWZkm8ZQ"
-      },
-      "source": [
-        "##### Copyright 2019 The TensorFlow Authors."
-      ]
-    },
-    {
-      "cell_type": "code",
-      "execution_count": 0,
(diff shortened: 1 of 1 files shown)


site/en/tutorials/text/word_embeddings.ipynb
```diff
@@ -597,7 +625,9 @@
         "else:\n",
         "  files.download('vecs.tsv')\n",
```
lamberta: @MarkDaoust 
I'm getting an error here running in public colab: `MessageError: TypeError: Failed to fetch`

masa-ita:
@lamberta @MarkDaoust 
I thought that changing file paths to absolute ones could solve this issue, but it fails again.
It seems that it will take some time before the files are available.
Relative paths worked second time.

lamberta:
Thanks @masa-ita for looking into it. Can you save the notebook using Colab? It makes the diff easier to see. Either download or use the "Save in GitHub" feature.

masa-ita:
@lamberta 
I revert the tsv file paths to relative and save from Colab.

site/en/tutorials/text/word_embeddings.ipynb
MarkDaoust: That's a Colab problem.

Sometimes it works the second time.

PR #1179, opened 2019-11-09
Updated links

nikochiko:
Few links were outdated (supported r1 documentation only). Updated according to https://www.tensorflow.org/api_docs


MarkDaoust:
Fixed the language binging link to point to: 

https://github.com/tensorflow/docs/blob/master/site/en/r1/guide/extend/bindings.md

Does that seem reasonable to you @lamberta ?

PR #1185, opened 2019-11-14
Update parse_fn as per tf 2.0 apis

sukant16:
Following changes will apply as per tf 2.0:
+  ```tf.FixedLengthFeature((), tf.string, "")``` to ```tf.io.FixedLenFeature([], tf.string, "")```
+  ```tf.parse_single_example(example, example_fmt)``` to ```tf.io.parse_single_example(example, example_fmt)```
+  ```tf.io.image.decode_image(parsed["image"])``` to ```tf.io.decode_image(parsed["image"])```

For reference, see here: 
+  https://www.tensorflow.org/api_docs/python/tf/io/parse_single_example
+  https://www.tensorflow.org/api_docs/python/tf/io/FixedLenFeature
+  https://www.tensorflow.org/api_docs/python/tf/io/decode_image

### DIFF
--- a/site/en/guide/data_performance.md
+++ b/site/en/guide/data_performance.md
@@ -49,11 +49,11 @@ image-label pairs suitable for training. The input pipeline is represented as a
 def parse_fn(example):
   "Parse TFExample records and perform simple data augmentation."
   example_fmt = {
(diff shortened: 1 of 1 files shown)


sukant16:
@googlebot I signed it!

PR #1193, opened 2019-11-17
Fixed typos and removed empty print() call

nikochiko:
Fixed typos and removed empty `print()` call


nikochiko:
@tfdocsbot Here is the [gist](https://colab.research.google.com/gist/nikochiko/b552d01a94cc3425b82126b88de0e1ff/saved_model.ipynb) with all blocks run. Works fine.

PR #1203, opened 2019-11-19
Fix parameters of loss function

tulasiram58827:
While caluclating the loss using SparseCategoricalCrossentropy you are passing predictions at (t)th step and actual output at (t+1)step so the caluclation of loss may be wrong. 
In the training part can you check you are sending input to the decoder at 0 time step and calculating the loss with predicted output at 0 time step and actual output at 1 time step.


tulasiram58827:
@googlebot I signed it!

PR #1204, opened 2019-11-19
Fix typo in custom metric name

tomknig:
I've spotted this (probably unintended) typo on https://www.tensorflow.org/guide/keras/train_and_evaluate


tomknig:
@googlebot I signed it!

PR #1219, opened 2019-11-21
Inconsistent indentation and exception

samuel500:
1. Parts of the notebook used 4 space indentation instead of 2
2. when changing BATCH_SIZE, the notebook would throw an exception because of a hard coded dummy variable batch size (line 640)


lamberta:
@anirudh161 please apply the "ready to pull" label and shepherd through the *internal safe review* after your lgtm. thanks

PR #1225, opened 2019-11-24
Fix loss function bug in GradientTape example

fibigerg:
Add from_logits=True to loss function in tf.GradientTape example. The code is broken without it.

ALLOW_H5=true

### DIFF
--- a/site/en/guide/keras/train_and_evaluate.ipynb
+++ b/site/en/guide/keras/train_and_evaluate.ipynb
@@ -1,4 +1,19 @@
 {
+  "nbformat": 4,
+  "nbformat_minor": 0,
+  "metadata": {
+    "colab": {
+      "name": "train_and_evaluate.ipynb",
+      "provenance": [],
+      "private_outputs": true,
+      "collapsed_sections": [],
+      "toc_visible": true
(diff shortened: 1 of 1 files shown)


fibigerg:
I dig deeper into this, comparing model.compile/fit and GradientTape learning progress on a simple model. There really should be `training=True` in the model call. I am not sure about `from_logits=True`. I had to remove it to have exactly the same learning curve as with model.compile/fit approach.

site/en/guide/keras/train_and_evaluate.ipynb
```diff
@@ -1575,7 +1575,7 @@
         "# Instantiate an optimizer.\n",
         "optimizer = keras.optimizers.SGD(learning_rate=1e-3)\n",
         "# Instantiate a loss function.\n",
-        "loss_fn = keras.losses.SparseCategoricalCrossentropy()\n",
+        "loss_fn = keras.losses.SparseCategoricalCrossentropy(from_logits=True)\n",
```
yashk2810: The last layer activation of the model is set to 'softmax'. Can you remove that so that from_logits=True will work?

PR #1231, opened 2019-11-26
IT: tutorials-keras-text_classification.ipynb - Translation

codeadmin-peritiae:


codeadmin-peritiae:
Error - Just partial translation...

PR #1240, opened 2019-11-27
Fix bad render on documentation website

nalepae:
While it's OK on Notebook

See [here](https://www.tensorflow.org/tutorials/keras/text_classification_with_hub#build_the_model).


nalepae:
@googlebot I signed it!

PR #1242, opened 2019-11-28
EN: keras-classification - Step splitting on model training

codeadmin-peritiae:


site/en/tutorials/keras/classification.ipynb
```diff
@@ -575,7 +575,8 @@
         "\n",
         "1. Feed the training data to the model. In this example, the training data is in the `train_images` and `train_labels` arrays.\n",
         "2. The model learns to associate images and labels.\n",
-        "3. You ask the model to make predictions about a test set—in this example, the `test_images` array. Verify that the predictions match the labels from the `test_labels` array.\n",
+        "3. You ask the model to make predictions about a test set—in this example, the `test_images` array.\n",
+        "4. Verify that the predictions match the labels from the `test_labels` array.\n",
```
lamberta: Seems like a reasonable step to add---but we should add a subheading below that corresponds to this step, otherwise it looks like we're not doing what we recommend.

PR #1259, opened 2019-12-03
[zh-CN] add chinese translation for datasets overview.ipynb

usakey:
translate https://github.com/tensorflow/datasets/blob/master/docs/overview.ipynb


gaoljhy:
signe cla pls

PR #1271, opened 2019-12-07
KO: retranslated data_performance.ipynb

JKIsaacLee:
synced with en/guide/data_performance.ipynb


lamberta:
Thanks for the quick fix, Isaac!

PR #1278, opened 2019-12-11
[zh-CN] add chinese translation of datasets overview

usakey:
translate https://github.com/tensorflow/datasets/blob/master/docs/overview.ipynb

### DIFF
--- a/site/zh-cn/datasets/overview.ipynb
+++ b/site/zh-cn/datasets/overview.ipynb
@@ -1,393 +1,487 @@
 {
- "cells": [
-  {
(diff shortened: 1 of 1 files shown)


gaoljhy:
lgtm

lamberta:
Thanks. Best to copy the original notebook or use the existing [notebook template](https://github.com/tensorflow/docs/blob/master/tools/templates/notebook.ipynb). Please save without the output cells in the notebook

usakey:
@lamberta Thanks for your suggestion. All cell output have been removed.

PR #1291, opened 2019-12-13
Add TensorFlow Egypt

nouran551:
Add TensorFlow Egypt


nouran551:
@googlebot I signed it!

PR #1301, opened 2019-12-15
Remove deprecated tf.data.Dataset.output_shapes

minoring:
Remove deprecated tf.data.Dataset.output_shapes
https://www.tensorflow.org/versions/r1.15/api_docs/python/tf/data/Dataset#output_shapes

Convert output_shapes into None which will be padded to the maximum size of that dimension in each
batch.

Also added links for padded_batch and fixed typo.

(Running in Google Colab gives the error below)
Screen Shot 2019-12-15 at 12 38 33 AM


yashk2810:
This is being fixed by a more permanent solution.

@MarkDaoust has a fix for it that is in flight.

PR #1302, opened 2019-12-15
zh-CN: xsum dataset

Forzoom:


gaoljhy:
lgtm

lamberta:
Appreciate it, but I don't think we want to maintain the TFDS translations in this repo. We don't translate reference docs because they are very difficult to keep in sync.

But there might be a more maintainable way to manage these at the https://github.com/tensorflow/datasets source. Can you file an issue there, please?

PR #1309, opened 2019-12-16
Fix typo

ntabris:
Change "otput" to "output"


ntabris:
@googlebot I signed it!

PR #1310, opened 2019-12-16
Fixed docker install verification script, to work with TF 2.0

PeteBlackerThe3rd:
I noticed that the python code used to verify the installation on the Docker install page is still using a Tensorflow 1.x API call and therefore always fails. This PR updates tf.random_normal to tf.random.normal so this code and succeeds with a correctly installed TF2.0 environment.


lamberta:
Thanks. This was merged in https://github.com/tensorflow/docs/commit/2bce806f2fd35c6aa797a693c87bdf3f448a55a0

PR #1314, opened 2019-12-21
replace depreciated `dataset.output_shapes`

chris-chris:
This tutorial doesn't work in Tensorflow 2.0.


yashk2810:
This is being fixed by a more permanent solution.

@MarkDaoust has a fix for it that is in flight.

yashk2810:
Thanks for the PR!

PR #1320, opened 2019-12-24
Fix installation docker command error

poguez:
When you arrive to https://www.tensorflow.org/install.
The command said you would run a Jupyter notebook on the container start, but you need to have the `-jupyter`image tag to do it. I was confused when I wanted to follow the instructions.


poguez:
@googlebot I signed it!

PR #1334, opened 2019-12-31
Added TUG Coimbatore Chapter

praveentcom:
The Coimbatore Chapter is a newly formed Tensorflow UserGroup hosted in Eventsprint Event Platform with addition of many enthusiasts in Tensorflow who has attended our events in the past via GDG and DevFests.


praveentcom:
@googlebot I signed it!

PR #1343, opened 2020-01-04
JA: Fix translations of the word "iterable"

AseiSugiyama:
I changed translations of the word "iterable" from `反復可能オブジェクト` to `イテラブル`.

We translated the word "iterable" as `反復可能オブジェクト` with [Japanese Documentation of Python](https://docs.python.org/ja/3/index.html) as a reference. Now, however, "iterable" is translated as `イテラブル` on the latest Python document translation.

I guess it is better to be consistent with it.


masa-ita:
I agree with you.
LGTM!

PR #1345, opened 2020-01-04
Add zip file removal /site/en/tutorials/text/image_captioning.ipynb

AseiSugiyama:
This PR adds a code to remove extracted zip files.

In this notebook, we download a large zip file and extract it.
Currently, the amount of disk space is limited to 68.40 GB in Colab. In this notebook, however, we are going to make cache files and it cause running out the disk space.

So, we have to remove zip files for adding disk space to complete this example.


AseiSugiyama:
@MarkDaoust Thank you for your review and suggestion! I made some lines of code to skip the download step if extracted files exist. Could you check it again?

lamberta:
Thanks. Looks like we have a merge conflict

PR #1359, opened 2020-01-08
Replace inline image syntax with img tag to specify image size

satojkovic:
Replace inline image syntax with img tag to specify image size on GitHub.


MarkDaoust:
Thanks!

PR #1363, opened 2020-01-08
Fix typo.

alisaaalehi:
Repetition of the phrase `is that` in the sentence: "One problem with the above experimental.sample_from_datasets approach is that is that it needs a separate tf.data.Dataset per class."


alisaaalehi:
@googlebot I signed it!

PR #1380, opened 2020-01-10
update the links for agents

yongwww:
The previous links are removed, need to update accordingly.


yongwww:
@googlebot I signed it!

PR #1388, opened 2020-01-11
Fix variable name

rickiepark:
`img_inputs` is not used in this notebook.
I think it must be `inputs`. :)
Thanks.

### DIFF
--- a/site/en/guide/keras/functional.ipynb
+++ b/site/en/guide/keras/functional.ipynb
@@ -175,7 +175,7 @@
       },
       "outputs": [],
       "source": [
-        "# just for illustration purposes\n",
+        "# Just for demonstration purposes\n",
         "img_inputs = keras.Input(shape=(32, 32, 3))"
       ]
     },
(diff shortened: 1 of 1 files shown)


lamberta:
Hi Haesun,

I think this is intentional since the previous line says: "For an input meant for images .... we would have used" ---meaning not for this example.

Maybe we can just add a code comment to make it clearer this block is just an example and not used?

rickiepark:
Oops sorry, I didn't see the line.
`inputs = keras.Input(shape=(784,))`
I knew that `inputs` is not defined. :(

Yes, adding comment is good idea.
How about the following?
```
# just for illustration purposes
img_inputs = keras.Input(shape=(32, 32, 3))
```

lamberta:
Sorry for the delay, Haesun. Yes, I think that comment would be a good addition

rickiepark:
Thanks @lamberta ,
I added a comment. :)

PR #1391, opened 2020-01-12
Indentation fixes for with statement.

Kumar-Kishan:
The variable gen_tape and disc_tape were not being used on the indentation level they were being defined


lamberta:
I don't believe this is correct. @yashk2810 to confirm

yashk2810:
Yup, the gradient calculation should happen outside the scope of the context manager. If you put it inside, then the tape with track the gradients for every batch which can result in an OOM error.

PR #1392, opened 2020-01-13
Fix install instructions for TensorRT

dg2:
Without the `libnvinfer-plugin6` package, Tensorflow apparently fails to load TensorRT with the following error:

W tensorflow/stream_executor/platform/default/dso_loader.cc:55] Could not load dynamic library 'libnvinfer_plugin.so.6'; dlerror: libnvinfer_plugin.so.6: cannot open shared object file: No such file or directory
W tensorflow/compiler/tf2tensorrt/utils/py_utils.cc:30] Cannot dlopen some TensorRT libraries. If you would like to use Nvidia GPU with TensorRT, please make sure the missing libraries mentioned above are installed properly.

Adding the package solves the issue


dg2:
@googlebot I signed it!

PR #1404, opened 2020-01-15
Disabled input shuffle when loading CSVs

Marek-Sedlacek:
In the function `tf.data.experimental.make_csv_dataset`, the `shuffle` argument is set to True by default. This will affect the final step in the notebook where predictions are compared with labels. Setting `shuffle` to False ensures the predictions and labels being compared are the same instances.

### DIFF
--- a/site/en/tutorials/load_data/csv.ipynb
+++ b/site/en/tutorials/load_data/csv.ipynb
@@ -244,6 +244,7 @@
         "      na_value=\"?\",\n",
         "      num_epochs=1,\n",
         "      ignore_errors=True, \n",
+        "      shuffle=False, \n",
         "      **kwargs)\n",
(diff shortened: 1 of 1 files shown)


site/en/tutorials/load_data/csv.ipynb
```diff
@@ -244,6 +244,7 @@
         "      na_value=\"?\",\n",
         "      num_epochs=1,\n",
         "      ignore_errors=True, \n",
+        "      shuffle=False, \n",
```
yashk2810, line 247: Can you make it a parameter instead?

So for raw_train_data it will be `get_dataset(train_file_path, shuffle=True)`

raw_test_data = get_dataset(test_file_path, shuffle=False)

lamberta:

PR #1406, opened 2020-01-17
updated tutorial

Mbah-Javis:
I think this might be helpful. I updated this [Tensorflow tutorial](https://github.com/tensorflow/docs/blob/master/site/en/r1/tutorials/estimators/cnn.ipynb) to Tensorflow version 2 :)


lamberta:
Thanks. We decided not to port this r1 tutorial because the recommended path is to use [model_to_estimator](https://www.tensorflow.org/guide/estimator#create_an_estimator_from_a_keras_model). The `compat.v1` tools are great for migration to TF2.

The `r1/` has been archived and in maintenance mode. We're not adding new docs to it. But this might make an interesting blog post to demonstrate a migration.

cc: @random-forests

Mbah-Javis:
@lamberta Okay :)

PR #1408, opened 2020-01-17
Add SequenceExample tutorial

ADBalici:
The documentation out there is a little bit sparse on how to use SequenceExample(s).

I hope the new .ipynb will help other users of TensorFlow.

I am not exactly sure how the banner containing links to Collab should be generated, as the contributors guide does not contain details on this process. Perhaps somebody can help me here?

### DIFF
--- a/site/en/tutorials/load_data/seq_example.ipynb
+++ b/site/en/tutorials/load_data/seq_example.ipynb
@@ -1,392 +1,399 @@
 {
- "cells": [
-  {
-   "cell_type": "markdown",
-   "metadata": {},
-   "source": [
-    "##### Copyright 2019 The TensorFlow Authors."
-   ]
+  "cells": [
+    {
+      "cell_type": "markdown",
+      "metadata": {},
+      "source": [
+        "##### Copyright 2019 The TensorFlow Authors."
+      ]
+    },
+    {
+      "cell_type": "code",
+      "execution_count": null,
+      "metadata": {
+        "cellView": "form"
+      },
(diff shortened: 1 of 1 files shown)


ADBalici:
@googlebot I fixed it.

ADBalici:
@lamberta @MarkDaoust @yashk2810

lamberta:
Thanks. There's a notebook template here: https://github.com/tensorflow/docs/blob/master/tools/templates/notebook.ipynb

Not sure when we'll be able to review, but hopefully someone from the community can take a preliminary look. Might want to ping the docs@tensorflow.org list. We're working on a process for tutorial/guide proposals and submissions so we can track these better.

MarkDaoust:
Hi @ADBalici, I see we don't have any examples of this API.

It's always felt like an over-compilcated way to do this sort of thing, and sees very little usage. (50x less than `tf.Example`)

There is nothing incorrect about your sample code, but we probably don't want to give this so much visibility.

I would be more comfortable just having quick, concrete doctest-samples added injected into to the docstrings for SequenceExample, and the surrounding APIs.

Also note that if we do move forward with this it will be important to have code that shows how to load the SequenceExamples out of the TFRecord file.

@rohan100jain, WDYT about this?

lamberta:
PR stalled but thanks.
For community contributions, best to engage with the TensorFlow SIGS: https://github.com/tensorflow/community/tree/master/sigs
Or the ML GDEs: https://github.com/ml-gde

PR #1411, opened 2020-01-19
Replace # symbol with translated text

satojkovic:


masa-ita:
Thank you for your PR.  I think this is an appropriate fix.
LGTM!

PR #1424, opened 2020-01-23
Update installation instructions for VS 2019 (post-2.1.0) packages

angerson:
TensorFlow 2.1.0 was built with Visual Studio 2019, which requires an additional DLL from Visual Studio 2019, `msvcp140_1.dll`. I updated the internal checker for this, but being unfamiliar with the installation instructions, I didn't update this page when the change occurred.

See also: tensorflow/tensorflow/commit/78ddcded185e60f83743faa194178a8e5979757b, https://github.com/tensorflow/tensorflow/issues/35036

### DIFF
--- a/site/en/install/pip.html
+++ b/site/en/install/pip.html
@@ -158,13 +158,15 @@
 <section>
 <h3>Windows</h3><!--python3-->
 <p>
-  Install the <em>Microsoft Visual C++ 2015 Redistributable Update 3</em>. This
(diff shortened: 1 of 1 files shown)


angerson:
@lamberta I'm not sure if the wording here does what I want -- the restriction applies to 2.1.0 and above, so should this be "greater than 2.1.0" as it is now, or "all versions since 2.1.0", or something else?

PR #1426, opened 2020-01-24
typo in save_and_load.ipynb

mulka:
shuold -> should


mulka:
@googlebot I signed it!

PR #1435, opened 2020-01-27
Python version corrected in doc

ashutosh1919:
There are issues like [#33374](https://github.com/tensorflow/tensorflow/issues/33374) which asks for installing tf in python 3.8 but as of now, it is not supporting. Also, there are many issues being made everyday about the same problem. So, this is the correction in the doc to illustrate that tf supports python version 3.4 to 3.7.

### DIFF
--- a/site/en/install/_index.yaml
+++ b/site/en/install/_index.yaml
@@ -15,6 +15,7 @@ landing_page:
         <table class="columns">
           <tr><td>
             <ul>
+              <li>Python 3.4–3.7</li>
               <li>Ubuntu 16.04 or later</li>
               <li>Windows 7 or later</li>
             </ul>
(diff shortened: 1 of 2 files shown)


mihaimaruseac:
Don't think it's worth it running hours of CI for a single change that would get fixed by the end of the month anyway.

mihaimaruseac:
Aplogoies @ashutosh1919 I thought this is on TF source code repo

ashutosh1919:
> Aplogoies @ashutosh1919 I thought this is on TF source code repo

It's completely fine @mihaimaruseac , I created this PR just because in last 5 days, I saw 7 issues being created of tensorflow not installing in python 3.8 errors.

PR #1441, opened 2020-01-28
Update to avoid error from function iter

atharva1503:
To fix the error due to function 'iter' in the cell: https://colab.research.google.com/github/tensorflow/docs/blob/master/site/en/tutorials/keras/text_classification_with_hub.ipynb#scrollTo=QtTS4kpEpjbi

Error was: RuntimeError: __iter__() is only supported inside of tf.function or when eager execution is enabled.
To fix it: Use the command 'tf.enable_eager_execution()' while calling the liberary at start

### DIFF
--- a/site/en/tutorials/keras/text_classification_with_hub.ipynb
+++ b/site/en/tutorials/keras/text_classification_with_hub.ipynb
@@ -138,6 +138,7 @@
         "except Exception:\n",
         "  pass\n",
         "import tensorflow as tf\n",
+        "tf.enable_eager_execution()\n",
         "\n",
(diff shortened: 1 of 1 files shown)


atharva1503:
But calling function  'tf.enable_eager_execution()' is only required for tf version <2

lamberta:
Thanks, but these are all for TF2 and eager is on by default.
TF1  notebooks are archived in the [site/en/r1/](https://github.com/tensorflow/docs/tree/master/site/en/r1) directory.

atharva1503:
Ok no problem.I got this now and my tf was not updated

PR #1443, opened 2020-02-05
zh-CN: install/gpu.md

Berrywrq:


lamberta:
Thanks @Berrywrq but community translations have moved to this repo: https://github.com/tensorflow/docs-l10n

And because of doc churn and reliability, we do not use community translations for the `/install/` directory (for ja, ko, and zh-cn). I've sent off a new translation job for that and expect tensorflow.org to be updated in a few weeks.

PR #1448, opened 2020-02-07
add `tf.` for prediction_layer in `/tutorials/images/transfer_learning`

jangsoopark:
To avoid confusion, 

Remove `keras = tf.keras` at the top.

And append `tf.` for prediction layers
`prediction_layer = tf.keras.layers.Dense(1)`


jangsoopark:
@googlebot  I fixed it.

PR #1452, opened 2020-02-09
Quick fix in text generation loop, fixes #36591

ndombe:
Remember all previously generated letters and use the string they form together as input for generating the next letter (instead of only considering the lastly generated letter).


ndombe:
@googlebot I signed it!

MarkDaoust:
Hi @ndombe,

Thanks for taking the time to make this PR.

But I do not think this change is correct.

1. We're using a "stateful" RNN: unless you call `.reset_states` the network remembers it's state, it remembers what it's seen so far predicts the next step. It is not predicting just based on the last character input (note how this change does't increase the text quality).
2. if we were not using a "stateful" RNN,  the network would return an `output, state` pair.  You can pass in the `state` in with the next input, so you con't need rerun the whole sequence to predict each letter. 
3. Passing the whole sequence to a "stateless-rnn" would work, but it would makes the runtime quadratic on the  sequence length, you can see this by comparing the runtime of your version to the original.

PR #1453, opened 2020-02-10
update required dependency

kuri-leo:
Fix https://github.com/tensorflow/tensorflow/issues/36588

[TF 2.1 release](https://github.com/tensorflow/tensorflow/releases/tag/v2.1.0) brings in this new dependency requirement

### DIFF
--- a/site/en/install/pip.html
+++ b/site/en/install/pip.html
@@ -31,6 +31,10 @@
   <li>Ubuntu 16.04 or later (64-bit)</li>
   <li>macOS 10.12.6 (Sierra) or later (64-bit) <em>(no GPU support)</em></li>
   <li>Windows 7 or later (64-bit) <em>(Python 3 only)</em></li>
+    <ul>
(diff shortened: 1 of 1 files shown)


site/en/install/pip.html
```diff
@@ -32,7 +32,8 @@ <h3>System requirements</h3>
   <li>macOS 10.12.6 (Sierra) or later (64-bit) <em>(no GPU support)</em></li>
   <li>Windows 7 or later (64-bit) <em>(Python 3 only)</em></li>
```
lamberta: Perhaps add as a bullet point under Windows.
We already do note this in the Windows tab selector.

kuri-leo: Just fixed that, is this right? (plz forgive my poor Html code 🤦‍♂

lamberta: Almost! Webmaster-in-training 😄

kuri-leo: Wow, thanks a lot! 👏

PR #1456, opened 2020-02-13
add installation instructions for TF2

rushabh-v:
resolves: https://github.com/tensorflow/tensorflow/issues/36595


MarkDaoust:
I don't think this is correct.

You probably have tensorflow 1.14 installed, and "pip install tensorflow" says "tensorflow is already installed".

Try "pip install -U tensorflow".

TheUnixDude:
This is a valid pull request, Mark. Pip does not handle installation or upgrades of Tensorflow correctly on Debian Stable.

MarkDaoust:
Okay, let's try to figure this out on #36595, and propose a more specific fix since this pip page is meant to cover a lot more than just Debian Stable.

PR #1457, opened 2020-02-13
Update the beginner.ipynb file to add Installation instruction link

TheUnixDude:
Provide a link to the Install site so there is a clear(er) indication that the installation is non-standard and requires tool versions which are not available by default in some supported distributions.


TheUnixDude:
Ah, this needs a CLA. I specifically do not and can not agree to the Google CLA.

PR #1459, opened 2020-02-14
Fix loss for transformer

vineetm:
Loss computation correctly masks the loss tensor, but while taking mean considers all the zero loss values too! The right method should be to divide the total loss by total active values in mask..


vineetm:
closing due to incorrect github user in commit. Will re-create a new PR

PR #1460, opened 2020-02-14
Fix Loss for Tranformer Tutorial

vineetm:
Hi,

In loss computation, taking a mean would compute incorrect loss, as mean would divide by number of values in mask. We should divide the total loss by all the non-padded values..


lamberta:
This may need to be rebased with the master branch

PR #1471, opened 2020-02-23
R1.0

ziadahmed12:
the tenser flow 1.0 is required for me to due the graduation project am in BUE in Egypt can you allow me to download this data with the instruction on how to install it and the directory which i want to paste it in

           thanks in advanced


lamberta:
This is a bad pull request so not sure I follow ...

2.x is the supported documentation on [tensorflow.org](https://www.tensorflow.org/overview),
but you can still find the r1.15 API reference docs here: https://www.tensorflow.org/versions/r1.15/api_docs/python/tf
with r1.x guides and tutorials archived here: https://github.com/tensorflow/docs/tree/master/site/en/r1

PR #1474, opened 2020-02-25
Avoid verbose lambda func and use simpler pandas.Series.map

Angus-Luo:
Please refer to examples in https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.map.html


MarkDaoust:
Nice. Thanks.

PR #1475, opened 2020-02-25
Update to Transformer URL in docs after reorganization in models repo

kweinmeister:


kweinmeister:
@lamberta No problem, submitted https://github.com/tensorflow/docs-l10n/pull/81

PR #1476, opened 2020-02-26
Fixed name in text_captioning

solaris33:
Fix wrong explanation


yashk2810:
Thanks for the PR!

I think hidden_size is more appropriate here and is more explicit as to what the shape should be.

PR #1479, opened 2020-02-28
mask in loss function did not have the correct shape

borundev:
The loss_function in master is

  mask = tf.math.logical_not(tf.math.equal(real, 0))
......
  loss_ *= mask

  but mask and loss have different shapes so it is not doing the intended action of masking away the loss when real values are 0 and instead leads to a broadcasted matrix.

The second line should instead be

loss_ *= tf.reshape(mask,(-1,))

PS. I am not quite familiar with how to submit ipynb files with git and it seems the whole file has been changed and is not showing a meaningful diff. If there is some procedure/tool to do it properly please let me know and I will do it.

### DIFF
--- a/site/en/tutorials/text/nmt_with_attention.ipynb
+++ b/site/en/tutorials/text/nmt_with_attention.ipynb
@@ -10,24 +10,11 @@
     },
     "kernelspec": {
       "display_name": "Python 3",
-      "language": "python",
       "name": "python3"
-    },
-    "language_info": {
-      "codemirror_mode": {
-        "name": "ipython",
-        "version": 3
-      },
-      "file_extension": ".py",
-      "mimetype": "text/x-python",
-      "name": "python",
-      "nbconvert_exporter": "python",
-      "pygments_lexer": "ipython3",
-      "version": "3.6.9"
     }
   },
   "nbformat": 4,
-  "nbformat_minor": 1,
+  "nbformat_minor": 0,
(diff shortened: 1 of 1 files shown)


borundev:
@googlebot I signed it!

borundev:
@googlebot I fixed it.

borundev:
@googlebot I fixed it.

borundev:
@googlebot I signed it!

borundev:
@googlebot I fixed it

borundev:
@googlebot I fixed it

lamberta:
I verified the CLA, thank you

lamberta:
To make the diffs easier to read, can you please run notebooks through [nbfmt.py](https://github.com/tensorflow/docs/blob/master/tools/nbfmt.py) ?
Thanks

borundev:
@lamberta I ran nbfmt on the notebook. it introduced some extra changes on top (like replacing an encoding of < with < itself). Presumably that's an intended behavior. Had me confused for a whole. My change is limited to line 676 of the master version.

Also googlebot changed my CLA back to no. Since you manually changed it to yes last time, can you let me know if there is something I can do to not have the bot do this again?

borundev:
I am closing the pull request as I think it is not a but in the context in which it was being used. I seem to have seen the error when I was experimenting with it and passing more general rank tensors.

PR #1480, opened 2020-02-28
Edit string shortcut for model compilation

jaketae:
The document states that string identifiers can be used as shortcut, yet the example given is no different from the compilation method shown in the code block above (presumably without strings identifiers, which is also confusing given that the metric was defined with a string identifier). The edit adds clarity by demonstrating how models can be compiled with and without string identifiers.

### DIFF
--- a/site/en/guide/keras/train_and_evaluate.ipynb
+++ b/site/en/guide/keras/train_and_evaluate.ipynb
@@ -362,7 +362,7 @@
         "If your model has multiple outputs, your can specify  different losses and metrics for each output,\n",
         "and you can modulate the contribution of each output to the total loss of the model. You will find more details about this in the section \"**Passing data to multi-input, multi-output models**\".\n",
         "\n",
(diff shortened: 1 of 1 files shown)


site/en/guide/keras/train_and_evaluate.ipynb
```diff
@@ -375,8 +375,8 @@
       },
       "outputs": [],
       "source": [
-        "model.compile(optimizer=keras.optimizers.RMSprop(learning_rate=1e-3),\n",
-        "              loss=keras.losses.SparseCategoricalCrossentropy(from_logits=True),\n",
```
lamberta: @MarkDaoust I think `from_logits=True` is the preferred usage these days, yes?

MarkDaoust: Exactly.

I reverted that line, and updated the text.

PR #1482, opened 2020-02-29
R2.0

lyj1998:
Something is wrong with position encoding,sin and cos should be concatenated.


MarkDaoust:
Hi @lyj1998, 

It's not clear what you're trying to do here.

If you're trying to edit the api_docs on tensorflow.org, you need to make those changes to the sopurce files in the tensorflow/tensorflow repository.

PR #1483, opened 2020-03-02
Minor edit to markdown inline code

jaketae:
A very minor edit from tf.keras.model to `tf.keras.model` for notational consistency and better readability.


MarkDaoust:
Hi @jaketae, 

But, IIRC, tensorflow.org doesn't handle links in headings very well, and the auto-api-linker will add one here.

So I'd prefer not to take this PR.

Thanks.

jaketae:
@MarkDaoust I just realized what you mean after seeing that other section headers seemed to also avoid inline code markdown for things like `input_fn`. Thanks for the heads up, and appreciate the quick feedback.

PR #1486, opened 2020-03-03
text_classification_rnn.py: Add mandatory `padded_shapes` argument in `padded_batch`

nalepae:


MarkDaoust:
Hi @nalepae,

Thanks for the PR.

I did this in preparation for the 2.2, where the new default behavior is sufficient for these examples. 

This notebook has a `pip install tf-nightly` at the top, so that it runs in Colab.

But I recognize that this is causing some confusion. Maybe I did it a little early. But the 2.2 branch is cut, and will be in rc soon.

If I accept this PR, I'll just have to come back and undo it at release.

So I rather keep it as is.

Thanks again.

nalepae:
Yes you'r right, I used the wrong tensorflow (NOT the nightly one).
But I'm wondering if it is really relevant to use tfnightly in the official
documentation.

For example I wanted to use padding for a professional use case, where I
don't want to use nightly tf.
I guess I'm not the only one in this case.

Anyway, thank you for your excellent documentation!

Le mar. 3 mars 2020 à 14:54, Mark Daoust <notifications@github.com> a
écrit :

PR #1489, opened 2020-03-04
change download url for dataset

Ir1d:
closes https://github.com/tensorflow/tensorflow/issues/34460
Changed the url according to the script `tensorflow/examples/speech_commands/train.py` .
Also, the previous url has some certification error

### DIFF
--- a/site/en/r1/tutorials/sequences/audio_recognition.md
+++ b/site/en/r1/tutorials/sequences/audio_recognition.md
@@ -402,7 +402,7 @@ change to customize the results for your own requirements.
 ### Custom Training Data
 
 By default the script will download the [Speech Commands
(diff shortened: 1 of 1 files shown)


Ir1d:
@lamberta I've commited the suggestion. But I'm actually a bit confused because the link in the code and in the [doc](https://github.com/Ir1d/docs/blame/patch-1/site/en/r1/tutorials/sequences/audio_recognition.md#L28) are both v0.02.tar.gz It is inconsistent with the code if its referring to `tensorflow/examples/speech_commands/train.py ` as shown [here](https://github.com/Ir1d/docs/blame/patch-1/site/en/r1/tutorials/sequences/audio_recognition.md#L24)

PR #1487, opened 2020-03-03
Fixed minor typo

jusonn:
In the introduction section of keras/functional.ipynb.
There are no "is" in "The Functional API a set of tools for building graphs of layers.".
Fixed to "The Functional API is a set of tools for building graphs of layers."

PR #1490, opened 2020-03-04
Copy edits to the Keras functional guide

lamberta:


lamberta:
Additional CLA verified in https://github.com/tensorflow/docs/pull/1487

PR #1497, opened 2020-03-12
Fix description error in classification tutorial

amnivek:
The last dense layer is a logits layer, instead of a softmax layer.


amnivek:
@googlebot I signed it!

PR #1505, opened 2020-03-19
fixed tensorflow issue #37713

khimraj:


khimraj:
@googlebot I signed it!

MarkDaoust:
Thanks @khimraj , 

This notebook starts with a `!pip install tf-nightly` to get the right version. 

I acknowledge that we could be clearer about pre-release docs like this, and we will be in the future.

But the 2.2 release is coming soon, and the simplified version will work there.

PR #1506, opened 2020-03-19
fix missing positional argument to padded_batch

jsenellart:
When running tutorial `load_data/text.ipynb` with tensorflow 2.1 - the following line fails:

```
train_data = train_data.padded_batch(BATCH_SIZE)
```

with the error message:
```
TypeError: padded_batch() missing 1 required positional argument: 'padded_shapes'
```

And the document actually mentions this argument: https://www.tensorflow.org/api_docs/python/tf/data/Dataset#padded_batch.

The proposed patch adds the expected parameters with default value to pad on the text sequence.


MarkDaoust:
Yeah, this is causing a lot of confusion.

I added the `pip install tf-nightly` at the top but that's not enough.

Thanks for the PR. I'm sending a fix for all the tutorials.

PR #1508, opened 2020-03-20
Correct the second argument of the plot_cm

thap:
should be test_predictions_resampled instead of test_predictions_weighted


MarkDaoust:
Thanks, the change looks good, but you need to sign the CLA before I can merge this.

thap:
@googlebot I signed it!

PR #1511, opened 2020-03-24
Update source.md

camiloramireza:
The actual directory of the  source tree in the docker image is /tensorflow_src not /tensorflow as specified in the documentation.


camiloramireza:
@googlebot I signed it!

PR #1515, opened 2020-03-26
[MRG] Rephrase subsection title for clarity in tutorials/text/nmt

mohamed-ali:
It is not clear in this subsection title: [Create a tf.data dataset](https://www.tensorflow.org/tutorials/text/nmt_with_attention#create_a_tfdata_dataset), whether the author wants to ask the reader to:

* "Create a tf.data.Dataset" : in that case there is a missing dot and capital D
* or say what is written: "Create a tf.data dataset".

The subtle difference makes the translation of this subsection title hard to convey a clear meaning. Therefore, I suggest using the clearer subsection title: *Use tf.data to create batches and shuffle the dataset* that is used in [Convolutional Variational Autoencoder](https://www.tensorflow.org/tutorials/generative/cvae#use_tfdata_to_create_batches_and_shuffle_the_dataset) tutorial.

### DIFF
--- a/site/en/tutorials/text/nmt_with_attention.ipynb
+++ b/site/en/tutorials/text/nmt_with_attention.ipynb
@@ -395,7 +395,7 @@
         "id": "rgCLkfv5uO3d"
       },
       "source": [
-        "### Create a tf.data dataset"
+        "### Use tf.data to create batches and shuffle the dataset"
       ]
     },
     {


site/en/tutorials/text/nmt_with_attention.ipynb
```diff
@@ -395,7 +395,7 @@
         "id": "rgCLkfv5uO3d"
       },
       "source": [
-        "### Create a tf.data dataset"
+        "### Use tf.data to create batches and shuffle the dataset"
```
lamberta: Unfortunately, the page nav isn't very wide, as you can see here: https://www.tensorflow.org/tutorials/text/nmt_with_attention
Open to changes that make this clear, but would like to avoid spilling over to a second line. Can always add further explanation to the text that follows.

mohamed-ali: @lamberta I thought it's okay because it's already used in this page with the same page navigation: https://www.tensorflow.org/tutorials/generative/cvae#use_tfdata_to_create_batches_and_shuffle_the_dataset

lamberta: I'd rather include that page in this fix :)

PR #1517, opened 2020-03-28
Fixbug #37971: Fixed all the broken links by appending .md in /docs/index.md

Rubix982:
This is in reference to issue #37971, which you can find [here](https://github.com/tensorflow/tensorflow/issues/37971).

Basically all the links in /docs/index.md were 'broken'. The links did not have a `.md` at the end for some reason. This commit basically replaces `"><code>` with `.md"><code>` so that all the links will now map properly instead.

I am not sure why any of the links did not a .md at the attend. I tried with all the links, none of the links worked before, only started to work after I appended a .md at the URL's place to the original URL string.


MarkDaoust:
There are a lot more broken links that I'll try to address in a more systematic way.

But this does fix this page

Rubix982:
@MarkDaoust please let me know about those as well, would love to solve them - it just takes 3-4 minutes of confirming URLs by appending some text.

PR #1526, opened 2020-03-31
Add nbfmt to GitHub CI

lamberta:
Similar to the [docs-l10n version](https://github.com/tensorflow/docs-l10n/blob/master/.github/workflows/ci.yaml), but doesn't require downloading `nbfmt.py`.

Tried on [a number of PRs](https://github.com/tensorflow/docs-l10n/actions) over there and seems to WAI.


lamberta:
cl/304075356

PR #1527, opened 2020-04-01
Add activation Fixes #38091

maxkaustav:
As mentioned in the document activation is sigmoid which is mentioned in the layer also

### DIFF
--- a/site/en/tutorials/keras/text_classification_with_hub.ipynb
+++ b/site/en/tutorials/keras/text_classification_with_hub.ipynb
@@ -304,7 +304,7 @@
         "model = tf.keras.Sequential()\n",
         "model.add(hub_layer)\n",
         "model.add(tf.keras.layers.Dense(16, activation='relu'))\n",
-        "model.add(tf.keras.layers.Dense(1, activation='sigmoid'))\n",
+        "model.add(tf.keras.layers.Dense(1))\n",
         "\n",
         "model.summary()"
       ]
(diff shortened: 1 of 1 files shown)


lamberta:
Thanks, but this was intentionally removed to prefer logits. Can you instead remove the text references, please?

maxkaustav:
Thank you,sure i will be correcting that

ashutosh1919:
Fixes [#38091](https://github.com/tensorflow/tensorflow/issues/38091)

huikang:
> Thanks, but this was intentionally removed to prefer logits. Can you instead remove the text references, please?

Without activation='sigmoid', the predictions are not in the range of (0, 1) as shown below, which is not interpretable.

```
pred = model.predict(test_data.batch(512))
print(pred)

[[-0.29496038]
 [ 1.2088487 ]
 [ 0.11580676]
 ...
 [-1.610341  ]
 [-0.8496179 ]
 [ 1.3117154 ]]
```

Could you help me to understand why not using sigmoid? Thanks.

lamberta:
@huikang It's a debate around interpretability versus numeric stability. A quick search brought up [this article](https://towardsdatascience.com/sigmoid-activation-and-binary-crossentropy-a-less-than-perfect-match-b801e130e31) that may be worth a look (disclaimer: I just skimmed it :)

huikang:
@lamberta , thanks for sharing the link, which is very helpful to me as new to data science.

One follow-up question in this particular example is that if we don't use `activation='sigmoid'`, the predicted value will be out of the range of (0,1). So, does that mean a prediction, say -0.29, is less likely to survive than 0? Or in other word, the larger the prediction is, the higher chance the passenger will survive.

maxkaustav:
> 
> 
> @lamberta , thanks for sharing the link, which is very helpful to me as new to data science.
> 
> One follow-up question in this particular example is that if we don't use `activation='sigmoid'`, the predicted value will be out of the range of (0,1). So, does that mean a prediction, say -0.29, is less likely to survive than 0? Or in other word, the larger the prediction is, the higher chance the passenger will survive.

usually, logit function in mathematics maps sigmoid(0,1) to (-inf,inf) so we get a huge range of stable values generally it is inverse of the sigmoid. So if 1 (survive) then the probability of surviving is for a higher value of logit.

PR #1528, opened 2020-04-01
Corrected order of params of loss fn in custom_training.ipynb

ashutosh1919:
Fixes [#38067](https://github.com/tensorflow/tensorflow/issues/38067).  
I agree with @Arktius that even if the loss here given is MSE and it is not affected by order of params, it is good to follow the convention of every loss function that `y_true` should be first and then `y_pred`.  
@MarkDaoust and @mihaimaruseac , Please review.


ashutosh1919:
@MarkDaoust , as per your comments, code changed. Please review.

PR #1529, opened 2020-04-02
nbfmt sync

lamberta:
```
find . -name "*\.ipynb" | xargs ./tools/nbfmt.py
```


yashk2810:
Oh, these fields. Good riddance!

PR #1530, opened 2020-04-03
Add proper activation function

kyscg:
In this load_data/csv tutorial (https://www.tensorflow.org/tutorials/load_data/csv), the final prediction values are weird. For instance:
```
Predicted survival: -236.86% | Actual outcome: SURVIVED
Predicted survival: -100.17% | Actual outcome: DIED
Predicted survival: -238.74% | Actual outcome: DIED
Predicted survival: 147.69% | Actual outcome: DIED
Predicted survival: 12.94% | Actual outcome: SURVIVED
```

This PR adds the `sigmoid` activation function to the last `Dense` layer, keeping the result between 0 and 1:

```
Predicted survival: 0.01%  | Actual outcome:  SURVIVED
Predicted survival: 100.00%  | Actual outcome:  DIED
Predicted survival: 0.02%  | Actual outcome:  SURVIVED
Predicted survival: 99.86%  | Actual outcome:  DIED
Predicted survival: 2.43%  | Actual outcome:  SURVIVED
```

### DIFF
--- a/site/en/tutorials/load_data/csv.ipynb
+++ b/site/en/tutorials/load_data/csv.ipynb
@@ -866,7 +866,7 @@
         "  preprocessing_layer,\n",
         "  tf.keras.layers.Dense(128, activation='relu'),\n",
         "  tf.keras.layers.Dense(128, activation='relu'),\n",
-        "  tf.keras.layers.Dense(1),\n",
+        "  tf.keras.layers.Dense(1, activation='sigmoid'),\n",
         "])\n",
         "\n",
         "model.compile(\n",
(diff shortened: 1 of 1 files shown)


lamberta:
We're intentionally using logits—though I agree the final section reads weird. I think it would be better to convert it to something more interpretable when displaying the results and not in the model. @MarkDaoust wdyt?

MarkDaoust:
Thanks.

Yes. se try to avoid outputting probabilities before calculating the loss. 

Moved sigmoid to the end, so the print-outs make sense.

PR #1538, opened 2020-04-12
PEP 8 fix in Quickstart for experts.

ReHoss:
Surround binary operators with a single space on either side in the format string method of the [Quickstart for experts](https://github.com/tensorflow/docs/blob/master/site/en/tutorials/quickstart/advanced.ipynb) notebook.

Thanks !

### DIFF
--- a/site/en/tutorials/quickstart/advanced.ipynb
+++ b/site/en/tutorials/quickstart/advanced.ipynb
@@ -355,11 +355,11 @@
         "    test_step(test_images, test_labels)\n",
         "\n",
         "  template = 'Epoch {}, Loss: {}, Accuracy: {}, Test Loss: {}, Test Accuracy: {}'\n",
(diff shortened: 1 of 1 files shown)


ReHoss:
Also, according to the [Google Python Style Guide](http://google.github.io/styleguide/pyguide.html) : 

> Use import statements for packages and modules only, not for individual classes or functions. Note that there is an explicit exemption for imports from the typing module.

Should the imports modified and the code adapted to the above claim ?

lamberta:
Hello.
Here's the relevant code style section for contributing to the TensorFlow code base: https://www.tensorflow.org/community/contribute/code_style
We haven't been enforcing this for docs since we'd prefer something automated. But if you combined many fixes into a single PR it might be nice. I just want to avoid having to merge a bunch of these since I'm sure there are countless style violations 😀

PR #1540, opened 2020-04-13
Fix duplicate 'the' in docs

septatrix:


septatrix:
@googlebot I signed it!

PR #1541, opened 2020-04-13
Estimator Tutorial code readability change

constantinexisc:
Someone forgot to use the `fc = tf.feature_column` variable that's declared on the feature creation code block, improves readability so I changed the whole rest of the notebook to actually use it.


constantinexisc:
@googlebot I signed it!

constantinexisc:
@lamberta Better to remove it I guess.

PR #1545, opened 2020-04-14
Update word_embeddings.ipynb

Eshan-Agarwal:


Eshan-Agarwal:
> Definitely don't want to use compat.v1 here.
> @MarkDaoust Can this be removed now?

I am using this first `train_datas.output_shapes` but it gives warning says it is depreciated.

MarkDaoust:
Yeah, everything in `compat.v1` is effectively deprecated too. 

I'm planning to delete this cell after the 2.2 launch (any day now).

It looks like there's no longer a supported way to get the nested shape structure from a `tf.data.Dataset`. The closest thing I can find is: `tf.nest.map_structure(lambda s:s.shape.as_list(), dataset.element_spec)`.

If there's a spot to mention that it's in the docs for padded batch.

PR #1551, opened 2020-04-27
Update rnn.ipynb

jjs1015:
The default `merge_mode` for `tf.keras.layers.Bidirectional` in v2.1.0 is `concate`, according to https://www.tensorflow.org/api_docs/python/tf/keras/layers/Bidirectional


jjs1015:
@googlebot I signed it!

PR #1553, opened 2020-05-03
cmap is ignored for imgshow with RGB channels

amitport:


8bitmp3:
Thanks for the PR, we haven't forgotten about it 🙂 

- The `cmap` argument in `imshow()` is indeed ignored for RGB(A) data in Matplotlib. 
- In addition, Matplotlib suggests that the `binary` colormap (in the `plt.cm` namespace) is redundant and may be removed from Matplotlib in future versions (we should use `gray_r` instead).

We’ll also lint your notebook with `nblint` (https://blog.tensorflow.org/2020/10/how-tensorflow-docs-uses-juypter-notebooks.html) and check for merge conflicts, since it’s been a while.

8bitmp3:
@googlebot I consent.

8bitmp3 (approved):
- Formatted/linted, applied tensorflow_docs.tools nbfmt.
- Added more accessible/"a11y" language (a bit less "see")
- Provided a name for the _TensorFlow 2 quickstart for experts_ link at the end of the tutorial

@amitport Thanks for waiting. Looking forward to your future contributions.

PR #1556, opened 2020-05-05
Fixing Oxford pets dataset download error in tutorials/images/segmentation.ipynb 

logiccodes:
There is a checksum error when running `tutorials/images/segmentation.ipynb` in Colab.

The latest tensorflow_datasets patch seems unable to solve this issue, see: https://github.com/tensorflow/datasets/issues/1978

As an alternative, I've updated the `tfds.load` function into the first method mentioned in https://github.com/tensorflow/tensorflow/issues/31171#issuecomment-619580795

Although this seems a bit complicated, it solves the checksum error.

### DIFF
--- a/site/en/tutorials/images/segmentation.ipynb
+++ b/site/en/tutorials/images/segmentation.ipynb
@@ -1,4 +1,20 @@
 {
+  "nbformat": 4,
+  "nbformat_minor": 0,
+  "metadata": {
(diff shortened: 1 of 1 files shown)


MarkDaoust:
The tfds team is working on a fix, so let's hold for now.

yashk2810:
This is now fixed.

PR #1558, opened 2020-05-05
Changed the input range for mobilenetV2

rabinadk1:
Previously it was [0,1] but the input mobilenetV2 needs it in the form of [-1,1]. So made changes in the notebook to accommodate the change.


rabinadk1:
@googlebot I signed it!

PR #1559, opened 2020-05-05
Fix minor typo, remove unused variable

jaketae:
"calcualted" to "calculated"

----------
**EDIT:** Also removed an unused variable, `style_result`


jaketae:
@MarkDaoust, thank you for the quick review. I've pushed another small commit to this PR, namely removing an unused variable `style_result`. I'm leaving a note here for your reference. Thanks again!

PR #1560, opened 2020-05-06
Fix minor typo

jessicayywu:
tecniques -> techniques


jessicayywu:
@googlebot I signed it!

PR #427, opened 2019-03-27
JA: add images.ipynb, tf-records.ipynb translated in site/ja/tutorials/load_data

masa-ita:
images.ipynb and tf-records.ipynb in site/ja/tutorials/load_data are translated in Japanese.

PR #1562, opened 2020-05-11
tensorflow-2.2.0 seems to work well with CUDA 10.2

WurmD:
TF 2.2 seems to work well with CUDA 10.2 in my server.

Should we make that official by updating the docs?


MarkDaoust:
See line #427

This is in a section called. "Tested Build configurations". It might work with plenty of other configurations, but be run a lot of tests on something before we make it official, so we can't accept this PR. 

Thanks for taking the time.

PR #1564, opened 2020-05-12
fix resolve minor typo

Gomyo:
Hello :)
I found a minor typo during tutorial.

It occurs NameError!

### DIFF
--- a/site/en/tutorials/load_data/csv.ipynb
+++ b/site/en/tutorials/load_data/csv.ipynb
@@ -420,7 +420,7 @@
       "outputs": [],
       "source": [
         "def pack(features, label):\n",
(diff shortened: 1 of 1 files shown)


MarkDaoust:
Hi, thanks for taking the time to send a PR. 

I don't see a problem with this notebook: 1) We have tests for this notebook, it is still passing. 2) I ran it in colab, it works. 3) That `label` argument refers to the function argument `label` defined just above:

```
def pack(features, label):
  return tf.stack(list(features.values()), axis=-1), label
```

Gomyo:
I made another typo 😭 It's my fault. sorry.

MarkDaoust:
No problem. Thanks for trying.

PR #1565, opened 2020-05-12
Definition of underfitting

PedroGFonseca:
My apologies if it's a wrong correction, but should the definition of underfitting not be that it has room for improvement in the training data? It currently says on the test data, which while technically correct does not seem to be what was intended.


PedroGFonseca:
PS: I'm actually an employee, not sure whether I should sign the https://cla.developers.google.com/ thing

MarkDaoust:
> PS: I'm actually an employee, not sure whether I should sign 

If you register your github account with [go/github](go/github) the google-bots will recognize you after that (if you configure git to use your google email)

PedroGFonseca:
Thanks @MarkDaoust , I'd rather not merge the two worlds if that's ok (I use github for personal stuff, and avoid mixing personal and work stuff). 

Is there a way around the CLA, or is it easier for someone else to just commit this change as themselves?

Harirai:
This issue has been fixed, it can be closed now.

PR #1566, opened 2020-05-12
Update text_classification_with_hub.ipynb

tommykoctur:
Added sigmoid activation to last Dense layer of the model, because below model in the description is said:  "Since this is a binary classification problem and the model outputs a probability (a single-unit layer with a sigmoid activation),"


tommykoctur:
As suggested by lamberta, closing

PR #1568, opened 2020-05-13
Removing "another" from the first option

PedroGFonseca:
The SavedModel solution comes first, but starts with "The SavedModel format is another way to serialize models". 

My guess is that there used to be an option before this one, but as it's the first it's odd to have an "another" in the sentence.


lamberta:
Ah, thanks.
The CLA bot isn't picking you up, may need to sign https://cla.developers.google.com/
or register your account at go/github ?

PR #1570, opened 2020-05-13

tommykoctur:
The last layer has linear activation, due to that this was changed (feel free to edit my English :) ):
1. 
The last layer is densely connected with a single output node. Using the sigmoid activation function, this value is a float between 0 and 1, representing a probability, or confidence level. For numerical stability, use the linear activation function that represents the logits.
to 

The last layer is densely connected with a single output node. For numerical stability, the `linear` activation function (default) that represents the logits is used. Another way around is to use the `sigmoid` activation function, this value is a float between 0 and 1, representing a probability, or confidence level.
2. 
(a single-unit layer with a sigmoid activation) to (a single-unit layer with a linear activation),


tommykoctur:
Another edit needed. Closing then.

PR #1574, opened 2020-05-15
Fix conda environment interaction syntax

daytonb:
The conda environment instructions had the "source" commands from virtualenv rather than the "conda" commands used by conda. I also changed some wording that called the conda environments virtualenvs.

### DIFF
--- a/site/en/install/pip.html
+++ b/site/en/install/pip.html
@@ -285,8 +285,7 @@
   is available.
 </aside>
 <p>
-  Create a new virtual environment by choosing a Python interpreter and making a
-  <code>./venv</code> directory to hold it:
+  Create a new conda environment, <code>venv<code>, including a Python interpreter and pip for installing the TensorFlow pip package:
 </p>
 {% dynamic if request.query_string.lang == "python2" %}
 <pre class="devsite-terminal devsite-click-to-copy">conda create -n <var>venv</var> pip python=2.7</pre>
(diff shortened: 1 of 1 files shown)


daytonb:
@googlebot I signed it!

lamberta:
Thanks, but this was intentional. Conda is not officially supported and the community maintained package often lags behind the release. But you can use Conda to install the pip package.

Even the page linked above (https://anaconda.org/conda-forge/tensorflow) shows the 1.13.2 version which seems a bit silly. If there's a more up to date page, we can change the link.

cc: @gunan

gunan:
I agree with @lamberta 
Conda is not officially supported, so our docs are focused on pip/virtualenv.

daytonb:
I didn’t change the instructions to use Conda to install the package only
to use the proper Conda commands for environment creation. If you want I’d
be willing to hop on a chat somewhere with you to discuss it in more
real-time manner.

daytonb:
Once again, I'd be more than happy to hop on gitter or some other chat platform to discuss your concerns. However, please review my documentation change. I did not advocate installing the tensorflow conda package. My commit fixes the following three problems:
1. You activate conda environments with `conda activate <envname>` not `source activate <envname>`
    * On Windows, there is not "source" command, so the current docs will get you an error from cmd or PowerShell
    * On Linux (and I presume mac) there is a source command, but there would be no "activate" file in the current directory, since conda stores the environments centrally (ala virtualenvwrapper) rather than in your project directory, so this command would lead to an error on Linux (and I presume mac)
2. You deactivate conda environments with `conda deactivate` not `source deactivate`
    * Again, on Windows there is no "source" command, so the current docs will get you an error message
    * On Linux (and I presume mac) there is a source command, but there is no "deactivate" file to source, so this command would lead to an error on Linux (and I presume mac). Virtualenv "activate" scripts define a "deactivate" _command_, it doesn't provide a deactivate script to source.
3. This one is less critical, but because conda environments are different than virtualenv environments, I think we should refer to them as conda environments, not virtual environments or virtualenvs for clarity.

gunan:
You are correct, the instructions for conda are confusing. Therefore, we decided to remove conda instructions, and just include a link to this page:
https://docs.anaconda.com/anaconda/user-guide/tasks/tensorflow/

As we do not continuously test or verify conda packages, our instructions just help add to user confusion.

daytonb:
Alright. That works. I wish I had found that page when I was looking to try TensorFlow for the first time because I don't have admin control to install the Microsoft redistributable DLLs to pip install it. I'm glad your link will make it a little bit easier for the next person in my boat :).

PR #1580, opened 2020-05-21
Fix tutorials/keras/overfit and underfit: failed to open TensorBoard

AseiSugiyama:
This PR fixes the issue about the failure to open the TensorBoard viewer in the section "View in TensorBoard".

image

To reproduce this, you may need to "Factory reset runtime".

This is caused by skipping to load the TensorBoard notebook extension. So I add a few lines of comments and codes to load it.

### DIFF
--- a/site/en/tutorials/keras/overfit_and_underfit.ipynb
+++ b/site/en/tutorials/keras/overfit_and_underfit.ipynb
@@ -1,21 +1,4 @@
 {
-  "nbformat": 4,
-  "nbformat_minor": 0,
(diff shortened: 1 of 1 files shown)


AseiSugiyama:
@lamberta How about this PR?

lamberta:
Hi AseiSugiyama@, I'm unclear.

Perhaps @wchargin can take a look? We typically don't require extensions for these notebooks but I guess this is already loaded in Colab? Is there a Jupyter/OSS equivalent?

PR #1585, opened 2020-05-28
switch to venv for virtual environment

grasskin:
Change install instructions and mentions of virtualenv to use venv.

@mihaimaruseac

### DIFF
--- a/site/en/install/pip.html
+++ b/site/en/install/pip.html
@@ -66,14 +66,12 @@
 <pre class="prettyprint lang-bsh">
 <code class="devsite-terminal">python3 --version</code>
 <code class="devsite-terminal">pip3 --version</code>
-<code class="devsite-terminal">virtualenv --version</code>
 </pre>
 
 <p>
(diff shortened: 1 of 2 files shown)


MarkDaoust:
OMG: 

https://twitter.com/drjtwit/status/1265561379253207042
https://www.liquidweb.com/kb/how-to-setup-a-python-virtual-environment-on-ubuntu-18-04/

lamberta:
Yeah, we explicitly mention support for "Ubuntu 16.04 or later" so these commands should run as listed and assume that.

Is `venv` pip-installable or do we need to install the Ubuntu package `python3-venv`? (I generally dislike relying on the distro packages because they're often out-of-date).

grasskin:
I couldn't find any mentions of it being pip-installable unfortunately, it seems users on certain Ubuntu will have to install it via apt-get

grasskin:
Confirmed ubuntu 18.04 does not have venv installed

lamberta:
Thanks for the update!

PR #1587, opened 2020-05-30
fix link on contribute/docs page

Molkree:
3 whitespaces were added after https:// in this commit https://github.com/tensorflow/docs/commit/d18c8352fd3fa57fbabdb239984e59c684a9fb52#diff-89f8d80a3fb313d1d00abbc99eb5c180
also extra typo at the bottom

### DIFF
--- a/site/en/community/contribute/docs.md
+++ b/site/en/community/contribute/docs.md
@@ -257,7 +257,7 @@ is a hosted notebook environment that makes it easy to edit—and run—notebook
 documentation. Notebooks in GitHub are loaded in Google Colab by passing the
 path to the Colab URL, for example,
 the notebook located in GitHub here:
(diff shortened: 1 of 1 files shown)


lamberta:
cc: @MarkDaoust ---in case [this commit](https://github.com/tensorflow/docs/commit/d18c8352fd3fa57fbabdb239984e59c684a9fb52#diff-89f8d80a3fb313d1d00abbc99eb5c180) indicates a problem with the automation

MarkDaoust:
I'm not sure what's up with the whitespace, but the "&#103;" may be to hide github path from the import script.

Molkree:
Should I return "g" as a `&#103;` symbol?

lamberta:
> Should I return "g" as a &#103; symbol?

Yeah, thanks. Maybe add an html comment after the paragraph to avoid confusion later:
`<!-- github.com path intentionally formatted to hide from import script. -->`

Molkree:
Oh, it probably was a mistake to include symbol code in the commit message, GitHub thought I'm linking PR.
P.S. And it's funny how it properly shows `&#103;` in the message here but converts it to just "g" letter in quoted text.

PR #1588, opened 2020-05-31
Remove Python 2 mentions

Molkree:
- Removed Python 2 from a bunch of pages where it seemingly was missed during EOL clean up.
- Changed Docker install page as `py3` tag was dropped. Added `custom-op` tag while there.
- [Fixed ](https://github.com/tensorflow/docs/compare/master...Molkree:remove_python_2?expand=1#diff-a2de1331b653964bd3ba0a0031add631L420)`tf.contrib.eager.num_gpus()` in install instructions as it's not available in TF2. Noticed it last year and it's still not fixed D: It even had [issue](https://github.com/tensorflow/tensorflow/issues/33396) created by someone.
- And some extra typo fixes I noticed.

### DIFF
--- a/site/en/install/docker.md
+++ b/site/en/install/docker.md
@@ -41,14 +41,17 @@ using the following format:
 :             : example\: *2.1.0*                                           :
 | `devel`     | Nightly builds of a TensorFlow `master` development         |
 :             : environment. Includes TensorFlow source code.               :
+| `custom-op` | Special experimental image for developing TF custom ops.    |
+:             : More info here: https://github.com/tensorflow/custom-op.    :
 
 Each base *tag* has variants that add or change functionality:
 
 | Tag Variants      | Description                                                                       |
 | ---               | ---                                                                               |
 | *`tag`*`-gpu`     | The specified *tag* release with GPU support. ([See below](#gpu_support))         |
-| *`tag`*`-py3`     | The specified *tag* release with Python 3 support.                                |
 | *`tag`*`-jupyter` | The specified *tag* release with Jupyter (includes TensorFlow tutorial notebooks) |
+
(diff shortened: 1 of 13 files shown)


lamberta:
Thanks, @Molkree 
Most of this looks correct, but wanted @gunan to take a look at some of the install updates and @angerson to take a look at the docker updates. Thanks

site/en/install/docker.md
```diff
@@ -41,15 +41,18 @@ using the following format:
...
 
 Each base *tag* has variants that add or change functionality:
 
 | Tag Variants      | Description                                                                       |
 | ---               | ---                                                                               |
 | *`tag`*`-gpu`     | The specified *tag* release with GPU support. ([See below](#gpu_support))         |
-| *`tag`*`-py3`     | The specified *tag* release with Python 3 support.                                |
 | *`tag`*`-jupyter` | The specified *tag* release with Jupyter (includes TensorFlow tutorial notebooks) |
 
+Versioned images <= 1.15.0 (1.x) and <= 2.1.0 (2.x) have Python 3 (3.5 for Ubuntu 16-based images; 3.6 for Ubuntu 18-based images) in images tagged "-py3" and Python 2.7 in images without "py" in the tag. Newer images have Python 3 in both tags. If you are using these containers for the first time, use the tags without "-py3" in them.
```
angerson, line 54: We now only publish Python 3 images, and no longer use the "-py" system at all.

angerson:
One small change to the Docker changes, but otherwise looks good to me. Thanks!

site/en/install/docker.md
Molkree: Oh, I understand, but it's just an explanation about older images which had different Python versions. I just copied it from [DockerHub](https://hub.docker.com/r/tensorflow/tensorflow/). I guess for new folks it might be a bit confusing and people specifically needing some old images would know the difference. Just to be sure, do you want me to completely remove this paragraph?

angerson: Yeah, that's a good idea. I updated the DockerHub page too.

Molkree:
@lamberta, is it still awaiting @gunan review?

PR #1590, opened 2020-06-01
nbfmt all

lamberta:
Cleanup: `nbfmt.py .`

### DIFF
--- a/site/en/guide/function.ipynb
+++ b/site/en/guide/function.ipynb
@@ -14,7 +14,7 @@
       "cell_type": "code",
       "execution_count": 0,
       "metadata": {
-        "cellView": "both",
+        "cellView": "form",
         "colab": {},
         "colab_type": "code",
         "id": "3jTMb1dySr3V"
--- a/site/en/guide/saved_model.ipynb
+++ b/site/en/guide/saved_model.ipynb
@@ -321,7 +321,7 @@
         "nohup tensorflow_model_server \\\n",
         "  --rest_api_port=8501 \\\n",
         "  --model_name=mobilenet \\\n",
-        "  --model_base_path=\"/tmp/mobilenet\" >server.log 2>\u00261\n",
+        "  --model_base_path=\"/tmp/mobilenet\" >server.log 2>&1\n",
         "```\n",
         "\n",
         "  Then send a request.\n",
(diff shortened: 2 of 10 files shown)


site/en/guide/function.ipynb
```diff
@@ -14,7 +14,7 @@
       "cell_type": "code",
       "execution_count": 0,
       "metadata": {
-        "cellView": "both",
+        "cellView": "form",
```
yashk2810, line 17: Why is the cellview changing here?

lamberta: This is what hides the license text and someone must have left it open. Another victory for nbfmt!

site/en/guide/variable.ipynb
```diff
@@ -415,7 +414,7 @@
   "metadata": {
     "colab": {
       "collapsed_sections": [],
-      "name": "intro_to_variables.ipynb",
+      "name": "variable.ipynb",
```
yashk2810, line 417: How was this changed?

site/en/guide/saved_model.ipynb
```diff
@@ -321,7 +321,7 @@
         "nohup tensorflow_model_server \\\n",
         "  --rest_api_port=8501 \\\n",
         "  --model_name=mobilenet \\\n",
-        "  --model_base_path=\"/tmp/mobilenet\" >server.log 2>\u00261\n",
+        "  --model_base_path=\"/tmp/mobilenet\" >server.log 2>&1\n",
```
MarkDaoust, line 324: I bet internal Colab will put these back if we don't add them to the copybara import script.

site/en/guide/variable.ipynb
MarkDaoust: It sets the colab "name" to the file-name.

lamberta: How or why? This is working as intended---it matches the filename

site/en/guide/saved_model.ipynb
lamberta: it will, and I agree we should (or at least it would be nice)

site/en/guide/variable.ipynb
yashk2810: Why does it do that?

What if I want the name to be what I set it to?

site/en/guide/saved_model.ipynb
MarkDaoust: cl/314171711

site/en/guide/variable.ipynb
lamberta: Why? `intro_to_variables.ipynb` does not exist, someone didn't update it when they moved the file. This isn't the title.

PR #1592, opened 2020-06-02
Add "flush=True" to progress-bar print statements

HinTak:
Due to terminal I/O buffering, the dots came out in batches when without the "flush=True" argument. This was confusing, and defeats the purpose of the dots being a progress bar for inexperienced users.


8bitmp3:
Hi @HinTak, thanks for the PR and sorry about the wait.

Just to recap - your proposal is:

```diff
...
for n in range(epochs):
  for m in range(steps_per_epoch):
    step += 1
    train_step(image)
-     print(".", end='')
+     print(".", end='', flush=True)
  display.clear_output(wait=True)
  display.display(tensor_to_image(image))
  print("Train step: {}".format(step))
...
```

That’s a very good point. To flush the buffer, especially when the `end` parameter is specified, it can be useful to add the `flush=True` argument inside `print()`.

I ran the code in a notebook and, when running the `for` training loop in the notebook, the output appeared to be very similar regardless of `flush=True` being used in the `print()` statement.

Let’s leave the code cell as it is for now to keep things simpler and revisit if users experience noticeable issues.

I'll close this PR for now - we can always revisit it. Thanks again and feel free to submit new contributions!

HinTak:
@8bitmp3 the final outcome is indeed identical, with or without flush. The difference is in the timing: e.g. If the total run takes 20 minutes, and it is supposed to do a 5% progress (ie every minute), with the flush, one statement indeed gets printed every minute; without the flush, all 20 statements come out at the end of 20 minute - that makes a "progress" print statement almost useless. So I ask you to reconsider.

8bitmp3:
Absolutely @HinTak! Thanks for the awesome ideas. Looping in @MarkDaoust.

PR #1594, opened 2020-06-04
Fix comment in autodiff.ipynb

dnotz:
In the comment it said 'dz_dx', however 'dz_dy' is meant


dnotz:
@googlebot I signed it!

PR #1596, opened 2020-06-05
Nbfmt outputs

lamberta:
nbfmt: check for private outputs before requiring field.

Run: `./tools/nbfmt.py site/`


tools/nbfmt.py
```diff
@@ -128,17 +128,17 @@ def clean_cells(data) -> None:
...
-    if private_outputs:
+    # Clear any code cell outputs if notebook set to private outputs.
+    if private_outputs and cell.get("outputs"):
       has_outputs = True
       # Clear code outputs
       cell["execution_count"] = 0
       cell["outputs"] = []
 
+    # Ensure outputs field exists since part of the nbformat spec.
+    if cell.get("outputs", None) is None:
```
yashk2810, line 139: Why move it?

lamberta: because it's triggering a warning it shouldn't (from chat)

PR #1600, opened 2020-06-13
fix minor typo - change inherantly to inherently

frankctan:


frankctan:
@googlebot I signed it!

PR #1601, opened 2020-06-16
Fixed grammatical and reference typos

samholt:
Fixed a reference in `site/en/tutorials/structured_data/time_series.ipynb`, referencing a task that appears above the text, however it incorrectly says below.

Also fixed grammar in `site/en/guide/keras/functional.ipynb`, making it clearer to the reader, changing `a` to the `the`, from `Shared layers are layer instances that are reused multiple times in a same model` to `Shared layers are layer instances that are reused multiple times in the same model`.


MarkDaoust:
Hi, @samholt 

Thanks for tanking the time to send us these fixes, but we're in the process of switching tensorflow.org/guide/keras to pull the updated guides: from https://github.com/keras-team/keras-io/tree/master/tf

We're going to be **deleting** this directory. 

This change, and all future changes should go to keras-io repository.

MarkDaoust:
Oh, wait, there are two files.

I'll merge it, but the `guide/keras` change will never gewt synced to keras-io.

PR #1602, opened 2020-06-18
Fix minor issues

lc0:


lc0:
Thanks for the great notebook, btw!

PR #1603, opened 2020-06-23
Fix numpy version to pre-1.19.0

adk9:
Fix numpy to pre-1.19.0 because of breaking ABI change in numpy 1.19.0 (numpy/numpy#15355). Because of this change, TF source builds break with numpy 1.19.0.

See tensorflow/tensorflow#40688.

### DIFF
--- a/site/en/install/source.md
+++ b/site/en/install/source.md
@@ -37,7 +37,7 @@ Install the TensorFlow *pip* package dependencies (if using a virtual environmen
 omit the `--user` argument):
 
 <pre class="prettyprint lang-bsh">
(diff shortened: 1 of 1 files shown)


adk9:
@googlebot I signed it!

lamberta:
Thanks for the fix, @adk9 , but this seems unfortunate :/

mihaimaruseac:
Follows https://github.com/tensorflow/tensorflow/commit/79518facb4b857af9d9d5df2da463fdbf7eb0e3e

It's too late to fix the ABI breakage now in the 2.3 release but we can do it for 2.4

PR #1605, opened 2020-06-23
fixed minor grammatical error

xrijo:
"These are initial guesses are not great." -> "These initial guesses are not great."


xrijo:
@googlebot I signed it!

PR #1608, opened 2020-06-29
Remove duplicate sentence & fix a typo

rickiepark:
>A v1.variable_scope is effectively a layer of its own. So rewrite it as a tf.keras.layers.Layer. See the guide for details.
>
>The general pattern is:
>
>- Collect layer parameters in __init__.
>- Build the variables in build.
>- Execute the calculations in call, and return the result.
>
>The v1.variable_scope is essentially a layer of its own. So rewrite it as a tf.keras.layers.Layer. See the guide for details.

First sentence is almost duplicate with last sentence. I just remove the first sentence.

And fix a typo "how to build a tool fr a similar use-case." --> "how to build a tool **for** a similar use-case."

Thanks. :)

### DIFF
--- a/site/en/guide/migrate.ipynb
+++ b/site/en/guide/migrate.ipynb
@@ -647,8 +647,6 @@
       "source": [
         "To convert this code, follow the pattern of mapping layers to layers as in the previous example.\n",
         "\n",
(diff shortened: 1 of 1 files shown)


MarkDaoust:
Thanks again Haesun.

PR #1609, opened 2020-06-29
Fix output of manual device placement on CPU:0

heaths:
The docs section for manually doing operations on a device (CPU:0, in this case) has the wrong output.


heaths:
@googlebot I signed it!

heaths:
Ping

lamberta:
Thanks for the ping.
I'm wondering why the outputs are saved and the notebook runtime isn't set to GPU. @yashk2810 @MarkDaoust Can I convert this notebook so we can test it?

yashk2810:
The `matmul` operation is placed on a GPU hence the output will contain `GPU:0`. The tensors `a` and `b` are placed on a CPU hence for the matmul op that triggers a copy from CPU -> GPU for the tensors. 

Please run this notebook on a GPU since the notebook is supposed to run on a GPU (gpu.ipynb).

lamberta (dismissed):
Ok. Yash is correctly points out that matmul does in on a GPU here.
To make this a little clearer, I add a code comment and set the notebook to use a GPU on Colab.

heaths:
Merge conflict resolved. Is this ready to merge?

MarkDaoust (approved):

MarkDaoust:
> I'm wondering why the outputs are saved

```
Num GPUs Available:  2
```

This notebook looks at some of the details around having multiple GPUs and can't easily be re-run to get those outputs.

PR #1611, opened 2020-07-04
enclosed numpy<1.19.0 in ' ' in the dependencies of build from source for Linux page.

electricSamarth:
If the older command was used bash gave the following error:
1.19.0: No such file or directory


electricSamarth:
> If the older command was used bash gave the following error:
> 1.19.0: No such file or directory

@googlebot I signed it!

PR #1618, opened 2020-07-09
Add clang-format install instructions to docs

mhansen:
Towards https://github.com/tensorflow/tensorflow/issues/41077. Not sure which flags you invoke `clang-tidy` with though.

`clang-format` is also available since Ubuntu 16.04: https://packages.ubuntu.com/search?keywords=clang-format


mhansen:
PTAL.

PR #1623, opened 2020-07-15
Threshold to be set 0 when using binary accuracy with raw prediction values (from_logits=True)

Dynmi:
This is from [#41413](https://github.com/tensorflow/tensorflow/issues/41413) in *tensorflow/tensorflow* by HaraBeron.

In this tutorial raw prediction values (form_logit=True) are used. So we have negative values and positive values, while
"prediction will be treated as a logit, or a raw prediction value. Positive numbers predict class 1, negative numbers predict class 0."

According to [https://www.tensorflow.org/api_docs/python/tf/keras/metrics/BinaryAccuracy](https://www.tensorflow.org/api_docs/python/tf/keras/metrics/BinaryAccuracy)
As per default, threshold value to classify is 0.5:

>tf.keras.metrics.BinaryAccuracy(
>    name='binary_accuracy', dtype=None, threshold=0.5
>)


MarkDaoust:
Thanks!

PR #1624, opened 2020-07-16
Remove unnecasary cast to tf.Variable

guoci:


MarkDaoust (approved):
given the name var_x, I switched it to a variable instead of a constant.

PR #1626, opened 2020-07-20
nbfmt all the things

lamberta:
Format with the latest `nbfmt`:

```
$ python3 -m pip install -U --user git+https://github.com/tensorflow/docs

$ python3 -m tensorflow_docs.tools.nbfmt .
```

### DIFF
--- a/site/en/guide/intro_to_graphs.ipynb
+++ b/site/en/guide/intro_to_graphs.ipynb
@@ -14,7 +14,7 @@
       "cell_type": "code",
       "execution_count": null,
       "metadata": {
-        "cellView": "both",
+        "cellView": "form",
         "colab": {},
         "colab_type": "code",
         "id": "yOYx6tzSnWQ3"
--- a/site/en/guide/migrate.ipynb
+++ b/site/en/guide/migrate.ipynb
@@ -2159,7 +2159,6 @@
     },
     "kernelspec": {
       "display_name": "Python 3",
-      "language": "python",
       "name": "python3"
     }
   },
--- a/site/en/guide/upgrade.ipynb
+++ b/site/en/guide/upgrade.ipynb
@@ -671,7 +671,6 @@
     },
     "kernelspec": {
       "display_name": "Python 3",
-      "language": "python",
       "name": "python3"
     }
   },
--- a/site/en/r1/tutorials/eager/automatic_differentiation.ipynb
+++ b/site/en/r1/tutorials/eager/automatic_differentiation.ipynb
@@ -311,7 +311,6 @@
     },
     "kernelspec": {
       "display_name": "Python 3",
-      "language": "python",
       "name": "python3"
     }
   },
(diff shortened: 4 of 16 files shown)


site/en/guide/migrate.ipynb
```diff
@@ -2159,7 +2159,6 @@
     },
     "kernelspec": {
       "display_name": "Python 3",
-      "language": "python",
```
yashk2810, line 2162: Why delete this?

If colab is adding this, it will always be added.

site/en/guide/intro_to_graphs.ipynb
```diff
@@ -14,7 +14,7 @@
       "cell_type": "code",
       "execution_count": null,
       "metadata": {
-        "cellView": "both",
```
yashk2810, line 17: Is colab doing this?

I think I have seen this in every PR where you nbfmt. Maybe wanna add a copybara transform so that this doesn't happen everytime.

lamberta: Wollf didn't fix it when I pointed it out in his CL 🤷‍♂️

yashk2810: Maybe add a copybara transform so that this doesn't happen again?

lamberta: But for automation, if there's a way to add nbfmt to the export then it would handle this

yashk2810: The automation is copybara. I don't know if its possible to run a script but for now, let's add this to copybara so that the diff churn is minimum.

lamberta: I think that's ot-of-scope for this :) I"m just running nbfmt
ANd I'd rather not rebuild nbfmt in the form of copybara transforms

yashk2810: Yeah.. I am not saying to add the copybara transform here but in a separate CL (not PR).

> nbfmt in the form of copybara transforms

I am not asking you to do that. The purpose of nbfmt is to reduce diff churn. For cell-view that's not the case. There are many similar copybara transform running of `<` and `>`, so adding this shouldn't be much of an issue. Also, majority of the recents nbfmt PR's consist of the cell-view transformation.

lamberta: > majority of the recents nbfmt PR

I don't believe this is the case. you're just noticing this in the same notebook that was called out and not fixed. please manage the copybara config how you like with @MarkDaoust , it's added to nbfmt to catch issues that slip by (like it did).

MarkDaoust: > Maybe add a copybara transform so that this doesn't happen again?

Colab's just saving the view state. Adding it to copybara is a bit trickier because we want all transforms to be reversible, but this isn't something you'd want to add back. There's probably a way to add come non-reversible transforms but that's more complicated.

If you want to enforce nbfmt on CLs you could add a pytest.

site/en/guide/migrate.ipynb
lamberta: This is a result of c/322176921
Colab is not adding it. Other environments are and it's not necessary

PR #1627, opened 2020-07-21
Softmax fix in CNN

swghosh:
Fixed softmax activation (added missing `activation` argument to last `Dense` layer) in CNN tutorial.


guoci:
@swghosh, the loss function is `SparseCategoricalCrossentropy(from_logits=True)`. So, the output layer is in logits and the softmax activation is not required.

swghosh:
Ohh, right @guoci. Sorry, I had overlooked that.

PR #1628, opened 2020-07-22
Fixed predictions output in the documentation

MichaelGrigoryan25:
The predictions part wasn't getting printed, so I fixed that :)


MichaelGrigoryan25:
@googlebot I signed it!

PR #1630, opened 2020-07-23
Fix inaccurate integrated gradients computation in tutorial

bentyeh:


bentyeh:
@googlebot I signed it!

bentyeh:
If someone who is very familiar with the integrated gradients method could review the math, that would be great.

Here's a summary of the changes I made:
- `integrated_gradients()` is now robust to batch sizes. Previously, if `m_steps % batch_size == 1`, then the function would fail because the function `integral_approximation()` called by `integrated_gradients()` requires at least 2 gradient values.
- `integrated_gradients()` now correctly accounts for the area between the last position in a previous batch and the first position in a new batch
- `integrated_gradients()` now correctly aggregates integrated gradients across batches.
- Accurate computation of integrated gradients allows reduction in number of steps required for examples.
- Fixed some typos.

Changes above were adapted from changes made to the lengthier version of this tutorial. See https://github.com/GoogleCloudPlatform/training-data-analyst/pull/1122.

MarkDaoust:
@dougkelly, can you review these changes?

dougkelly:
Yes, I will take a look.

dougkelly:
Hi @bentyeh,

To echo GoogleCloudPlatform/training-data-analyst#1122, thanks again for the excellent feedback and engagement. Can you please add me to your fork so I can push changes to this PR like the longer form version fork. I will be accepting many of your changes with a few alternative fixes.

Cheers,
Doug

bentyeh:
@dougkelly Thanks for taking a look! I've added you to my forked repos.

dougkelly:
Hi @bentyeh,

I pushed fixes and responded in more detail to your PR on the longer version at GoogleCloudPlatform/training-data-analyst#1122. I updated this PR as well accordingly. 

- Alternative pushed. @tf.function removed. Batch gradients, concat gradient batches, keeping just the averaging reduction.
- Alternative pushed. @tf.function removed. Batch gradients, concat gradient batches, keeping just the averaging reduction.
- Verified. `m_steps` were significantly reduced with corrections to batching computation.
- Accepted. Thanks for catching typo on the image height.

Please take a look and let me know if you have any questions and this should be good to submit.

bentyeh:
The only concern I have is about the removal of `@tf.function`, as explained in the PR discussion on the longer version: https://github.com/GoogleCloudPlatform/training-data-analyst/pull/1122#issuecomment-668967738. Otherwise, looks good. Thanks @dougkelly!

dougkelly:
Comments addressed in longer IG tutorial in more detail and merged https://github.com/GoogleCloudPlatform/training-data-analyst/pull/1122. 

To document developments on this PR as well:
- Alternative implementation pushed with @tf.function + TensorArray. This includes workarounds for b/163121904 and b/163083602 after testing in default TF2.3 and TF2.2 Colab and JupyterLab environments and review with autograph team members. This approach was agreed upon to provide the best mix of speed and memory performance improvements, highlighting TF2 features, and readability for learners.

Thanks for the contributions @bentyeh!

@MarkDaoust this is ready for final review.

site/en/tutorials/interpretability/integrated_gradients.ipynb
```diff
@@ -964,34 +965,42 @@
...
+        "  \n",
+        "  # Stack path gradients together row-wise into single tensor.\n",
+        "  total_gradients = gradient_batches.stack()\n",
         "\n",
-        "    # 2. Generate interpolated images\n",
-        "    batch_interpolated_inputs = interpolate_images(baseline=baseline,\n",
-        "                                                   image=image,\n",
-        "                                                   alphas=batch)\n",
+        "  # 4. Integral approximation through averaging gradients.\n",
+        "  avg_gradients = integral_approximation(gradients=total_gradients)\n",
```
MarkDaoust, line 998: Just to be sure I understand:

1. The `TensorArray` is necessary here because you want to get the full list of gradients out of the loop.
1. You want the whole list so that the edges between  batches don't mess with the trapezoidal integration (you loose the trapezoid between the last element of one batch and the first element of the next.
1. It would have been messy to do the `trapz` in the loop, and carry-over the last element to the next batch. 
1. It would have been less precise to do a rectangular integration.

Okay.

dougkelly: 1. Yes exactly. In addition to working performantly with @tf.function, a side benefit of using TensorArray on this tutorial is highlighting how to add different batch sizes to a TensorArray with scatter() which has been the subject of a few stackoverflow posts. One of the autograph engineers made a really nice analogy between scatter() and a Python lists extend() method and write() to a list's append() method.
2. Yes. Particularly for this version of the tutorial, I felt adding additional code for just this integration method and explanatory text on the endpoint batching issue without covering or visualizing different integration methods like the longer version would be difficult to understand for learners without consulting the longer version of the tutorial and/or alternative resources.
3. bentyeh actually implemented this approach and it is definitely a valid alternative. You can initialize a last_batch tensor, keep track of the last batch by enumerating the batch, have an if statement to check the method, concat the batches to make sure you have the endpoints, do a tf.math.reduce_sum on each gradient batch, add the gradient batches together and divide the total gradients outside the loop by m_steps. This added additional lines of code and explanatory text and the decision to go with an alternative really came down to teachability and readability as opposed to just performance. I felt it was important to highlight to learners that they are averaging across the image gradients of the interpolated path in line with the original paper author's implementation and presentations and keep the focus on IG not the integration methods. In testing, this approach was evaluated with 6 others and the resulting implementation came out significantly faster but this is just a side benefit to teachability and readability considerations and there might even be additional performance optimizations missed as well.
4. Yes the longer version of the tutorial has plots highlighting how Riemann Trapezoidal generally converges quicker and with fewer m_steps as well as how it requires the interval endpoints compared to alternative Riemann methods. This was identified in experiments by the original authors in section 4 of https://arxiv.org/abs/1908.06214 and is how they initially implemented it. However, there are still edge case functions where this might not be the case so it is still important to point out alternatives in the longer version. I have even seen interpretability libraries with Gauss-Legendre quadrature rule integral approximation as well for IG and other gradient-based interpretability methods.

PR #1639, opened 2020-07-29
Fix dataset for stateful RNN use

guoci:
as documented in https://www.tensorflow.org/guide/keras/rnn#cross-batch_statefulness


MarkDaoust:
Hi. 

Thanks, thast's a cool trick. But we would like to keep this tutorial pretty simple, and this is a bit of a distraction.

I'd rather just drop the "stateful" flag here and, at most, link to that cross-batch statefulness section.

PR #1640, opened 2020-08-03
Fix errors about index lacking

coderkodor:
The "predictions_array" needed to be indexed with [i] for latter use.


coderkodor:
> @googlebot I signed it!

@googlebot I signed it!

MarkDaoust:
Thanks. But please test your changes before submitting.

PR #1642, opened 2020-08-04
Update docs to reflect error when PTX is not present

manivaradarajan:
Update GPU docs to document error that users may see when PTX is absent. 
This can happen because TF >= 2.3 has dropped PTX to reduce TF pip package size.

Fixes #41892

### DIFF
--- a/site/en/install/gpu.md
+++ b/site/en/install/gpu.md
@@ -34,7 +34,7 @@ For releases 1.15 and older, CPU and GPU packages are separate:
 
 ## Hardware requirements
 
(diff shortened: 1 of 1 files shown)


yashk2810:
Thanks!

lamberta:
Thanks. Added edit to move into a separate note

sanjoy:
When will this show up on tensorflow.org?

yashk2810:
The jobs are automated. You can see the schedule at go/tf-projects. It should be up today around 6:30 pm

PR #1649, opened 2020-08-16
fix bug - variable not used as intended

ma7555:
encoding_dims variable is not used, so encoded dimensions will not change regardless of how the class is initialized.


ma7555:
@googlebot I signed it!

PR #1650, opened 2020-08-16
Only link texts, not actual links, to TF 2.3.0 Java artefacts updated

mikaelhg:
Whoever updated the Tensorflow 2.3.0 Java artifact links only updated the link texts, not the actual links, which still pointed to an old version.


mikaelhg:
You can just reproduce these minor changes, rather than use the patch, as I going through the CLA process would be a waste of my time, just for a few rows.

MarkDaoust:
Thanks, but It looks like this it getting fixed by #1654.

PR #1652, opened 2020-08-23
Replace tf.Example with tf.train.Example

hoenickf:
tf.Example does not exist and is now replaced with tf.train.Example

### DIFF
--- a/site/en/tutorials/load_data/tfrecord.ipynb
+++ b/site/en/tutorials/load_data/tfrecord.ipynb
@@ -41,7 +41,7 @@
         "id": "HQzaEQuJiW_d"
       },
       "source": [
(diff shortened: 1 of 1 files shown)


site/en/tutorials/load_data/tfrecord.ipynb
```diff
@@ -894,7 +894,8 @@
       },
       "outputs": [],
       "source": [
-        "# Write the `tf.Example` observations to the file.\n",
+        "# Write the `tf.train.Example
```
lamberta, line 897: Problem loading JSON. Maybe here

MarkDaoust: Fixed.

PR #1657, opened 2020-08-26
fix: only use tf.keras instead of keras

tgkei:
In <a href="https://www.tensorflow.org/tutorials/keras/classification?hl=en">Basic classification</a>,
```keras``` is used in two different way. first is using ```tf.keras``` imported by ```import tensorflow as tf```, second way is importing ```from tensorflow import keras``` and use ```keras```. It seems unclear and might be hard to understand the difference between ```keras``` library and ```keras``` which is imported by ```from tensorflow import keras``` to new user if ```tf.keras``` is handled differently in same doc, so changed to only use ```tf.keras```

### DIFF
--- a/site/en/guide/ragged_tensor.ipynb
+++ b/site/en/guide/ragged_tensor.ipynb
@@ -63,6 +63,17 @@
       "cell_type": "markdown",
       "metadata": {
         "colab_type": "text",
+        "id": "5DP8XNP-6zlu"
+      },
+      "source": [
+        "\n",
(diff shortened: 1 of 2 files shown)


tgkei:
@googlebot I signed it!

MarkDaoust (approved):
We don't have a global policy for this, but being consistent within one doc is definitely good.

Thanks.

MarkDaoust:
🤔

It looks like I'm unable to merge this, because the source fork has been deleted.

PR #1659, opened 2020-08-30
Update gpu.ipynb

marwan116:
Closes https://github.com/tensorflow/tensorflow/issues/42708

- move `c = tf.matmul(a, b)` so that the operation is called within the `tf.device` context.


marwan116:
@googlebot I signed it!

marwan116:
I see this notebook lint failure:
```
Fail | google::second_person | Prefer second person instead of first person: https://developers.google.com/style/person
```

which is strange given my change was only in code and not in the notebook text, so I presume this lint rule was introduced after this notebook was committed to the repo and wasn't ever applied retroactively ? 

Also the lint rule doesn't state which line/cell of the notebook violates the rule so it will be hard for me to fix this without additional pointers ...

MarkDaoust (approved):

MarkDaoust:
Thanks, but this is intentional. 

It's putting the constants on the CPU, and executing the matmul on the GPU.

PR #1662, opened 2020-09-03
Fix visualization

Mandroide:
image


MarkDaoust:
Thanks for the PR, but this is running fine in colab and displaying fine on tensorflow.org.

PR #1663, opened 2020-09-04
Clean up the table in docker.md

sanjoy:
The rendering is broken today for the last row on the website.  On github preview the other rows are rendered incorrectly as well.


MarkDaoust (commented):
That'll work. Thanks.

It turns out the problem was that you need to escape colons inside markdown table continuation rows.

PR #1665, opened 2020-09-08
Update actor_critic.ipynb

Dynmi:
This is from Issue [43029](https://github.com/tensorflow/tensorflow/issues/43029).
You should update apt source before installing new package, which is a good habit.

### DIFF
--- a/site/en/tutorials/reinforcement_learning/actor_critic.ipynb
+++ b/site/en/tutorials/reinforcement_learning/actor_critic.ipynb
@@ -174,7 +174,7 @@
       "source": [
         "%%bash\n",
         "# Install additional packages for visualization\n",
-        "sudo apt-get update",
+        "sudo apt-get update\n",
         "sudo apt-get install -y xvfb python-opengl > /dev/null 2>&1\n",
         "pip install pyvirtualdisplay > /dev/null 2>&1\n",
         "pip install git+https://github.com/tensorflow/docs > /dev/null 2>&1"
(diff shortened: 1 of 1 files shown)


site/en/tutorials/reinforcement_learning/actor_critic.ipynb
```diff
@@ -174,6 +174,7 @@
       "source": [
         "%%bash\n",
         "# Install additional packages for visualization\n",
+        "sudo apt-get update\n",
```
bhack, line 177: In what kind of setup is this required? In a container?

8bitmp3: If the notebook is designed for running in Colab first, this may not be necessary. I've tested it and it works out of the box. However, it's a valid point. Is there a final decision on this? @lamberta @MarkDaoust

yashk2810: We do run this inside a container for testing and publishing.

You can remove the `apt-get update` since we do that in the container. If we hit any errors, we will add the `apt-get update`.

MarkDaoust: @yashk2810: That's the whole PR.

yashk2810: Okay, since it runs end-to-end, I am going to close the PR.

PR #1667, opened 2020-09-09
Fix graph optimizzation link

bhack:
Fixes: https://github.com/tensorflow/tensorflow/issues/43089


MarkDaoust (approved):

bhack:
@MarkDaoust is there anyhting new commited on master in the meantime? Cause I see a conflict now.

MarkDaoust:
It looks like this has been fixed in the mean time.

bhack:
@MarkDaoust P.s. just two times in a day. We cannot compete with direct commits.

PR #1674, opened 2020-09-15
Sort classes with the same order as TOC

bhack:
Fix: https://github.com/tensorflow/tensorflow/issues/43224 /cc @yashk2810

### DIFF
--- a/tools/tensorflow_docs/api_generator/generate_lib.py
+++ b/tools/tensorflow_docs/api_generator/generate_lib.py
@@ -167,7 +167,7 @@ class Module(TocNode):
   @property
   def children(self):
     return sorted(
-        self._children, key=lambda x: (x.full_name.upper(), x.full_name))
+        self._children, key=lambda x: (x.full_name.lower(), x.full_name))
 
   @property
   def submodules(self):
--- a/tools/tensorflow_docs/api_generator/generate_lib.py
+++ b/tools/tensorflow_docs/api_generator/generate_lib.py
@@ -167,7 +167,7 @@ class Module(TocNode):
   @property
   def children(self):
     return sorted(
-        self._children, key=lambda x: (x.full_name.lower(), x.full_name))
+        self._children, key=lambda x: (x.full_name))
 
   @property
   def submodules(self):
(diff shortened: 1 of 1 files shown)


tools/tensorflow_docs/api_generator/generate_lib.py
```diff
@@ -167,7 +167,7 @@ def title(self):
   @property
   def children(self):
     return sorted(
-        self._children, key=lambda x: (x.full_name.upper(), x.full_name))
+        self._children, key=lambda x: (x.full_name.lower(), x.full_name))
```
yashk2810, line 170: I would just go with `sorted(x.full_name`.

Can you also run this on tensorflow and show me a diff of what the changes are before and after on the _toc.yaml file?

bhack: Ok. Check https://gist.github.com/bhack/d62c1fd9b5373fd5d2275f34d3ab0ae1

tools/tensorflow_docs/api_generator/generate_lib.py
```diff
@@ -167,7 +167,7 @@ def title(self):
   @property
   def children(self):
     return sorted(
-        self._children, key=lambda x: (x.full_name.upper(), x.full_name))
+        self._children, key=lambda x: (x.full_name))
```
yashk2810, line 170: The below code is consistent with Google's python linter.

```
return sorted(self._children, key=lambda x: x.full_name)
```

This sorting order is consistent with the All symbols page so LGTM

tools/tensorflow_docs/api_generator/generate_lib.py
```diff
@@ -167,7 +167,7 @@ def title(self):
   @property
   def children(self):
     return sorted(
-        self._children, key=lambda x: (x.full_name.upper(), x.full_name))
+        self._children, key=lambda x: x.full_name)
```
yashk2810, line 170: Can you put this line on the same line as `return sorted`? (Google's linter will complain if you don't do that)

So 

`return sorted(self._children, key=lambda x: x.full_name)` 

instead of 

```
return sorted(
    self._children, key=lambda x: x.full_name)
```

bhack (commented):
I am mobile now.. so I cannot run lint.. if can fix on your side

yashk2810 (approved):

PR #1676, opened 2020-09-17
Fix minor typos

ghabault:
Just found few typos in the text (some might require to reproduce the outputs and performance figures)


ghabault:
@googlebot I signed it!

MarkDaoust (approved):

PR #1677, opened 2020-09-21
Update __init__.py

KaiserLeave:
line 40: min(2 * width + 1, len(values)) #fix situation where the number of epoch is too small in training to produce a valid plot


KaiserLeave:
@googlebot I signed it!

MarkDaoust:
Thanks for the fix!

MarkDaoust (approved):

aavishkarmishra (approved):

PR #1678, opened 2020-09-21
replace deprecated model.get_losses_for call by model.losses in migration guide

DorianKodelja:
See issue https://github.com/tensorflow/tensorflow/issues/43341


DorianKodelja:
@googlebot I signed it!

DorianKodelja:
@googlebot I fixed it.

DorianKodelja:
@lamberta Hi, I'm contacting you since you're among the requested reviewers:
The linting test failed on part of the migration guideline that I didn't modify. 
Should I fix it in the same PR or create a separated issue for these ?
In the first case, i'm not sure how to replace "Native" and replacing "We" by "you"  doesn't feel right either

MarkDaoust:
@tomerk, @omalleyt12 : Replacing `get_losses_for` with `.losses` in a custom estimator `model_fn`, is this change correct?

tomerk:
This isn't a safe change to make. Estimators w/ custom model_fn's are generally pretty disrecommended, and basically use tf1 graphs/are in a deprecated-ish state. As such, they'll only work w/ the deprecated tf1-style methods especially when it comes to things like capturing losses.

Maybe this is something that should be made explicitly clearer in the migration guide? "Here's how you can use keras in an estimator for transitional migration purposes, but you really shouldn't be using estimators w/ custom model_fns anyway"?

MarkDaoust:
Thanks @tomerk.

@DorianKodelja, can you make that change instead?

MarkDaoust:
I'm sending a fix for that.

PR #1683, opened 2020-09-24
Add nightly GitHub Actions workflow to sync Keras guides.

lamberta:
Adds nightly workflow to generate notebooks and sync the Keras guides from https://github.com/keras-team/keras-io/tree/master/tf

These updates will live on the `snapshot-keras` branch.
Tested in https://github.com/lamberta/tf-docs/runs/1162780925?check_suite_focus=true

Commit: https://github.com/lamberta/tf-docs/commits/snapshot/keras

ref: b/169166095

### DIFF
--- a/.github/workflows/bot-nightly.yaml
+++ b/.github/workflows/bot-nightly.yaml
@@ -1,0 +1,65 @@
+# Nightly jobs run by a bot collaborator.
+name: Nightly jobs
+on:
+  repository_dispatch:
+    types: [nightly]
+
+jobs:
+  snapshot-source:
+    name: Update Keras guides
+    if : ${{ github.actor == 'tfdocsbot' }}
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        repository: keras-team/keras-io
+        path: keras-io
+    - uses: actions/checkout@v2
+      with:
+        ref: snapshot-keras
+        fetch-depth: 0
+        path: docs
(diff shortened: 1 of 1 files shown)


.github/workflows/bot-nightly.yaml
```diff
@@ -0,0 +1,64 @@
...
+        cd keras-io/scripts/
+        echo "[${GITHUB_WORKFLOW}] Generate Keras guides ..."
+        python3 autogen.py generate_tf_guides  # Output to ../tf/
+        echo "[${GITHUB_WORKFLOW}] Format notebooks ..."
+        python3 -m tensorflow_docs.tools.nbfmt ../tf/
+    - name: Sync docs repo
+      env:
+        KERAS_GUIDES_DIR: site/en/guide/keras/
+      run: |
+        rsync --archive --del --checksum ./keras-io/tf/ "./docs/${KERAS_GUIDES_DIR}"
```
yashk2810, line 50: If we are deleting `keras-io/tf` then do we need to rsync?

yashk2810 (approved):

.github/workflows/bot-nightly.yaml
lamberta: Added `mkdir -p` if we delete `keras-io/tf/`.
rsync just makes it easy to delete notebooks over here if they delete notebooks over there.

yashk2810 (approved):

PR #1689, opened 2020-09-29
CyclicGAN: corrected identity loss

sarthakforwet:
As compared to the [original paper](https://arxiv.org/abs/1703.10593), it seemed that the identity loss defined currently should be changed.


MarkDaoust (approved):
Thanks for taking the time to make a PR and pointing out that this section is really unclear.

that equation is correct, as is, I've reverted it and added a sentence to clarify.

PR #1692, opened 2020-10-03
added option to pickle tokenizer

alt-nikitha:
I wanted to separately use this model once i've completed training. So along with saving checkpoints, I thought it would be a nice option to include saving the tokenizer as well, since the dataset it is fitted on is extremely large to reload. The model can now be loaded from the checkpoints folder and the tokenizer can also be loaded with pickle. The model can be evaluated against images during deployment or testing phases independently.


MarkDaoust:
duplicate of #1693

PR #1693, opened 2020-10-03
added option to pickle

alt-nikitha:
Option to use pickle to save tokenizer object. This removes the need to reuse or save the large dataset whenever the model is evaluated.


MarkDaoust:
There's nothing wrong with this.

But I'm working on updating these docs to use: [tf.keras.layers.experimental.preprocessing.TextVectorization](https://www.tensorflow.org/api_docs/python/tf/keras/layers/experimental/preprocessing/TextVectorization?version=nightly) which will let us save and deploy the TextProcessing as part of the saved model.

So thanks, but this will just get deleted in a week or two if we merge it now.

PR #1696, opened 2020-10-04
DeepDream : use a single `tf.roll` op to shift the two dimensions instead of two ops

kobejean:
In the DeepDream tutorial, the `tf.roll` op is being used inefficiently by using 2 operations to roll two dimensions one at a time. The `tf.roll` supports the ability to shift multiple dimensions simultaneously so instead of doing:

```python
shift = tf.random.uniform(shape=[2], minval=-maxroll, maxval=maxroll, dtype=tf.int32)
shift_down, shift_right = shift[0],shift[1] 
img_rolled = tf.roll(tf.roll(img, shift_right, axis=1), shift_down, axis=0)
```

Why not do the equivalent in one `tf.roll` operation:

```python
shift = tf.random.uniform(shape=[2], minval=-maxroll, maxval=maxroll, dtype=tf.int32)
img_rolled = tf.roll(img, shift=shift, axis=[0,1])
```

That way we can showcase how to use `tf.roll` in a better way.


MarkDaoust (approved):
Thanks!

PR #1699, opened 2020-10-06
drop_remainder = True at tf.dataset.batch

edwardyehuang:
drop_remainder = True at tf.dataset.batch is necessary in TPU as it requires fixed dims, otherwise it may cause some op failed to compile such as dilation conv


MarkDaoust:
You do have a point about fixed shapes, I hesitate to include this since it doesn't matter for this example and would take some explaination. IIRC the `tf.function` will just retrace new shapes.

Do you have a short example of a case that does fail, where this is more important?

PR #1702, opened 2020-10-07
Switch off stateful RNN

guoci:
The sequence is not processed correctly for use of stateful RNN.
As documented in https://www.tensorflow.org/guide/keras/rnn#cross-batch_statefulness
Suggested in #1639.


MarkDaoust (approved):
Thanks.

PR #1709, opened 2020-10-12
update gpu software requirements docs

ymodak:
Fixes #https://github.com/tensorflow/tensorflow/issues/43933


lamberta (approved):
Thanks. Probably a less stable link but better for users, so let's try it.

PR #1643, opened 2020-08-05
dcgan.ipynb Architectural and hyperparameters change

arshagarwal:
The following architectural and parameter changes have been made in this pull request to corroborate with the Dc-Gan paper and also to improve performance:
- LeakyRelu in the Generator has been replaced by Relu.
- Dropouts in the Discriminator have been replaced by BatchNormalization.
- Adam optimizer's beta_1 has been changed to 0.5.
- Lastly, the weights have been initialized from Normal Distribution.

**Note: All these changes significantly improve training time, for MNIST number of epochs decrease from about 25 to 15, and similar improvements can be expected with bigger datasets.**
The changes have been saved in file named **Copy_of_dcgan.ipynb** in the same folder as existing file.

PR #1715, opened 2020-10-14
Added Random Normal Init and fixed optimizer hyperparameters

arshagarwal:
In response to the discussions in this [pull request](https://github.com/tensorflow/docs/pull/1643), this pull request  adds Random normal initializer with mean 0 and std-dev 1, and fixes the optimizer hyper-parameters as per DC-GAN paper.

I am sure there is a more graceful way to do this, but I found this way much easier than to fix code in the same pull request.
Sorry about that!


MarkDaoust:
Friendly ping @yashk2810

lamberta:
Merge conflict, please rebase

site/en/tutorials/generative/dcgan.ipynb
```diff
@@ -273,32 +246,32 @@
...
       "execution_count": null,
       "metadata": {
-        "colab": {},
-        "colab_type": "code",
         "id": "6bpTcDqoLWjY"
       },
       "outputs": [],
       "source": [
+        "# Random Normal model initializers as per Dc-gan paper\n",
+        "RN_initializers=tf.keras.initializers.RandomNormal(mean=0.0,stddev=0.02)\n\n",
```
8bitmp3, line 254: That's a great observation about the zero-centered Normal, thanks for bringing this up.

I have a few style-related suggestions, if you don't mind: 

- Change the DCGAN spelling from "Dc-gan" (source: https://arxiv.org/pdf/1511.06434.pdf) (in the code comment)
- Change to `weight initializer`, instead of `model initializers` (also in the code comment)
- Add a blank line before the generator method
- Add a space in the params section in `RandomNormal()` after a comma (I think it's PEP 8: https://www.python.org/dev/peps/pep-0008/)

```suggestion
        "# Random Normal weight initializer as per the DCGAN paper\n",
        "RN_initializers=tf.keras.initializers.RandomNormal(mean=0.0, stddev=0.02)\n\n",
        
```

Also, since it's a weight initializer (singular), maybe you should rename the variable to `RN_initializer` without the `s`.

Anyway, these are just recommendations 👍

PR #1728, opened 2020-10-25
nbfmt: Support using nbfmt as a component of shell pipeline

CourierKyn:
nbfmt can read source of a notebook from stdin and output to stdout


CourierKyn:
@googlebot I signed it!

PR #1729, opened 2020-10-25
nbfmt: Support using nbfmt as a component of shell pipeline

CourierKyn:
nbfmt now can read source of a notebook from stdin and can output formatted contents to stdout.


lamberta:
Thanks, @CourierKyn !
I'm all for bash pipelines but what is the use case for this feature? `curl` directly to it?
One issue I see is that it's skipping over the filename check which is required for us. A quick and dirty solution is just to create a temp file of possible a `mkfifo` to get around these limitations.
This change seems a bit much for this feature and is why I've been putting off a review 😅

CourierKyn:
Thanks for reply, @lamberta. I currently use this feature for git filter and diff. The problem does exist, because it can't set filename automatically according to the notebook path. I'm trying to find a solution to get the filename in this mode.

PR #1737, opened 2020-11-06
Correct the LayerNormalization axis from 1 to -1

shabie:
The paper applies it to the last dimension. The current implementation takes the default parameter value for `axis` which is 1 for LayerNormalization.

The correct value, IMHO, should be `axis=-1` for both encoder and decoder.


shabie:
@googlebot I signed it!

shabie:
I don't know what this CI error means honestly. Found "we -> you". The only "we" in the notebook is in an example of spanish translation which I never changed to begin with...

MarkDaoust:
Hi.

Thanks for taking the time to send a PR.

> The paper applies it to the last dimension

Yes.

> The current implementation takes the default parameter value for axis which is 1 for LayerNormalization.

It looks like you misread. The default is `-1`:

https://www.tensorflow.org/api_docs/python/tf/keras/layers/LayerNormalization
https://github.com/tensorflow/tensorflow/blob/v2.3.1/tensorflow/python/keras/layers/normalization.py#L1042-L1069

So there's no need for this change.

PR #1739, opened 2020-11-08
Remove redundant pandas default value parameters

albertvillanova:
Remove default value parameters:
- `index_col=None` by default
- `header=None` by default when `names` are passed


MarkDaoust (dismissed):
Normally we consider the r1/ notebooks "archived". But there's nothing wring with these changes, so we'll accept them.

PR #1741, opened 2020-11-09
Fixed formula for violinplot diagram

Blacklemming:
The formula was, which is wrong in this context and messes up the violinplot diagram :
df_std = (df - train_mean) / train_std 
It must be : 
df_std = (train_df - train_mean) / train_std


Blacklemming:
@googlebot I signed it!

Blacklemming:
My bad, seems there was some parts of the notbook not compiled :-)

PR #1742, opened 2020-11-12
Initialize encoder hidden state for each batch

wangyouzhong:
It seems more reasonable to initialize encoder hidden state for each batch during training process


lamberta:
Thanks @wangyouzhong 
Unfortunately, we can't review or merge until you sign a CLA: https://cla.developers.google.com/ 
Please re-open after you've signed. Thank you

PR #1743, opened 2020-11-13
Fix an error

ruddyscent:
The unique characters are stored at `vocab` and the last index of them is `len(vocab)-1`.

### DIFF
--- a/site/en/tutorials/text/text_generation.ipynb
+++ b/site/en/tutorials/text/text_generation.ipynb
@@ -262,7 +262,7 @@
         "id": "tZfqhkYCymwX"
       },
       "source": [
-        "Now you have an integer representation for each character. Notice that you mapped the character as indexes from 0 to `len(unique)`."
+        "Now you have an integer representation for each character. Notice that you mapped the character as indexes from 0 to `len(vocab)-1`."
       ]
     },
     {


MarkDaoust:
Sorry @ruddyscent, this tutorial moved to the tensorflow_text repository.

8bitmp3:
@ruddyscent Thanks again for the PR and the patience. 

The [_Process the text_](https://www.tensorflow.org/text/tutorials/text_generation#process_the_text) section of the _Text generation with an RNN_ tutorial was rewritten in January 2021 (commit: https://github.com/tensorflow/docs/commit/7f4d0f1780089308106b8ee3336dca24b730d84f) utilizing the new preprocessing API that maps strings from a vocab to integers. You may notice the updated tutorial has a new import: `from tensorflow.keras.layers.experimental import preprocessing`. 

The API is called `tf.keras.layers.experimental.preprocessing.StringLookup`:

> The preprocessing.StringLookup layer can convert each character into a numeric ID. It just needs the text to be split into tokens first....

and you can read more about it in the docs: https://www.tensorflow.org/api_docs/python/tf/keras/layers/experimental/preprocessing/StringLookup 

Thank you again and we look forward to your future contributions :+1:

PR #1744, opened 2020-11-13
Update data.ipynb

bibhas2:
The modified section had a defect. The only reason the defect was not visible in the output was because the label width of 5 is exactly half of the feature length of 10. In that case the index ``[:-label_length]`` will be same as ``[:label_length]``. The correct index is ``[:label_length]``.

I have changed the label length to 3 so that any problem with the code shows up.

On a related note, I see that a label length of 5 is used throughout this notebook. I do not have the time right now to see if similar defect exists elsewhere.

### DIFF
--- a/site/en/guide/data.ipynb
+++ b/site/en/guide/data.ipynb
@@ -2242,14 +2242,14 @@
       "outputs": [],
       "source": [
         "feature_length = 10\n",
-        "label_length = 5\n",
+        "label_length = 3\n",
         "\n",
         "features = range_ds.batch(feature_length, drop_remainder=True)\n",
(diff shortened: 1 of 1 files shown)


bhack:
Just to reference the issue https://github.com/tensorflow/tensorflow/issues/44779

bhack:
@bibhas2 You need to sign the [CLA](https://github.com/tensorflow/docs/pull/1744#issuecomment-726806795)

lamberta:
Thanks but we need the contributor to sign a CLA before we can review and merge. Closing this PR but please reopen after signing.

PR #1745, opened 2020-11-15
[tutorials/generative] Use passed argument in `print_stats`

mtsokol:
Hi!
`preds` used in `print_stats` refers to a variable defined in next cell so not using it will result in `print_stats` fail.


MarkDaoust (approved):

8bitmp3:
Thanks for spotting this @mtsokol 👍  A function shouldn't ignore its arguments. 

Looking forward to your future contributions. 

Feel free to join the TensorFlow community at discuss.tensorflow.org

PR #1746, opened 2020-11-17
Fix some typos in TF NumPy guide

Molkree:
Fix section links (it doesn't work when first letter is in upper case).
Measure vectorized/unvectorized code the same number of times.

### DIFF
--- a/site/en/guide/tf_numpy.ipynb
+++ b/site/en/guide/tf_numpy.ipynb
@@ -70,7 +70,7 @@
       "source": [
         "## Overview\n",
         "\n",
-        "TensorFlow implements a subset of the [NumPy API](https://numpy.org/doc/stable), available as `tf.experimental.numpy`. This allows running NumPy code, accelerated by TensorFlow, while also allowing access to all of TensorFlow's APIs."
+        "TensorFlow implements a subset of the [NumPy API](https://numpy.org/doc/1.16), available as `tf.experimental.numpy`. This allows running NumPy code, accelerated by TensorFlow, while also allowing access to all of TensorFlow's APIs."
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


site/en/guide/tf_numpy.ipynb
```diff
@@ -155,12 +155,12 @@
       "source": [
         "### Type promotion\n",
         "\n",
-        "TensorFlow NumPy APIs have well-defined semantics for converting literals to ND array, as well as for performing type promotion on ND array inputs. Please see [`np.result_type`](https://numpy.org/doc/stable/reference/generated/numpy.result_type.html) for more details. . When converting literals to ND array, NumPy prefers wide types like `tnp.int64` and `tnp.float64`.\n",
+        "TensorFlow NumPy APIs have well-defined semantics for converting literals to ND array, as well as for performing type promotion on ND array inputs. Please see [`np.result_type`](https://numpy.org/doc/1.16/reference/generated/numpy.result_type.html) for more details. When converting literals to ND array, NumPy prefers wide types like `tnp.int64` and `tnp.float64`.\n",
```
lamberta, line 158: Pretty sure we do not want to maintain these versioned links going forward. What's wrong with linking to the stable version?

Molkree: "APIs are based on and have been tested with NumPy 1.16 version." from [here](https://www.tensorflow.org/api_docs/python/tf/experimental/numpy). Sure, now there's no difference between 1.16 version and current stable 1.19 pages, but this can change. On the other hand I can understand desire to not maintain links up to date if/when TF bumps baseline.
The very first link in this guide leads to 1.16 version of docs (line 73), should be changed too then?

lamberta: Hmm. Ok, thanks for pointing it out.

@agarwal-ashish what is the plain for maintaining the tf.numpy versions? Pointing to 1.19 docs when working off 1.16  might confusion, if not errors. We can hard-link to 1.16 (like this proposed pull request), but this is difficult to maintain without some automation for tf.numpy releases.

lamberta: Or @yashk2810 , can we hook up a versioned numpy to auto-link to ref docs like the tf ref docs?

8bitmp3: Hi @Molkree @lamberta @yashk2810 AFAIK, no major changes in NumPy are planned for a while (at NumPy docs, we use `stable` in links to APIs). Therefore, I suggest to stick to the `stable` branch in the link + "APIs are based on and have been tested with NumPy 1.16 version" should be enough.

lamberta: Thanks, all.
We've decided to keep the links pointing to stable here in the source and will add a preprocessing step when publishing the site to point to the latest supported version (currently 1.16)

Molkree: Okay, so I am reverting the ones I changed, back to stable, correct? What about the one that was pointing at 1.16 even before me? :D It's a the very beginning of guide.

lamberta: Correct :)
All NumPy API links should point to "doc/stable" and we'll handle the the rest

yashk2810: There are complications with the preprocessing approach.

For now, revert back to the 1.16 link and we'll see what we can do on our end.

Molkree:
Any progress, @yashk2810?

yashk2810 (approved):

PR #1750, opened 2020-11-24
Format quotations

tocom242242:
I changed double quotations to single quotations.


lamberta:
Thanks but this seems like unnecessary churn for our docs pipeline and translations. Closing

PR #1755, opened 2020-11-27
Updated the older link to Ubuntu 16.04 (CUDA 10.1)

vishalsingha:
Updated the older link  with newer link for the installation of Ubuntu 16.04 (CUDA 10.1).


vishalsingha:
@googlebot I signed it!

lamberta (approved):
Thanks for the fix!

PR #1756, opened 2020-11-29
Small fix in RNN text generation docs

asaf92:
The stateful parameter should only be true when generating text. During training the data is shuffled, so keeping state of previous entries is wrong.


asaf92:
@googlebot I signed it!

asaf92:
This is also discussed here: https://stackoverflow.com/questions/60934038/misleading-training-data-shuffle-for-stateful-gru-in-tensorflow-text-generation/

PR #1757, opened 2020-11-29
Guide for installing Tensorflow-GPU in WSL2

Avditvs:
Added instructions enabling the use of CUDA acceleration inside Windows Subsystem for Linux 2.

### DIFF
--- a/site/en/install/gpu.md
+++ b/site/en/install/gpu.md
@@ -167,6 +167,69 @@ complicates installation of the NVIDIA driver and is beyond the scope of these i
 </code>
 </pre>
 
+## Windows Subsystem for Linux 2 setup
+The following instructions describe how to set up CUDA inside a WSL2 instance. This is still experimental and can lead to bugs or unexpected behaviours.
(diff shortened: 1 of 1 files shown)


lamberta:
Thanks, @Avditvs , this is cool.

Unfortunately, since this is not an officially supported target we can't commit to maintaining this section.

@angerson Is this something that could live under [SIG BUILD](https://github.com/tensorflow/build)?

angerson:
Yeah, this could go under Build, certainly. The other lead and are just now
rearranging the folders to fit our rearrangement proposal
<https://github.com/tensorflow/community/pull/331>, so something like this
would be a welcome new addition to test it out.

lamberta:
Sounds good, thanks @angerson 

@Avditvs Can you submit this PR to the SIG Build repo, please? https://github.com/tensorflow/build
They have a [public mailing list](https://groups.google.com/a/tensorflow.org/g/build) if you want to discuss. Closing this PR.

PR #1758, opened 2020-11-29
Define and update num_steps

lakshmanok:


MarkDaoust:
`num_steps` is set to the number of steps per epoch earlier in the notebook, and the "batch" variable in this cell already contains the epoch-step number, so I don't think this change is necessary.

Thanks.

PR #1759, opened 2020-12-03
Update gpu.md set LD_LIBRARY_PATH env var ubuntu

mdautrey:
Setting LD_LIBRARY_PATH is required so that tf-gpu works properly with ubuntu. It deserves a specific section in the doc, I think.


mdautrey:
@googlebot I signed it!

mdautrey:
A document written by NVidia clarifies how to set environment variables to make Cuda works properly : https://docs.nvidia.com/cuda/pdf/CUDA_Installation_Guide_Linux.pdf

PR #1760, opened 2020-12-05
[*.py] Rename "Arguments:" to "Args:"

SamuelMarks:
I've written custom parsers and emitters for everything from docstrings to classes and functions. However, I recently came across an issue with the TensorFlow codebase: inconsistent use of `Args:` and `Arguments:` in its docstrings. It is easy enough to extend my parsers to support both variants, however it looks like `Arguments:` is wrong anyway, as per:

  - https://google.github.io/styleguide/pyguide.html#doc-function-args @ [`ddccc0f`](https://github.com/google/styleguide/blob/ddccc0f/pyguide.md)

  - https://chromium.googlesource.com/chromiumos/docs/+/master/styleguide/python.md#describing-arguments-in-docstrings @ [`9fc0fc0`](https://chromium.googlesource.com/chromiumos/docs/+/9fc0fc0/styleguide/python.md)

  - https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html @ [`c0ae8e3`](https://github.com/sphinx-contrib/napoleon/blob/c0ae8e3/docs/source/example_google.rst)

Therefore, only `Args:` is valid. This PR replaces them throughout the codebase.

PS: For related PRs, see tensorflow/tensorflow/pull/45420

### DIFF
--- a/tools/tensorflow_docs/api_generator/parser_test.py
+++ b/tools/tensorflow_docs/api_generator/parser_test.py
@@ -610,7 +610,7 @@ class ParserTest(parameterized.TestCase):
     downgrader = parser._DowngradeH1Keywords()
     doc = downgrader(h1_docstring)
     self.assertIn('\n  ```\n  # comment\n  ```', doc)
-    self.assertIn('\nArguments:', doc)
+    self.assertIn('\nArgs:', doc)
     self.assertIn('\nExample:', doc)
     self.assertIn('\nReturns:', doc)
     self.assertIn('\nRaises:', doc)
--- a/tools/tensorflow_docs/api_generator/public_api.py
+++ b/tools/tensorflow_docs/api_generator/public_api.py
@@ -31,7 +31,7 @@ _TYPING = frozenset(
 def ignore_typing(path, parent, children):
   """Removes all children that are members of the typing module.
 
-  Arguments:
+  Args:
     path: A tuple or name parts forming the attribute-lookup path to this
       object. For `tf.keras.layers.Dense` path is:
         ("tf","keras","layers","Dense")
@@ -92,7 +92,7 @@ def local_definitions_filter(path, parent, children):
     subpackage
       from my_package._utils import *
   ```
-  Arguments:
+  Args:
     path: A tuple or name parts forming the attribute-lookup path to this
       object. For `tf.keras.layers.Dense` path is:
         ("tf","keras","layers","Dense")


tools/tensorflow_docs/api_generator/parser_test.py
```diff
@@ -610,7 +610,7 @@ def test_downgrade_h1_docstrings(self):
     downgrader = parser._DowngradeH1Keywords()
     doc = downgrader(h1_docstring)
     self.assertIn('\n  ```\n  # comment\n  ```', doc)
-    self.assertIn('\nArguments:', doc)
+    self.assertIn('\nArgs:', doc)
```
SamuelMarks, line 613: This is correct, but you'll likely need to change code elsewhere to guarantee that `Arguments` is not generated. I think it's this code though, that I've already changed: https://github.com/tensorflow/tensorflow/pull/45420/files#diff-23d5b1c21f17ffbee338aed996ac92e9e247f940a461464e1810bc67801e6769

SamuelMarks, line 613: This is correct (I think), but you'll likely need to change code elsewhere to guarantee that `Arguments` is not generated. For example, I updated this function call: https://github.com/tensorflow/tensorflow/pull/45420/files#diff-23d5b1c21f17ffbee338aed996ac92e9e247f940a461464e1810bc67801e6769

lamberta:
Thanks, @SamuelMarks 

Per the [Google Python style guide](https://google.github.io/styleguide/pyguide.html), yes, "Args:" is correct. Though I believe this tool accepts both "Args:" and "Arguments:" and I think is the correct behavior if we're following Postel's principle ("be conservative in what you do, be liberal in what you accept from others"). cc: @yashk2810

yashk2810:
Hey @SamuelMarks, This API generator supports a lot of projects other than TensorFlow. Given that both Args and Arguments is accepted convention according to google python style guide, we need to allow both of them here.

I can see that you are only changing some parts here, but those parts need to be Arguments because it acts as a test for the API generator.

SamuelMarks:
No worries. In that case I would recommend that you Googlers coordinate with the [internal] authors of the style guide, to permit both variants.

yashk2810:
I think both are permitted. AFAIK, pylint doesn;t complain if Arguments is used and if pylint doesn't complain, then its permitted :)

> wider TF Ecosystem
Yes, but this API generator is also used for projects outside of the TF ecosystem :)

SamuelMarks:
Yahh but that's just pylint (also pyflakes through this plugin - https://github.com/PyCQA/pydocstyle/blob/7921a6d/src/pydocstyle/checker.py#L74)

So if everyone else is accepting both, then maybe it's time to update the official docstring style guides from Google?

PR #1761, opened 2020-12-07
Externally loading the tensorboard

ariG23498:
In the docs present
The tensorboard is not loaded with the `%load_ext tensorboard`. In my opinion tensorboard would run in colab after the tensorboard is externally loaded in the colab notebook first.

I have managed to add the same to first load and then host tensorboard.


MarkDaoust (approved):
Thanks.

PR #1764, opened 2020-12-09
fix minor typos in tensor.ipynb

lemonez:


lemonez:
@googlebot I signed it!

site/en/guide/tensor.ipynb
```diff
@@ -148,7 +148,7 @@
         "id": "G3IJG-ug_H4u"
       },
       "source": [
-        "A \"matrix\" or \"rank-2\" tensor has 2-axes:"
```
lamberta, line 151: This should actually be "two axes". According to the Google style guide, we spell out numbers under 10: https://developers.google.com/style/numbers

And for tensors, according to the TensorFlow style guide we use something like "two-dimensional tensor with shape [...]": https://www.tensorflow.org/community/contribute/docs_style#tensors

Can you make those changes while you're in here? Thanks

lemonez: Sure, will do that. Thanks for the guidance.

lemonez:
I committed the requested changes. It is now failing lint check for 1st vs. second person, but none of my edits should have affected that.

lamberta (approved):
Thanks!

PR #1774, opened 2020-12-17
Update cnn.ipynb

ymodak:
fixes https://github.com/tensorflow/tensorflow/issues/45801


lamberta:
We don't use softmax and prefer logits. Can you keep the code the same update the text instead? Thanks

ymodak:
done

lamberta (approved):
Thanks!

PR #1781, opened 2020-12-21
Fix incorrect link to save/restore model in multi-worker training tutorial

terrytangyuan:
The existing link does not work.


lamberta (approved):
Thanks for the fix!

PR #1783, opened 2020-12-23
use Pandas timestamp instead of datetime timestamp

jcjveraa:
While the pandas dataframe seems to be in UTC (my inspector shows 'Z'  for zulu timezone, =UTC), Using datetime.datetime.timestamp seems to interpret the time as the local time. This results in different outputs depending on your local timezone (clearly visible in the ' Time of Day signal plot' ).

Using the pandas equivalent function works uniformly independent from local environment.

### DIFF
--- a/site/en/tutorials/structured_data/time_series.ipynb
+++ b/site/en/tutorials/structured_data/time_series.ipynb
@@ -398,7 +398,7 @@
       },
       "outputs": [],
       "source": [
-        "timestamp_s = date_time.map(datetime.datetime.timestamp)"
+        "timestamp_s = date_time.map(pd.Timestamp.timestamp)"
       ]
     },
     {


jcjveraa:
@googlebot I signed it!

8bitmp3 (approved):
Thank you @jcjveraa 

@wolffg @MarkDaoust PTAL

(Reference docs: https://docs.python.org/3/library/datetime.html#datetime.datetime.timestamp and https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Timestamp.html)

8bitmp3:
Thanks again @jcjveraa 👍 =

jcjveraa:
You're welcome of course :)

Op vr 11 jun. 2021 14:22 schreef 8bitmp3 ***@***.***>:

PR #1788, opened 2020-12-28
Redundant piece of code.

ariG23498:
In this tutorial the text vectorization takes place at two places. This is redundant piece of code. Would make sense to remove one of the vectorization before the final vocab change.


MarkDaoust (approved):
Thanks.

PR #1790, opened 2020-12-30
Correct call to Hyperband tuner

JoshEZiegler:
Previously omitted the 'tuners' subdirectory.  Unclear to me if this worked for a previous version, but this is how it's done in the docs and it works.

### DIFF
--- a/site/en/tutorials/keras/keras_tuner.ipynb
+++ b/site/en/tutorials/keras/keras_tuner.ipynb
@@ -237,7 +237,7 @@
       },
       "outputs": [],
       "source": [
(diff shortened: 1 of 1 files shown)


8bitmp3:
Hi @JoshEZiegler Thanks for the PR and sorry about the wait. I've checked with the team including @haifeng-jin - the recommended way for the built-in tuners is to call them as e.g. `kt.Hyperband`. If anything changes, we can reopen this PR. By the way, feel free join the TF community at https://discuss.tensorflow.org if you haven't already. Looking forward to your future contributions 👍

PR #1791, opened 2020-12-30
Update GPU support

bastiaanv:
Added instructions for ubuntu 20.04


bastiaanv:
Closing it, because there are a lot of commands not work. Need to test it on a clean computer, then I will open a new merge reques

PR #1800, opened 2021-01-15
Fix typos

ruddyscent:
1. 'output sequence as a in a single step.' -> 'output sequence in a single step.'

2. 'You could take any of single single-step multi-output models' -> 'You could take any of the single-step multi-output models'


MarkDaoust (approved):
Thanks!

PR #1801, opened 2021-01-15
Edit __new__ to not return Warning

MatthewTafazoli:
Hello,

For the "__new__" function I received the following warning:

```
WARNING:tensorflow:From <ipython-input-3-2d96a5e17d28>:17: calling DatasetV2.from_generator (from tensorflow.python.data.ops.dataset_ops) with output_types is deprecated and will be removed in a future version.
Instructions for updating:
Use output_signature instead
WARNING:tensorflow:From <ipython-input-3-2d96a5e17d28>:17: calling DatasetV2.from_generator (from tensorflow.python.data.ops.dataset_ops) with output_shapes is deprecated and will be removed in a future version.
Instructions for updating:
Use output_signature instead

```
So I went through the official documentation, and then updated the lines accordingly. The code now runs without the warning.

Best Regards,

Matthew Tafazoli

### DIFF
--- a/site/en/guide/data_performance.ipynb
+++ b/site/en/guide/data_performance.ipynb
@@ -102,7 +102,6 @@
     },
     {
       "cell_type": "code",
-      "execution_count": null,
       "execution_count": 2,
       "metadata": {
         "id": "IqR2PQG4ZaZ0"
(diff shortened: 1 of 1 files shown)


MatthewTafazoli:
@googlebot I signed it!

MarkDaoust (approved):
Thanks!

PR #1810, opened 2021-01-30
Use probabilty output for the discriminator

EvolveYourMind:
The output of the discriminator should be the probability (sigmoid activation) that the given image is real or fake, in order to be consistent with the BCE optimisation.


lamberta:
Hmm. I know we're intentionally not using sigmoid in favor of logits on tensorflow.org. cc: @yashk2810

site/en/tutorials/generative/dcgan.ipynb
```diff
@@ -332,7 +332,7 @@
         "    model.add(layers.Dropout(0.3))\n",
         "\n",
         "    model.add(layers.Flatten())\n",
-        "    model.add(layers.Dense(1))\n",
+        "    model.add(layers.Dense(1, activation='sigmoid'))\n",
```
yashk2810, line 335: This is intentional because while this can make the model output more directly interpretable, this approach is discouraged as it's impossible to provide an exact and numerically stable loss calculation for all models when using a softmax output.

See https://www.tensorflow.org/tutorials/quickstart/beginner for the full explanation

MarkDaoust: And you can't change the output activation here without also changing the loss handling later on.

EvolveYourMind: > The BCE has from_logits=True.

Oh, I see, that makes sense. sorry for the time waste

MarkDaoust: No problem. Thanks for trying to help out. 👍

PR #1812, opened 2021-02-01
Fix argument for listing GPU

yanp:
`tf.config.list_physical_devices('gpu')` does not correctly list the GPUs, changing to `GPU` fixes it. Reference: https://www.tensorflow.org/api_docs/python/tf/config/list_physical_devices


lamberta (approved):
Thanks!

PR #1814, opened 2021-02-03
Map the text dataset to the proper function.

ariG23498:
In this PR I have rectified the mapping function.
Previously the `text_ds` was mapped to the `vectorize_layer` which is a `TextVectorization` layer. Now it correctly maps to the `vectorixe_text` method that is written in the code cell itself.

### DIFF
--- a/site/en/tutorials/text/word2vec.ipynb
+++ b/site/en/tutorials/text/word2vec.ipynb
@@ -959,7 +959,7 @@
         "  return tf.squeeze(vectorize_layer(text))\n",
         "\n",
         "# Vectorize the data in text_ds.\n",
-        "text_vector_ds = text_ds.batch(1024).prefetch(AUTOTUNE).map(vectorize_layer).unbatch()"
+        "text_vector_ds = text_ds.batch(1024).prefetch(AUTOTUNE).map(vectorize_text).unbatch()"
       ]
     },
     {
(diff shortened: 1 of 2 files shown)


ariG23498:
I have also tried the code with the previous setting and the output is the same.
I am thinking whether it would be better to **remove** the function :point_down: 
```python
def vectorize_text(text):
  text = tf.expand_dims(text, -1)
  return tf.squeeze(vectorize_layer(text))
```

And keep the previous mapping to the layer.
```python
# Vectorize the data in text_ds.
text_vector_ds = text_ds.batch(1024).prefetch(AUTOTUNE).map(vectorize_layer).unbatch()
```
Would that make sense?

site/en/tutorials/text/word2vec.ipynb
```diff
@@ -959,7 +959,7 @@
         "  return tf.squeeze(vectorize_layer(text))\n",
```
MarkDaoust, line 959: I think `vectorize_text` is just left over from an earlier version, before the layer could handle this particular input. 

So I'm +1 on just deleting the function and using the layer.

ariG23498: Done! Thanks :smile:

MarkDaoust (approved):

PR #1816, opened 2021-02-11
Switched parameters in loss function in file site/en/guide/basic_training_loops.ipynb

smorrin28:
In the loss visualization step, the parameters for the loss function were in the wrong order (while this doesn't affect the outcome, it may be confusing).


MarkDaoust (approved):
Thanks!

PR #1817, opened 2021-02-12
Validation set is not used

goungoun:
As pointed out in tensorflow/tensorflow#28371, validation set is not used. I removed unused variables `input_tensor_val`, `target_tensor_val` and its package import.


goungoun:
> @googlebot I signed it!

MarkDaoust:
Hi, thanks for taking the time to make the change.

Even if this tutorial doesn't use the validation set, it's still better to split it off before trying any training.

I think if there's an update to make here it should be in the other direction, we could add a cell or two to evaluate the the model on the validation set.

goungoun:
@MarkDaoust, It's been two years since it is raised by the community, but still not implemented or annotated. Is there a certain reason to hesitate?

PR #1820, opened 2021-02-17
Note lack of support for macOS 11 (Big Sur)

jameshfisher:
See https://github.com/tensorflow/tensorflow/issues/45120 and https://github.com/tensorflow/tensorflow/issues/47205 - installation, at least with the standard method, does not work on Big Sur.

Better to note this lack of support explicitly rather than make each macOS 11 user debug the installation.

### DIFF
--- a/site/en/install/_index.yaml
+++ b/site/en/install/_index.yaml
@@ -26,7 +26,7 @@ landing_page:
             </ul>
           </td><td>
             <ul>
-              <li>macOS &gt;10.12.6 (Sierra) but &lt;11 (Big Sur) (no GPU support)</li>
+              <li>macOS &gt;=10.12.6 (Sierra) but &lt;11 (Big Sur) (no GPU support)</li>
               <li>Raspbian 9.0 or later</li>
             </ul>
           </td></tr>
--- a/site/en/install/_index.yaml
+++ b/site/en/install/_index.yaml
@@ -26,7 +26,10 @@ landing_page:
             </ul>
           </td><td>
             <ul>
-              <li>macOS &gt;=10.12.6 (Sierra) but &lt;11 (Big Sur) (no GPU support)</li>
+              <li>
+                macOS &gt;=10.12.6 (Sierra) but &lt;11 (Big Sur) (no GPU support).
+                For Big Sur and GPU support, use <a href="https://github.com/apple/tensorflow_macos">tensorflow_macos</a>.
+              </li>
               <li>Raspbian 9.0 or later</li>
             </ul>
           </td></tr>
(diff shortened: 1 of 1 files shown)


site/en/install/_index.yaml
```diff
@@ -26,7 +26,10 @@ landing_page:
             </ul>
           </td><td>
             <ul>
-              <li>macOS 10.12.6 (Sierra) or later (no GPU support)</li>
+              <li>
+                macOS &gt;=10.12.6 (Sierra) but &lt;11 (Big Sur) (no GPU support).
+                For Big Sur and GPU support, use <a href="https://github.com/apple/tensorflow_macos">tensorflow_macos</a>.
```
mihaimaruseac, line 31: Isn't that one only for TF on the M1 chip?

lamberta: Readme says "hardware acceleration is supported on Macs with M1 and Intel-based Macs through Apple’s ML Compute framework."
Though the TF version lags behind a bit.

jameshfisher: I'm on an x86 Mac running macOS 11. tensorflow_macos installs and runs OK for me (so far). Mainline Tensorflow does not.

8bitmp3: Hi @jameshfisher, macOS 11.1 here without M1  (11.3 beta is the latest, I think). Installing TF2 with PiPy worked as recently as January.

jameshfisher: @8bitmp3 interesting - how exactly did you install it? The standard `pip install tensorflow` worked for you?

8bitmp3: @jameshfisher Yep. Have you tried setting up a new tf2 virtualenv/docker for a "clean-install"?

jameshfisher: @8bitmp3 interesting - I tried it with virtualenv, and it installs! Thanks. However, it's very unclear why virtualenv makes a difference, and the TensorFlow install instructions don't say to use virtualenv.

So the mystery deepens on https://github.com/tensorflow/tensorflow/issues/47205. 

Docker I would consider cheating - it's basically installing on Linux, not macOS.

@8bitmp3 which method are you using on macOS? virtualenv or docker?

8bitmp3: I'm glad this helped @jameshfisher. I use `virtualenv` or `pyenv` because on macOS you can end up with too many Python interpreters. Check out this post: https://opensource.com/article/19/5/python-3-default-mac (note that homebrew tends to installs the latest one even if you don't need it).

jameshfisher:
Closing in favor of https://github.com/tensorflow/docs/pull/1826

mihaimaruseac:
Closing for #1826

PR #1826, opened 2021-02-19
Note requirement for pip 20.3+ on macOS 11

jameshfisher:
With pip <20.3 (which is expected on macOS, because `python@3.8` only includes pip 20.2.4), installation fails with "No matching distribution found for tensorflow".

Underlying cause is the lack of TensorFlow packages tagged as `macosx_11`: https://github.com/tensorflow/tensorflow/issues/45120

pip 20.3+ changes behavior to allow `macosx_10` packages on macOS 11: https://github.com/pypa/pip/issues/9138

### DIFF
--- a/site/en/install/_index.yaml
+++ b/site/en/install/_index.yaml
@@ -39,7 +39,7 @@ landing_page:
     - heading: Download a package
       description: >
         <p>Install TensorFlow with Python's <em>pip</em> package manager.</p>
-        <aside class="note">TensorFlow 2 packages require a <code>pip</code> version >19.0.</aside>
+        <aside class="note">TensorFlow 2 packages require a <code>pip</code> version version 20.3+ when running on macOS 11, or >19.0 on all other systems.</aside>
         <p>Official packages available for Ubuntu, Windows, macOS, and the Raspberry Pi.</p>
         <p>See the <a href="./gpu">GPU guide</a> for CUDA®-enabled cards.</p>
 


mihaimaruseac (approved):

site/en/install/_index.yaml
```diff
@@ -39,7 +39,7 @@ landing_page:
     - heading: Download a package
       description: >
         <p>Install TensorFlow with Python's <em>pip</em> package manager.</p>
-        <aside class="note">TensorFlow 2 packages require a <code>pip</code> version >19.0.</aside>
+        <aside class="note">TensorFlow 2 packages require a <code>pip</code> version 20.3+ when running on macOS 11, or >19.0 on all other systems.</aside>
```
mihaimaruseac, line 42:
```suggestion
        <aside class="note">TensorFlow 2 packages require a <code>pip</code> version >20.3 when running on macOS 11, or >19.0 on all other systems.</aside>
```

Changing for consistency

mihaimaruseac (approved):

mihaimaruseac (dismissed):

site/en/install/_index.yaml
```diff
@@ -39,7 +39,7 @@ landing_page:
     - heading: Download a package
       description: >
         <p>Install TensorFlow with Python's <em>pip</em> package manager.</p>
-        <aside class="note">TensorFlow 2 packages require a <code>pip</code> version >19.0.</aside>
+        <aside class="note">TensorFlow 2 packages require a <code>pip</code> version >20.3 when running on macOS 11, or >19.0 on all other systems.</aside>
```
jameshfisher, line 42: Technically it should be >=20.3, as 20.3 is sufficient

(I wonder whether this is also true for the ">19.0")

lamberta: Yeah, this is indicated on [pip-specific page](https://github.com/tensorflow/docs/blob/master/site/en/install/pip.html#L34) and should follow here. Ideally, that's the page with the more exhaustive requirements and this provides the "high level" for folks that don't click through

lamberta (approved):

PR #1829, opened 2021-02-24
Stop doubling up on ".tar.gz" extension

jameshfisher:
Providing filename with extension results in a file on disk called "aclImdb_v1.tar.gz.tar.gz"

`get_file` appears to want a filename without an extension (though the documentation for it is unclear)


8bitmp3:
@jameshfisher yup, can confirm `aclImdb_v1.tar.gz.tar.gz` is what you get. Thanks for the fix:

```diff
dataset1 = tf.keras.utils.get_file(
-                                     "aclImdb_v1.tar.gz", url,
+                                     "aclImdb_v1", url,
                                    untar=True, cache_dir='.',
                                    cache_subdir='')

MarkDaoust (approved):

PR #1830, opened 2021-02-24
Don't import tensorflow_datasets - it's not used (or even installed)

jameshfisher:
Following this tutorial locally causes: ModuleNotFoundError: No module named 'tensorflow_datasets'

If we want to keep it then we need: pip install tensorflow-datasets

But I think we don't want to keep it because the module is not even referenced in the rest of the tutorial


jameshfisher:
My mistake - the problem is a lack of install. Fix should be to install.

jameshfisher:
(Closed in favor of https://github.com/tensorflow/docs/pull/1831)

PR #1831, opened 2021-02-24
pip install tensorflow-datasets - required for tensorflow_datasets module used in tutorial

jameshfisher:
Following this tutorial locally causes: `ModuleNotFoundError: No module named 'tensorflow_datasets'`

### DIFF
--- a/site/en/tutorials/keras/text_classification_with_hub.ipynb
+++ b/site/en/tutorials/keras/text_classification_with_hub.ipynb
@@ -107,7 +107,7 @@
         "\n",
         "The tutorial demonstrates the basic application of transfer learning with [TensorFlow Hub](https://tfhub.dev) and Keras.\n",
         "\n",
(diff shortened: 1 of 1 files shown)


8bitmp3:
@jameshfisher I think the reason why there's no `pip install tensorflow_datasets` is because it's preloaded into your regular Colab environment (`Requirement already satisfied:...`). Having said that, `tensorflow_hub` is also preinstalled, so it's a good idea to "pip install" both in case someone clones the notebook and runs it elsewhere.

jameshfisher:
@8bitmp3 agree - I think these tutorials should translate painlessly to a non-notebook environment. That's how I'm currently using them - copying pieces into local Python files and/or local repl.

lamberta (approved):
Thanks!
Yes, agree. Many times the author works directly in Colab so it's an easy step to miss.

PR #1832, opened 2021-02-24
Clarify get_dummies behavior

jameshfisher:
I was mystified by this `get_dummies` call:

1. "Dummies" means nothing to me; I didn't know it was another term for one-hot. I hinted at this by mentioning the function in the text.
2. How on earth does `get_dummies` know that `Origin` is the categorical property? Why not `Cylinders` also? It turns out the current code kinda "works by accident" - it's because the dtype of the Origin column is "object". If it was "string", it wouldn't work. I fixed this by being explicit about which column it should convert to one-hot.

### DIFF
--- a/site/en/tutorials/keras/regression.ipynb
+++ b/site/en/tutorials/keras/regression.ipynb
@@ -252,7 +252,7 @@
         "id": "8XKitwaH4v8h"
       },
       "source": [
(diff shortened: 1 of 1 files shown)


MarkDaoust (dismissed):
Thanks.

lamberta (approved):
Rebased with master to fix merge conflict

PR #1833, opened 2021-02-26
Don't guess at default batch size

jameshfisher:
I was confused about this isolated `batch_size` variable - turns out it's (ab)using the fact that 32 is the current default batch size if unspecified

But if we need to know the batch size, we should specify it

### DIFF
--- a/site/en/tutorials/keras/save_and_load.ipynb
+++ b/site/en/tutorials/keras/save_and_load.ipynb
@@ -402,6 +402,7 @@
         "model.fit(train_images, \n",
         "          train_labels,\n",
         "          epochs=50, \n",
+        "          batch_size=batch_size, \n",
         "          callbacks=[cp_callback],\n",
         "          validation_data=(test_images, test_labels),\n",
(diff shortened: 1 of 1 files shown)


8bitmp3:
Good idea @jameshfisher. And if the `batch_size` argument is not explicitly defined in `fit()`, the users should be reminded that it's "taken care of" and set to 32 by default. "I like this ~~stock~~ PR." 👍 

P.S. I personally wouldn't call the `batch_size` variable "isolated" as it appears to be created specifically to serve the model saving arg—`save_freq` every 5 x 32 (`batch_size`) times—in the custom `ModelCheckpoint` callback 🤷.  But you already know that 😄.

```py
batch_size = 32
...
cp_callback = tf.keras.callbacks.ModelCheckpoint(
    filepath=checkpoint_path,
    verbose=1,
    save_weights_only=True,
    save_freq=5*batch_size)
```

MarkDaoust (approved):
Yes. Thanks.

PR #1838, opened 2021-03-02
Update transfer_learning_with_hub.ipynb

shivaditya-meduri:
This model expects pixel values in [-1,1], more info on https://www.tensorflow.org/tutorials/images/transfer_learning

### DIFF
--- a/site/en/tutorials/images/transfer_learning_with_hub.ipynb
+++ b/site/en/tutorials/images/transfer_learning_with_hub.ipynb
@@ -372,7 +372,7 @@
         "id": "L0Btd0V3C8h4"
       },
       "source": [
-        "TensorFlow Hub's conventions for image models is to expect float inputs in the `[0, 1]` range. Use the `Rescaling` layer to achieve this."
+        "TensorFlow Hub's conventions for image models is to expect float inputs in the `[-1, 1]` range. Use the `Rescaling` layer to achieve this."
       ]
     },
     {


8bitmp3:
That's an interesting point. Technically, both scales—`[0, 1]` and `[-1, 1]`—can work, depending on which `offset` argument value you use in the `Rescale` layer. 

The API docs (https://www.tensorflow.org/api_docs/python/tf/keras/layers/experimental/preprocessing/Rescaling) have the following example:

> To rescale an input in the `[0, 255]` range to be in the `[0, 1]` range, you would pass `scale=1./255`.
> To rescale an input in the `[0, 255]` range to be in the `[-1, 1]` range, you would pass `scale=1./127.5, offset=-1`.

In the [Transfer learning with TensorFlow Hub](https://www.tensorflow.org/tutorials/images/transfer_learning_with_hub) tutorial that you want to modify, the float inputs (chosen by the author) are in the `[0, 1]` range, so the `Rescaling` API is used as follows:

```py
normalization_layer = tf.keras.layers.experimental.preprocessing.Rescaling(1./255)
```

where `scale=1./255` and `offset=0.0` (default value, as per the [API docs](https://www.tensorflow.org/api_docs/python/tf/keras/layers/experimental/preprocessing/Rescaling)).

But if the example used the `[-1, 1]` range, our line of code would have probably been written like this:

```py
normalization_layer = tf.keras.layers.experimental.preprocessing.Rescaling(1./127.5, offset= -1)
```

I also noticed that in the [Working with preprocessing layers](https://www.tensorflow.org/guide/keras/preprocessing_layers) guide it says:

> Your inference model will be able to process raw images or raw structured data, and will not require users of the model to be aware of the details of e.g. the tokenization scheme used for text, the indexing scheme used for categorical features, whether image pixel values are normalized to `[-1, +1]` or to `[0, 1]`, etc. 

Maybe I'm wrong here, looping in @yashk2810 @lamberta PTAL 👍

lamberta:
Thanks. @MarkDaoust ptal

MarkDaoust:
Thanks for flagging.

@shivaditya-meduri: Yes these are both Mobilenet V2.  But as @8bitmp3 points out, when you export a model like this you can put whatever rescaling layer you want on the input. 

As your link points out, the implementation of MovileNetV2 in tf.keras.applications uses [-1,1]

But fop consistancy with itself tensorflow_hub's rule is that [all image models take `[0,1]` as input](https://www.tensorflow.org/hub/common_saved_model_apis/images#image_input). 

It's too bad they don't match, but it's too late for us to fix now. 

> will not require users of the model to be aware of the details of e.g. the tokenization scheme used for text, the indexing scheme used for categorical features, whether image pixel values are normalized to [-1, +1] or to [0, 1], etc.

Yes. If [all the models took [0, 255] as input](https://xkcd.com/927/) then we wouldn't need to worry about it.

So, If you export a model be sure to document what you expect as input. If you use an off the shelf model, check what input it expects.

PR #1840, opened 2021-03-02
Define loss_object before usage in generator_loss function

jameshfisher:
I was confused by this reference to an as-yet-undefined variable. It's not an error as long as the function is not called, but it's bad style


MarkDaoust (approved):
Good idea, thanks.

PR #1841, opened 2021-03-03
Fix link bug from 'basic_training_loops.ipynb' to 'distributed_training.ipynb'

crutcher:
There's a broken link on `basic_training_loops.ipynb` to `distributed_training.ipynb`

https://www.tensorflow.org/guide/basic_training_loops#next_steps


lamberta (approved):
Thanks for the fix!

PR #1842, opened 2021-03-03
Update gpu.md

amahendrakar:
Removed redundant commands to install NVIDIA drivers

The `sudo apt-get install --no-install-recommends nvidia-driver-450` command seems redundant as latest drivers are installed and overwritten in the next step. Please check the below screenshot for reference. 

- Above command installs NVIDIA driver version 450

- After installing development and runtime libraries the driver version is updated to 460 

Relevant to issue [#47405](https://github.com/tensorflow/tensorflow/issues/47405). Change was previously merged in PR [#1749](https://github.com/tensorflow/docs/pull/1749/).

### DIFF
--- a/site/en/install/gpu.md
+++ b/site/en/install/gpu.md
@@ -109,10 +109,6 @@ wget http://developer.download.nvidia.com/compute/machine-learning/repos/ubuntu1
 <code class="devsite-terminal">sudo apt install ./nvidia-machine-learning-repo-ubuntu1804_1.0.0-1_amd64.deb</code>
 <code class="devsite-terminal">sudo apt-get update</code>
 
-# Install NVIDIA driver
-<code class="devsite-terminal">sudo apt-get install --no-install-recommends nvidia-driver-450</code>
-# Reboot. Check that GPUs are visible using the command: nvidia-smi
-
 <code class="devsite-terminal">wget https://developer.download.nvidia.com/compute/machine-learning/repos/ubuntu1804/x86_64/libnvinfer7_7.1.3-1+cuda11.0_amd64.deb</code>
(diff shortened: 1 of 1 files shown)


site/en/install/gpu.md
```diff
@@ -109,10 +109,6 @@ wget http://developer.download.nvidia.com/compute/machine-learning/repos/ubuntu1
 <code class="devsite-terminal">sudo apt install ./nvidia-machine-learning-repo-ubuntu1804_1.0.0-1_amd64.deb</code>
 <code class="devsite-terminal">sudo apt-get update</code>
 
-# Install NVIDIA driver
-<code class="devsite-terminal">sudo apt-get install --no-install-recommends nvidia-driver-450</code>
-# Reboot. Check that GPUs are visible using the command: nvidia-smi
```
sanjoy, line 114: Should we still keep this line somehow, about verifying that the GPUs are visible?

sanjoy (changes requested):

site/en/install/gpu.md
8bitmp3: @sanjoy Agreed. Running `nvidia-smi` is a good practice 👍 

Suggestion:

```
# Check that GPUs are visible using the command:
<code class="devsite-terminal">nvidia-smi</code>
```

(This way the user would be more inclined to run the step instead of having an option to run it IMO)

cc @lamberta

amahendrakar: @sanjoy,
Done. Added the line to check GPU visibility, after the drivers are installed. 

Thanks!

sanjoy (approved):

lamberta (approved):

PR #1845, opened 2021-03-11
Remove unused "x = inputs" assignment

jameshfisher:
Removing this assignment results in an identical model and less confusing code

### DIFF
--- a/site/en/tutorials/images/segmentation.ipynb
+++ b/site/en/tutorials/images/segmentation.ipynb
@@ -394,10 +394,9 @@
       "source": [
         "def unet_model(output_channels):\n",
         "  inputs = tf.keras.layers.Input(shape=[128, 128, 3])\n",
-        "  x = inputs\n",
         "\n",
         "  # Downsampling through the model\n",
-        "  skips = down_stack(x)\n",
+        "  skips = down_stack(inputs)\n",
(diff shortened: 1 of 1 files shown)


8bitmp3:
I think it serves a purpose from the UX point of view. You can follow `x` and see what happens to the inputs as they get propagated through the model: 1) they get downsampled; 2) then then get upsampled with skips; 3) finally, they get pushed through last layer. So, maybe we should leave this, but that's just my opinion. cc @lamberta @MarkDaoust 

```py
def unet_model(output_channels):
  inputs = tf.keras.layers.Input(shape=[128, 128, 3])
  x = inputs

  # Downsampling through the model
  skips = down_stack(x)
  x = skips[-1]
  skips = reversed(skips[:-1])

  # Upsampling and establishing the skip connections
  for up, skip in zip(up_stack, skips):
    x = up(x)
    concat = tf.keras.layers.Concatenate()
    x = concat([x, skip])

  # This is the last layer of the model
  last = tf.keras.layers.Conv2DTranspose(
      output_channels, 3, strides=2,
      padding='same')  #64x64 -> 128x128

  x = last(x)

  return tf.keras.Model(inputs=inputs, outputs=x)
```

yashk2810 (approved):

PR #1846, opened 2021-03-11
set minimum Python version to 3.6 (since TF 2.4)

zenogantner:
This was not yet reflected on https://www.tensorflow.org/install and https://www.tensorflow.org/install/pip
This PR removes the mentions of Python 3.5 on these pages.


mihaimaruseac (approved):

lamberta (approved):
Thanks!

PR #1850, opened 2021-03-15
Fix plot_spectrogram() x axis size

cashlo:
use np.linspace to get X axis so it works on other audio length. 
np.arange(16000, step=height + 1) would give a wrong size when used on audio of different size


cashlo:
@googlebot I signed it!

8bitmp3:
@MarkDaoust @cashlo @lamberta I've worked with audio and music data (mostly using `librosa`) and here are my suggestions:

1) Instead of writing a custom `plot_spectrogram()` function, we could use matplotlib's  [`matplotlib.pyplot.specgram`](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.specgram.html).

2) The second plot is in the time-frequency domain where time is represented by the x-axis. Therefore, it probably shouldn't have a length of `16,000` (which is the sample rate or samples per unit (16Hz)). The sample waveform (`label`: `stop`) is less than 1 second long, as shown in one of the cells that runs `display.display(display.Audio(waveform, rate=16000))`. 

Here's what I propose:

```py
SAMPLE_RATE = 16000

fig, (ax1, ax2) = plt.subplots(2, figsize=(12, 8))
timescale = np.arange(waveform.shape[0])
ax1.plot(timescale, waveform.numpy())
ax1.set_title('Waveform')
ax1.set_xlim(0, SAMPLE_RATE)
ax2.specgram(waveform.numpy(), Fs=SAMPLE_RATE)
ax2.set_title('Spectrogram')
plt.show()
```

image

You can double check this using the librosa library which is preloaded in Colab:

```py
import librosa
import librosa.display

temp = librosa.stft(librosa.to_mono(waveform.numpy()))
librosa.display.specshow(librosa.amplitude_to_db(np.abs(temp), ref=np.max), y_axis='log', x_axis='time')
```

image

8bitmp3:
In addition, there may be some "redundant" code @MarkDaoust @cashlo @lamberta :

```diff
for waveform, label in waveform_ds.take(1):
  label = label.numpy().decode('utf-8')
  spectrogram = get_spectrogram(waveform)

print('Label:', label)
print('Waveform shape:', waveform.shape)
print('Spectrogram shape:', spectrogram.shape)
print('Audio playback')
- display.display(display.Audio(waveform, rate=16000))
+ display.Audio(waveform, rate=16000)
```

I've reset the notebook, ran the cells again, and `display.Audio(waveform, rate=16000)` plays the audio:

image

MarkDaoust:
Thanks @cashlo and @8bitmp3.

> `plot_spectrogram()` function, we could use matplotlib's `matplotlib.pyplot.specgram`.

The reason to prefer a custom plot function here is that we wrote a `get_spectrogram` function, and we want to view the input/output. So matplotlib's spectrogram is nice, but it creates the spectrogram internally.  

> librosa

Cool, either way, I'm not concerned about which plot tool is used. I just want:

1.  The input/output of the get_spectrogram function to be visible
2. The x-axis tick-labels to match time on both or samples on both.

> `- display.display(display.Audio(waveform, rate=16000))`
> `+ display.Audio(waveform, rate=16000)`

Right. You don't need the `display.display(` if it's the last line of a code cell.

@8bitmp3, do you want to merge this and send the other fixes in another PR?

8bitmp3 (approved):
@MarkDaoust @cashlo 👍

PR #1852, opened 2021-03-17
Make plot() customizable to show more data

martinwicke:
If you increase max_subplots today, there's an out of range error because 3 is hardcoded in the plt.subplot call.


MarkDaoust (approved):
Thanks Martin!

PR #1853, opened 2021-03-18
Change initializers to object instances

martinwicke:
If callables are passed as initializers, the resulting model cannot be saved. Minimal repro:

```python
M = tf.keras.Sequential([tf.keras.layers.Dense(1, kernel_initializer=tf.initializers.zeros)])
M.build([2,1])
M.save('test_model')
```
[...blah...]
  File "/usr/local/lib/python3.8/site-packages/tensorflow/python/keras/initializers/__init__.py", line 135, in serialize
    return generic_utils.serialize_keras_object(initializer)
  File "/usr/local/lib/python3.8/site-packages/tensorflow/python/keras/utils/generic_utils.py", line 245, in serialize_keras_object
    config = instance.get_config()
TypeError: get_config() missing 1 required positional argument: 'self'

I'm guessing the initializer still works if passed as a callable. Nevertheless, passing a callable should probably be an error, *or* saving it should result in a human-readable error.

Independent of the real fix, the tutorials should not include such questionable use.


yashk2810 (approved):

yashk2810:
Thanks Martin!

martinwicke:
@fchollet since I'm not sure who's in charge of saving from Keras now. Should I bother to file a bug? It's not severe since the fix is trivial, and I suppose most people can infer roughly what's happening from the stack trace.

MarkDaoust:
> Should I bother to file a bug?

I just tried your code to reproduce this  with `tf-nightly`. It works fine there, so I think the bug is fixed.

martinwicke:
Oh, great! Then this PR also probably wasn't all that necessary. Sorry for not working at head.

PR #1855, opened 2021-03-25
Update Save and Load models to be platform independent

Rishit-dagli:
I was referring back to the [Save and load models](https://www.tensorflow.org/tutorials/keras/save_and_load) tutorials running this notebook on my personal Windows machine and I noticed it using POSIX commands which give an error on Windows machines. An easy fix can be made to solve this using `os` commands instead.

I also found an associated issue with it https://github.com/tensorflow/tensorflow/issues/33923 , which should be closed by this PR.

cc: @MarkDaoust


MarkDaoust (commented):
Thanks.

PR #1859, opened 2021-03-29
[transformer.ipynb] Modify member variable: self.tokenizer -> self.encoder.

dansuh17:
`self.encoder` seems to be more intuitive since it is an instance of `Encoder`.
It also is more symmetric with the other member variable `self.decoder`.


MarkDaoust (approved):
This looks right to me.

PR #1864, opened 2021-03-30
Remove roadmap broken link

bhack:
Fixes https://github.com/tensorflow/community/issues/206

We could reintroduce something if/when we will find an operative plan for https://github.com/tensorflow/community/issues/29


lamberta (approved):
Thanks!

PR #1867, opened 2021-04-02
Update create_op.md

galeone:
Hi!

I analyzed several things related to the custom ops in TensorFlow and I found some lack of detail in the documentation and some problem. The complete report is in this article: https://pgaleone.eu/tensorflow/bazel/abi/c++/2021/04/01/tensorflow-custom-ops-bazel-abi-compatibility/

In short, these are the changes: 

- Use TensorFlow 2.0 style (eager execution) in python example - remove `tf.Session`
- Fix the GPU Kernel code, including missing headers, adding the operation registration, and fixing a wrong include ("example.h" doesn't exist)
- Add a section on how to compile custom ops with CUDA kernels.

### DIFF
--- a/site/en/guide/create_op.md
+++ b/site/en/guide/create_op.md
@@ -188,6 +188,8 @@ Here is an example implementation.
 #ifndef KERNEL_EXAMPLE_H_
 #define KERNEL_EXAMPLE_H_
 
+#include <unsupported/Eigen/CXX11/Tensor>
+
 template <typename Device, typename T>
 struct ExampleFunctor {
   void operator()(const Device& d, int size, const T* in, T* out);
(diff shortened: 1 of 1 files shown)


bhack:
I think that we need to check this with info maintained at https://github.com/tensorflow/custom-op/blob/master/README.md

galeone:
> I think that we need to check this with info maintained at https://github.com/tensorflow/custom-op/blob/master/README.md

These are totally different things IFAIK. When I fixed in the example, is the proper way of building a custom op directly from the TensorFlow source code, while the custom-op repo relies upon the docker image (and a more complex structure).

bhack:
Yes I know but it was just an internal reminder that we are maintaing two Doc sources about custom ops.

lamberta:
Thanks, @galeone ! Blog post looks nice. Let's see if we can find someone to review this.
Last I heard, the [tensorflow/custom-op](https://github.com/tensorflow/custom-op) repo is the most up-to-date instructions. And agree with @bhack that it would be nice to consolidate these—probably over there but can link to it from the site.

bhack:
@lamberta We have already a link here in the head. But we could always diverge with multiple sources.

> Note: To guarantee that your C++ custom ops are ABI compatible with TensorFlow's
official pip packages, please follow the guide at
[Custom op repository](https://github.com/tensorflow/custom-op). It has an
end-to-end code example, as well as Docker images for building and distributing
your custom ops.

mihaimaruseac (approved):
Thank you so much for this.

galeone:
> Thank you so much for this.

It's a pleasure :smile:

PR #1868, opened 2021-04-02
Add a reference to our Governance

bhack:
We could move or refactoring that later

### DIFF
--- a/site/en/community/contribute/community.md
+++ b/site/en/community/contribute/community.md
@@ -2,6 +2,8 @@
 
 An open source project isn't just about the code, it's also about the community of users, developers, writers, researchers, and other contributors. You can help grow and support this community.
 
+Please read the TensorFlow [Code and Collaboration governance](https://github.com/tensorflow/community/blob/master/governance/code-and-collaboration.md)
+
 ## Community support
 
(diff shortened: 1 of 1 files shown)


ewilderj:
If we're touching this page, definitely needs an update. the SIG list is badly outdated :(

bhack:
Yes we know:
-  We are and we could go out of sync with mailing-lists.md
- SIGS list is outdated vs SIGS folders in community
- No Gitter links

But in the meantime I think that it is better to have a connection to our Contribution Governance with this PR so that it will easier to review what we want to do with this page as whole later.

theadactyl:
I think we want to move the file this refers to to the website, though?

theadactyl:
Agreed that this page also needs a refresh (or removal). I will assign to
myself to propose a direction when I'm back from vacation.

lamberta (approved):
Sounds like there's future work on the way but this is worth merging now. Thanks

theadactyl:
@bhack can you assign me a bug for updating the content?

bhack:
>  @bhack can you assign me a bug for updating the content?

I don't have this "power" on the repository. I can mention you but probably it is not the same for tracking.

theadactyl:
That works -- I'll assign to myself

PR #1870, opened 2021-04-05
`actor_critic.ipynb`: Remove unused imports.

nalepae:
Tested OK in `colab` without these imports.


nalepae:
The lint error seems not to be related to my PR.

lamberta:
Thanks.

> The lint error seems not to be related to my PR.

Yes, that's correct. But since you're in here can you fix it, please? :)

nalepae:
Done. Replacing `we` by `you` was not very "logical" in this context.
I tried to used various formulations.
Hope it will fit.

lamberta (approved):
Awesome. Thank you—appreciate it!

nalepae:
By the way, unless I'm mistaken, following parts are false:

> We keep a running record of episode rewards using a queue. Once 100 trials are reached, the oldest reward is removed at the left (tail) end of the queue and the newest one is added at the head (right). A running sum of the rewards is also maintained for computational efficiency.

> Cartpole-v0 is considered solved if average reward is >= 195 over 100 
consecutive trials

Since you're not averaging on the last 100 episodes, but you use a exponential running reward:

```Python
running_reward = episode_reward*0.01 + running_reward*.99
```

(I don't see any queue in the code).

I guess to be really compliant with the [CartPole-v0 definition](https://gym.openai.com/envs/CartPole-v0/):

> CartPole-v0 defines "solving" as getting average reward of 195.0 over 100 consecutive trials.
This environment corresponds to the version of the cart-pole problem described by Barto, Sutton, and Anderson [Barto83].

we really should use a queue.

I can modify the code to fit to the documentation (in a new PR), if you agree with that.

lamberta:
> we really should use a queue.

I'm ok with that if it doesn't increase tutorial complexity since we're more focused on the TF learning path. wdyt @MarkDaoust ?

MarkDaoust:
Using [collections.deque](https://docs.python.org/3/library/collections.html#collections.deque) in the main loop? 

That should only be a couple of lines. I'd approve that. 

Don't use `tf.queue`, they don't give you a good way access the whole list (for the average), and are less "pythonic".

nalepae:
> Using collections.deque in the main loop?

Yes it was exactly what I was thinking about.
Let me create a new commit in this PR for that.

nalepae:
Done.

MarkDaoust (approved):
THanks!

PR #1886, opened 2021-04-30
Update cyclegan.ipynb

ymodak:
Fixes usage of `BATCH_SIZE` in https://github.com/tensorflow/tensorflow/issues/48826


lamberta (approved):
Thanks for the fix!

PR #1778, opened 2020-12-18
Lint TensorFlow: Mixed Precision guide

8bitmp3:
Splitting PR #1736 into smaller digestible bits. Cheers @lamberta @MarkDaoust 

There's a section on a technique for mixed precision training called "loss scaling" (I also read about it here https://arxiv.org/abs/1910.12385). It looks like we can say "a loss scale value" or "a loss scale hyperparameter". Here's my new suggestion:

```diff
The basic concept of loss scaling is simple:
- simply multiply the loss by some large number, say $1024$, 
+ multiply the loss by some large number, say $1024$, 
- We call this number the *loss scale*...
- You can call this number the *loss scale*.
+ and you get the *loss scale* value.
...
```

(I've removed "simply" to avoid repetition/duplication.)

Thank you 🙏

PR #1894, opened 2021-05-20
Minor revision of the TF Mixed Precision guide

8bitmp3:
The same version of an old PR https://github.com/tensorflow/docs/pull/1778 (linting). Fresh fork.

PTAL @MarkDaoust @lamberta 

(As before, _"there's a section on a technique for mixed precision training called "loss scaling" (I also read about it here https://arxiv.org/abs/1910.12385). It looks like we can say "a loss scale value" or "a loss scale hyperparameter""_...)


CODEOWNERS
```diff
@@ -2,10 +2,10 @@
 # Last matching pattern takes preecedence.
 
 # Default owners for everything in repo.
-*  @lamberta @MarkDaoust @yashk2810 @anirudh161
+*  @lamberta @MarkDaoust @8bitmp3
```
lamberta, line 5: Did you rebase? This update is already checking in master: https://github.com/tensorflow/docs/blob/master/CODEOWNERS

lamberta (dismissed):

CODEOWNERS
lamberta: I rebased this PR with master and force pushed to fix this.

lamberta (approved):

PR #1898, opened 2021-05-23
Update csv.ipynb

VedantMadane:
`/functional.ipynb` page shows a 404 Page not found error where as `/functional` is the page for Keras' functional API page.


8bitmp3 (approved):
Thanks for catching the broken link @VedantMadane Appreciate it!

PR #1904, opened 2021-06-08
direct cuda installation for tensorflow gpu

rahul263-stack:


rahul263-stack:
This file is for cuda installation along with gpu

8bitmp3:
Thank you. The information about cuDNN is already covered in https://www.tensorflow.org/install/gpu#software_requirements with links to NVIDIA docs. In addition, to install TensorFlow with PyPI (the latest is v2.5), `pip install tensorflow` should be enough.

PR #1906, opened 2021-06-10
`steps_per_execution` is no more experimental

rabinadk1:
Change the `experimental_steps_per_execution` to `steps_per_execution` in TPU Guide


MarkDaoust (approved):
Thanks.

PR #1909, opened 2021-06-17
Update classification.ipynb

Saduf2019:
added missing "plt.show()" statement.

Fix for #50231


8bitmp3:
Hi @Saduf2019 Thanks for the PR! We have a fix on the way, so I will close this. Note that it's probably best to edit the code in a notebook instead of a JSON. If you want to know more about that process, we can show you how to do it.

```
        "plot_value_array(1, predictions_single[0], test_labels)\n",
        "_ = plt.xticks(range(10), class_names, rotation=45)"
        "plt.show()"
```
(The second to last line is probably missing `\n",`.)

PR #1912, opened 2021-06-24
Fix validation dataset

drosin:
The validation dataset is the same as training dataset.
This is fixed by taking 80% of train for training and 20% for
validation.


drosin:
@googlebot I signed it!

MarkDaoust (approved):
Thanks!

PR #1919, opened 2021-07-16
update gpu turtorial

fsx950223:


8bitmp3 (approved):
Thank you. LGTM. @MarkDaoust @lamberta PTAL - the `tf.config` API [changed in v2.1] (https://github.com/tensorflow/tensorflow/releases/tag/v2.1.0).

MarkDaoust (commented):
Thanks for the fixes. 

The original author thought it was important to demonstrate this with 2 GPU available. 

That's why the outputs are included with 2.

Does the doc still make sense if there's only one?

MarkDaoust (approved):
I did a quick read through. This seems fine. The fact that there were two GPUs before doesn't seem important.

PR #1922, opened 2021-07-19
Fix load_data/text.ipynb

18520339:
1. Change tf-text to stable version:
- I got this error when using the nightly version in Colab and fixed it with the stable version

Screenshot 2021-07-19 084024

2. Use Softmax activation in export_model that have num_labels > 1:
- For models that have num_labels > 1, I think it would be more accurate when using Softmax activation instead of Sigmoid at the last layer although Sigmoid still get the correct result

3. Use BinaryCrossentropy loss for export_model in the [last section](https://www.tensorflow.org/tutorials/load_data/text?hl=en#export_the_model_3):
- Model in this section is for the Binary classification task, so I think the export_model would be more accurate when using Sigmoid activation at the last layer and BinaryCrossentropy for the loss function

### DIFF
--- a/site/en/tutorials/load_data/text.ipynb
+++ b/site/en/tutorials/load_data/text.ipynb
@@ -2,16 +2,21 @@
   "cells": [
     {
       "cell_type": "markdown",
-      "source": [
-        "##### Copyright 2018 The TensorFlow Authors.\n"
-      ],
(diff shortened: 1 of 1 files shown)


18520339:
@googlebot I signed it!

PR #1925, opened 2021-07-29
fix: https://github.com/tensorflow/tensorflow/issues/38800#issuecomment-660535430

seunggabi:
> https://github.com/tensorflow/docs-l10n/pull/542

### ModuleNotFoundError : from tensorflow.compat.v1 import *
- https://github.com/tensorflow/tensorflow/issues/38800#issuecomment-660535430


lamberta:
See https://github.com/tensorflow/docs-l10n/pull/542#issuecomment-888479915

lamberta:
(sorry, didn't mean to close this one)

MarkDaoust:
I do mean to close it.

We have many tests that would fail if this were actually not working.

The problem is most likely an issue with the way TF was installed.

I can't really investigate without specific instructions to reproduce this.

And even if we knew the cause we can't go back and fix old releases.

PR #1928, opened 2021-08-02
Update pix2pix.ipynb

philyoun:
We need `tf.newaxis`  in order to be called by the generator.


philyoun:
There was difference between my `test_dataset` and your `test_dataset`.
I figured out this after I ran whole code again in shared colab.
Close this comment.

PR #1935, opened 2021-08-23
typos in the subsection 'Non-strict execution'

Tsingularity:
In the given examples, the unused function is tf.gather instead of tf.math.bincount.

Also, in the webpage, the output of the first example (unused_return_eager) is also wrong. It should report an error instead of a tf.Tensor as displayed in the current webpage.


Tsingularity:
@googlebot I signed it!

8bitmp3 (approved):
LGTM, thanks for spotting this.

PR #1938, opened 2021-09-08
R2.4

chris881:
some question by reading the code .
env = gym.make("CartPole-v0")
num_actions = env.action_space.n
- In the CartPole example . evn is like following . in fact the action space is spaces.Discrete(2). how to define this num_actions when the action space type is spaces.box(low=1,high=1,shape=(3,1))?
- In  run_episode(), there is a line .I really don't understand. here is the code. whati is the exaclty mean [0,0]
 # Sample next action from the action probability distribution
        action = tf.random.categorical(action_logits_t, 2)**[0, 0]**


MarkDaoust:
Something went wrong here, it looks like there was a bad merge.

PR #1950, opened 2021-09-29
Update regression.ipynb

RenuPatelGoogle:
Changed Normalization layer structure from 'preprocessing.Normalization'  to 'tf.keras.layers.Normalization' .

### DIFF
--- a/site/en/tutorials/keras/regression.ipynb
+++ b/site/en/tutorials/keras/regression.ipynb
@@ -419,7 +419,7 @@
       },
       "source": [
         "### The Normalization layer\n",
-        "The `preprocessing.Normalization` layer is a clean and simple way to build that preprocessing into your model.\n",
+        "The `tf.keras.layers.Normalization` layer is a clean and simple way to build that preprocessing into your model.\n",
         "\n",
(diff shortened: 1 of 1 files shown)


lamberta:
Thanks! You'll need to sign the CLA with this account or register it internally: https://cla.developers.google.com/

tharakadesilva:
I was going to create a PR for this issue and I noticed that you have started to address it. However, I have a couple of comments for you:

1. We need to change `preprocessing` to `layers` in line [553](https://github.com/tensorflow/docs/blob/61b354842738dbdc71e6ca0bf3de4c31291eb700/site/en/tutorials/keras/regression.ipynb#L553) as well.
2. We are no longer using the import `from tensorflow.keras.layers.experimental import preprocessing`. Please remove [this](https://github.com/tensorflow/docs/blob/61b354842738dbdc71e6ca0bf3de4c31291eb700/site/en/tutorials/keras/regression.ipynb#L152).
3. [trivial] Since we are importing `from tensorflow.keras import layers` at the start of the file, it might be cleaner to do `layers.Normalization` instead of `tf.layers.Normalization`.

PR #1968, opened 2021-11-04
expand_dims to correctly test multi-hot

danielsparing:
Without expanding dimensions, multi-hot encoded columns are not correctly illustrated (they will be parsed as single-record multi-label, not single-label multi-record).

(Another question is whether we need multi-hot for these columns instead of one-hot, as they are not multi-label?)


MarkDaoust (approved):

MarkDaoust:
Thanks.

8bitmp3:
Thank you @danielsparing and @MarkDaoust. Really appreciate it. I'll review the doc to make sure it passes the CI / Notebook lint test 👍

8bitmp3:
The notebook uses `Keras-native` and the CI testis checking for words found in the inclusive wordlist. LGTM!

PR #1969, opened 2021-11-08
Be more clear about train_step and test_step

MalcolmSlaney:
Be more clear that the train_step and test_step are also moved to the accelerator.  This was not clear before.  This makes it explicit, which should help people understanding TPU efficiency.


8bitmp3:
@rchao Can you PTAL? Thanks!

site/en/guide/distributed_training.ipynb
```diff
@@ -508,7 +508,7 @@
         "Here's what you need to change in your code:\n",
         "\n",
         "1. Create an instance of the appropriate `tf.distribute.Strategy`.\n",
-        "2. Move the creation of Keras model, optimizer and metrics inside `strategy.scope`.\n",
+        "2. Move the creation of Keras model, optimizer and metrics inside `strategy.scope`. Thus the code in the model's run(), train_step(), and test_step() will all be distributed and executed on the accelerator(s).\n",
```
rchao, line 511: I'm not sure what "model's run()" refers to. Also, please backtick the symbols. Thanks!

MalcolmSlaney: Done.  Made the references explicit by describing them as methods.

(Let me know if I need to be even more explicit.)

Thanks for the suggestion.

rchao: Thanks! `train_step()` and `test_step()` make sense to me, but I'm not sure what `run()` refers to because there is not such `run()` method of a Keras model.

MalcolmSlaney: Oops, I should have said call method (not run).  Sorry for the tardy response... I had to look up the right terminology.

MalcolmSlaney:
Ping?  Is the update (using the proper terminology) still pending?  I think I have pushed everything correctly...

MarkDaoust:
Hi, thanks for the ping.

I'll see if I can get this merged.

MarkDaoust (approved):

PR #1977, opened 2021-11-24
Update transfer_learning.ipynb

tilakrayal:
Changed `lr` to `learning_rate`, which removes a deprecation warning.


8bitmp3 (approved):
LGTM

(https://www.tensorflow.org/api_docs/python/tf/keras/optimizers/RMSprop)

PR #1978, opened 2021-11-28
Remove a hard coded constant for the image classification tutorial.

jpalmer:
The constant can be inferred from other parts of the code.  Making this change allows for this example to be generalized more easily.

### DIFF
--- a/site/en/tutorials/images/classification.ipynb
+++ b/site/en/tutorials/images/classification.ipynb
@@ -529,7 +529,7 @@
       },
       "outputs": [],
       "source": [
-        "num_classes = 5\n",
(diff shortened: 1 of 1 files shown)


8bitmp3 (approved):
SG! 

@MarkDaoust WDYT?

```python
class_names = train_ds.class_names
print(class_names)
```

Output (5 classes in the example):
```
['daisy', 'dandelion', 'roses', 'sunflowers', 'tulips']
```

Therefore, 

```python
num_classes = len(class_names)
```

instead of  `num_classes = 5` (hardcoded).

MarkDaoust (approved):

PR #1980, opened 2021-12-01
Functions reference fixed

lucasSaavedra123:
The reference to functions was broken. I changed the link to [this one](https://colab.research.google.com/github/tensorflow/docs/blob/master/site/en/guide/function.ipynb#scrollTo=i2MVoIVaNApG)

### DIFF
--- a/site/en/guide/intro_to_graphs.ipynb
+++ b/site/en/guide/intro_to_graphs.ipynb
@@ -606,7 +606,7 @@
         "id": "PUR7qC_bquCn"
       },
       "source": [
-        "`print` is a *Python side effect*, and there are [other differences](function#limitations) that you should be aware of when converting a function into a `Function`."
+        "`print` is a *Python side effect*, and there are [other differences](https://colab.research.google.com/github/tensorflow/docs/blob/master/site/en/guide/function.ipynb#scrollTo=i2MVoIVaNApG) that you should be aware of when converting a function into a `Function`."
       ]
     },
     {
--- a/site/en/guide/intro_to_graphs.ipynb
+++ b/site/en/guide/intro_to_graphs.ipynb
@@ -606,7 +606,7 @@
         "id": "PUR7qC_bquCn"
       },
       "source": [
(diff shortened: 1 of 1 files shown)


MarkDaoust (approved):
Thanks. That link probably only worked on tensorflow.org.

Relative links work everywhere and keep your context (tensorflow.org, github, colab), the #fragment part works everywhere except colab.

site/en/guide/intro_to_graphs.ipynb
```diff
@@ -606,7 +606,7 @@
         "id": "PUR7qC_bquCn"
       },
       "source": [
-        "`print` is a *Python side effect*, and there are [other differences](function#limitations) that you should be aware of when converting a function into a `Function`."
+        "`print` is a *Python side effect*, and there are [other differences](./function.ipynb#limitations) that you should be aware of when converting a function into a `Function`."
```
8bitmp3, line 609: Thanks @lucasSaavedra123 and @MarkDaoust. Section titles may change in future, unlike the file names, so let's use the following:

```suggestion
        "`print` is a *Python side effect*, and there are other differences that you should be aware of when converting a function into a `Function`. Learn more in the _Limitations_ section of the [Better performance with tf.function](function.ipynb) guide."
```

site/en/guide/intro_to_graphs.ipynb
```diff
@@ -606,7 +606,7 @@
         "id": "PUR7qC_bquCn"
       },
       "source": [
-        "`print` is a *Python side effect*, and there are [other differences](function#limitations) that you should be aware of when converting a function into a `Function`."
+        "`print` is a *Python side effect*, and there are other differences that you should be aware of when converting a function into a `Function`. Learn more in the _Limitations_ section of the [Better performance with tf.function](function.ipynb) guide."
```
8bitmp3, line 609:
```suggestion
        "`print` is a *Python side effect*, and there are other differences that you should be aware of when converting a function into a `Function`. Learn more in the _Limitations_ section of the [Better performance with tf.function](./function.ipynb) guide."
```

PR #1982, opened 2021-12-02
Update transfer_learning_with_hub.ipynb

RenuPatelGoogle:
Updated "val_ds = tf.keras.preprocessing.image_dataset_from_directory" to "val_ds = tf.keras.utils.image_dataset_from_directory" in Dataset section and executed successfully.


MarkDaoust (approved):

8bitmp3:
Thanks again @RenuPatelGoogle Apologies, it looks like this has already been fixed

PR #1993, opened 2021-12-30
Add a complete code example for loading a model

rangsimanketkaew:
I added a complete code example that can be used to confirm if the TensorFlow C++ API (built from source code with bazel) works.


MarkDaoust (commented):
Nice.

Is there a full version of this example somewhere that we can point to with this code?

If there are simple instructions to run this example we could do that in the the notebook a bit like we did here:

https://github.com/tensorflow/docs/blob/master/site/en/install/lang_c.ipynb

rangsimanketkaew:
@MarkDaoust  I have my own repo for simply coding TF: https://github.com/rangsimanketkaew/tensorflow_cpp_api/blob/main/load_model_tensorflow_cpp.md. I can also help write a notebook for in-depth working with TF Ops, Graph, ClientSession, and etc. via C++ API.

There you can also find instructions for both saving and loading a model. This might be helpful for the users to get started as the documentation on the official TF website lacks a gentle introduction to and tutorial of the C++ API.

Let me know if you need anything else.

PR #1995, opened 2022-01-13
Corrected issues with image size in TiledGradients of deepdream tutorial

dreavjr:
There are two issues with how image size is handled in the tiled version of the deepdream tutorial:

Major problem: `img_rolled.shape` is resolved in build time of the tf.function, i.e., it is always [None, None, 3] independently of the actual size of the tensor. This is not immediately apparent in the example because the tile size is larger than the sample image, but becomes evident when a small tile size (e.g., 50) is used.

Minor problem: the coordinates are named counter-intuitively, with `xs` and `x` assigned to the vertical coordinate and `ys` and `y` to the horizontal coordinate.

This pull request solves both issues, switching the variable names, and passing the dynamic image size as a parameter to `TiledGradients.__call__`.


markmcd (approved):
LGTM

PR #2002, opened 2022-01-27
Clarify multi_worker_with_keras.ipynb example code (#1)

Obliman:
Suggesting that the "mnist.py" file be renamed to something else like "mnist_setup.py" since the resulting "import mnist" is confusingly in conflict with the mnist package. This threw me off until I re-read the instructions, since I had placed the code in one file which resulted in errors executing mnist.mnist_dataset and mnist.build_and_compile_cnn_model.

### DIFF
--- a/site/en/tutorials/distribute/multi_worker_with_keras.ipynb
+++ b/site/en/tutorials/distribute/multi_worker_with_keras.ipynb
@@ -194,7 +194,7 @@
         "id": "fLW6D2TzvC-4"
       },
       "source": [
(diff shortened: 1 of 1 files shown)


MarkDaoust (approved):
This seems like a reasonable change, I'm just not sure which "mnist" package are you talking about.

Obliman:
The mnist package could be either:
pip install python-mnist
or
pip install mnist
Both are imported as "import mnist" like the tutorial file

MarkDaoust:
LGTM, Thanks!

8bitmp3 (dismissed):

PR #2010, opened 2022-02-08
Update input.ipynb

sushreebarsa:
Updated the broken link in section using "https://www.tensorflow.org/guide/distributed_training"


MarkDaoust (commented):
I don't think this is right.

It looks like that link is meant to go to the "Use experimental_distribute_values_from_function for arbitrary tensor inputs" section near the bottom of this page... there's an anchor down there. So this is just missing a leading `#`.

MarkDaoust (approved):

PR #2011, opened 2022-02-08

sushreebarsa:
Changed the naming from tf.metrics.BinaryAccuracy to tf.keras.metrics.BinaryAccuracy.


MarkDaoust:
Doesn't seem worth it. Both exist.

PR #2027, opened 2022-02-17
Update custom_training_walkthrough.ipynb

sushreebarsa:
Changed the naming from 'tf.argmax' to  'tf.math.argmax' in the documentation.


8bitmp3 (approved):
LGTM

(Note:`tf.argmax` is an alias)

PR #2039, opened 2022-02-25
Bugfix in concatenating gradient batches in integrated gradients colab.

mhavasi:
tf.stack only accounted for the last batch of gradients, instead of collecting the gradients from all batches. All but the last batch were discarded. The correct implementation is to concatenate the gradient batches by the first axis. This way, all the gradients are correctly accounted for.

Tested the changes by running the notebook in googe colab.


MarkDaoust (approved):
Oops, that was my mistake. Thanks!

PR #2042, opened 2022-03-02
update to checkpoint callback options (save_frequency)

gdarkwah:
introduced the number of batches ('n_batches') option for the save frequency instead of  'batch_size'. Using 'batch_size' works in this tutorial because the length of the training data is 1000 which coincidentally results in a rounded value of ~32 when it is divided by the 'batch_size'. In cases when the number of samples is not 1000, this will result in the model saving at different epoch frequencies other than after every 5 epochs.

the definition of 'save_freq' (https://www.tensorflow.org/api_docs/python/tf/keras/callbacks/ModelCheckpoint#args) clearly refers to the number of batches ('n_batches' in this context) and not the number of samples in a batch ('batch_size').


8bitmp3:
@MarkDaoust @markmcd PTAL

MarkDaoust (approved):
LGTM thanks.

site/en/tutorials/keras/save_and_load.ipynb
```diff
@@ -385,12 +385,17 @@
         "\n",
         "batch_size = 32\n",
         "\n",
+        "# calculate the number of batches per epoch\n",
```
8bitmp3, line 388:
```suggestion
        "# Calculate the number of batches per epoch\n",
```

8bitmp3 (approved):

PR #2043, opened 2022-03-03

pindinagesh:
Changed the naming from tf.metrics.BinaryAccuracy to tf.keras.metrics.BinaryAccuracy


MarkDaoust:
Hi. Thanks for the PR. tf.keras.metrics is the prefered path, but both of these refer to the same object. PR do take some work for us to merge. So if you do want to fix stuff like this it's best to batch a bunch of small fixes into a single PR. One PE each it a lot of overhead for us..

PR #2057, opened 2022-04-09
Fix grammatical error in custom_layer docs

NikSchaefer:
Fix a grammatical error within the custom layer docs:

Change "... common layers as a well as easy ways for you to ..." to "common layers as well as easy ways for you to"


8bitmp3 (approved):
Thank you!

PR #2059, opened 2022-04-11
Correct plot labels of AUPRC

jung-hunsoo:
The AUPRC plot takes precision for x-axis and recall for y-axis.


MarkDaoust (approved):

8bitmp3 (approved):
Thanks for the catch!

PR #2060, opened 2022-04-12
Fix the Bazel version error in tested_build_configuration

gadagashwini:
Tensorflow version 2.8.0 build requires:

> The project you're trying to build requires Bazel 4.2.2

Specified version on [`tensorflow/configure.py`](https://github.com/tensorflow/tensorflow/blob/master/configure.py#L48) is `_TF_MIN_BAZEL_VERSION = '4.2.2'`

### DIFF
--- a/site/en/install/source.md
+++ b/site/en/install/source.md
@@ -419,7 +419,7 @@ Success: TensorFlow is now installed.
 
 <table>
 <tr><th>Version</th><th>Python version</th><th>Compiler</th><th>Build tools</th></tr>
-<tr><td>tensorflow-2.8.0</td><td>3.7-3.10</td><td>GCC 7.3.1</td><td>Bazel 4.2.1</td></tr>
+<tr><td>tensorflow-2.8.0</td><td>3.7-3.10</td><td>GCC 7.3.1</td><td>Bazel 4.2.2</td></tr>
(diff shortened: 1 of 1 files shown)


gadagashwini:
@MarkDaoust ,
When we try to build Tensorflow 2.8.0 using Bazel.
Result says : `The project you're trying to build requires Bazel 4.2.2`

MarkDaoust:
In configure.py:

* [On master the min bazel version is 4.2.2](https://github.com/tensorflow/tensorflow/blob/master/configure.py#L48)
* [On 2.8 the min bazel version is 4.2.1](https://github.com/tensorflow/tensorflow/blob/r2.8/configure.py#L48)

So it won't work with master.

When exactly are you getting that error?
Is the sequence:

* Checkout r2.8 
*  Run configure [pass]
*  bazel build -> `The project you're trying to build requires Bazel 4.2.2`

gadagashwini:
My bad!
Tensorflow branch 2.8.0 works fine with `Bazel 4.2.1`
```
git checkout r2.8 
./configure
bazel build   #Run without any problem
```
Tensorflow master brach requires  `Bazel version 4.2.2`
```
git checkout master 
./configure
bazel build   #The project you're trying to build requires Bazel 4.2.2
```

MarkDaoust:
Thanks for confirming.

Personally, I just use [bazelisk](https://github.com/bazelbuild/bazelisk) so I never have to worry about bazel versions.

PR #2061, opened 2022-04-14
Updated deprecated attributes with the new one

RenuPatelGoogle:
Updated and replaced the deprecated attributes max_tokens with "num_tokens" and vocab_size with "vocabulary_size" to remove the warnings from output.


MarkDaoust (approved):

8bitmp3 (approved):
Thanks for checking the API. Can confirm `num_tokens` is the correct arg. https://www.tensorflow.org/api_docs/python/tf/keras/layers/CategoryEncoding?hl=en

PR #2066, opened 2022-04-26
Updated tf.add naming in TF documentation

sushreebarsa:
Updated the naming from 'tf.add' to 'tf.math.add' in the documentation.


8bitmp3 (approved):
Thanks, LGTM!

fyi if you use the full namespace like `tf.math.add`, `tf.matmul`, and `tf.linalg.inv`, then the doc links should get autogenerated.

PR #2092, opened 2022-06-24
Format tf.data.dataset in music generation tutorial, lint the doc

RenuPatelGoogle:
I have updated the tf.data.dataset link with the correct URL path(https://www.tensorflow.org/api_docs/python/tf/data/Dataset). Earlier this link was routing to tensorflow datasets page.

### DIFF
--- a/site/en/tutorials/audio/music_generation.ipynb
+++ b/site/en/tutorials/audio/music_generation.ipynb
@@ -680,7 +680,7 @@
         "id": "xIBLvj-cODWS"
       },
       "source": [
-        "Next, create a [tf.data.Dataset](https://www.tensorflow.org/api_docs/python/tf/data/Dataset) from the parsed notes."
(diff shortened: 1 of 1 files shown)


site/en/tutorials/audio/music_generation.ipynb
```diff
@@ -680,7 +680,7 @@
         "id": "xIBLvj-cODWS"
       },
       "source": [
-        "Next, create a [tf.data.Dataset](https://www.tensorflow.org/datasets) from the parsed notes."
+        "Next, create a [tf.data.Dataset](https://www.tensorflow.org/api_docs/python/tf/data/Dataset) from the parsed notes."
```
8bitmp3, line 683: Good catch. We could also use the full API name and the link should get auto-generated.

```suggestion
        "Next, create a `tf.data.Dataset` from the parsed notes."
```

8bitmp3 (approved):

PR #2100, opened 2022-07-13
Updated correct link for model subclassing API

RenuPatelGoogle:
The old link was routing to `the Sequential model` page which was not relevant, so updated the model subclassing API link with the exact page.

### DIFF
--- a/site/en/tutorials/quickstart/advanced.ipynb
+++ b/site/en/tutorials/quickstart/advanced.ipynb
@@ -165,7 +165,7 @@
         "id": "BPZ68wASog_I"
       },
       "source": [
-        "Build the `tf.keras` model using the Keras [model subclassing API](https://www.tensorflow.org/guide/keras#model_subclassing):"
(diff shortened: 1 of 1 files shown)


site/en/tutorials/quickstart/advanced.ipynb
```diff
@@ -165,7 +165,7 @@
         "id": "BPZ68wASog_I"
       },
       "source": [
-        "Build the `tf.keras` model using the Keras [model subclassing API](https://www.tensorflow.org/guide/keras#model_subclassing):"
+        "Build the `tf.keras` model using the Keras [model subclassing API](https://www.tensorflow.org/guide/keras/custom_layers_and_models#the_model_class):"
```
8bitmp3, line 168: Thanks @RenuPatelGoogle 

Since the section title may change in future and then break the link, let's just use the main URL here.

8bitmp3:
```suggestion
        "Build the `tf.keras` model using the Keras [model subclassing API](https://www.tensorflow.org/guide/keras/custom_layers_and_models):"
```

8bitmp3 (approved):

PR #2114, opened 2022-08-14
Remove semi-colon symbol when plotting metrics in Classification on imbalanced data tutorial

artemboiko1:
remove unneeded symbol


8bitmp3 (approved):
Thanks!

MarkDaoust:
Note that ending the last line with a `;` prevents jupyter/Colab from printing the object returned by the last function. 

Removing this probably causes  something like `<Legend object 0x234234234>`  to be printed above the plot.

PR #2124, opened 2022-09-09
Fixed broken link in XLA documentation

mohantym:
Fixed broken link in XLA documentation for "Use XLA with tf.function" at line 224


8bitmp3 (approved):
Thanks!

PR #2129, opened 2022-09-23
formula fix

jancervenka:
In the gradient descent with momentum formula, I think the gradient function $L^\\prime$ should use $x^{[t-1]}$ to cumpute the new delta.

$$
\\Delta_x^{[t]} = lr \\cdot L^\\prime(x^{[t-1]}) + p \\cdot \\Delta_x^{[t-1]}
$$


8bitmp3:
@MarkDaoust @markmcd PTAL

markmcd:
LGTM. The whole point is to calculate $x^{[t]}$, so that makes sense.

Thanks for the fix.

PR #2130, opened 2022-09-23
Updated correct link for tf.keras API

RenuPatelGoogle:
Updated correct link for tf.keras API from "https://www.tensorflow.org/guide/keras/sequential_model" to "https://www.tensorflow.org/api_docs/python/tf/keras"


8bitmp3:
The link originally takes users to the first Keras guide. https://www.tensorflow.org/guide/keras has a redirect to the Sequential guide. Closing this PR for now.

PR #2135, opened 2022-10-01
Fixed few grammar mistakes in create_op.md

mohantym:
Fixed few grammar mistakes in create_op.md found using inspect code from Pycharm


8bitmp3:
With the exception of commas, please review the changes suggested by software before submitting them. For example, capitalization of "it" here is incorrect:

```
To make the op available as a regular function `import`-able from a Python
module, It may be useful to have the `load_op_library` call in a Python source
```

PR #2136, opened 2022-10-02
Update video.ipynb

natravedrova:
Use the function's input parameter in the code

list_files_from_zip_url(zip_url) - zip_url is not used in the code, but it seems like it should be used instead of the global variable URL


8bitmp3:
@shilpakancharla PTAL @MarkDaoust @markmcd

MarkDaoust (approved):
Thanks!

PR #2141, opened 2022-10-18
Fix pip update command bug

BLUE-DEVIL1134:
It is a well know fact that **pip cannot update itself while running itself**, so pip has to run as a thread inside `python -m` command.

### DIFF
--- a/site/en/install/pip.md
+++ b/site/en/install/pip.md
@@ -260,7 +260,7 @@ The following NVIDIA® software are only required for GPU support.
     installation to be sure you're running the latest version.
 
     ```bash
-    python -m pip install --upgrade pip
+    pip install --upgrade pip
     ```
 
     Then, install TensorFlow with pip.
@@ -361,7 +361,7 @@ The following NVIDIA® software are only required for GPU support.
     installation to be sure you're running the latest version.
 
     ```bash
-    python -m pip install --upgrade pip
+    pip install --upgrade pip
     ```
 
     Then, install TensorFlow with pip.
@@ -598,7 +598,7 @@ The following NVIDIA® software are only required for GPU support.
     installation to be sure you're running the latest version.
 
     ```bash
-    python -m pip install --upgrade pip
+    pip install --upgrade pip
     ```
 
     Then, install TensorFlow with pip.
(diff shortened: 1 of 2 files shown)


8bitmp3:
@MarkDaoust @shilpakancharla PTAL

markmcd:
> It is a well know fact that pip cannot update itself while running itself

Is this for Windows? It's fine on other platforms. Apologies if this is obvious to everyone running Windows, but could you provide a link for context?

My only suggestion is that `python3` is used as the binary elsewhere in the instructions, so we should be consistent, otherwise users with multiple installations may end up being surprised.

BLUE-DEVIL1134:
@markmcd any update on the pr?

BLUE-DEVIL1134:
@8bitmp3 @markmcd @MarkDaoust guys, any update?

site/en/install/pip.md
```diff
@@ -260,7 +260,7 @@ The following NVIDIA® software are only required for GPU support.
     installation to be sure you're running the latest version.
 
     ```bash
-    pip install --upgrade pip
+    python -m pip install --upgrade pip
```
MarkDaoust, line 263:
```suggestion
    pip install --upgrade pip
```


Please revert.

site/en/install/pip.md
```diff
@@ -361,7 +361,7 @@ The following NVIDIA® software are only required for GPU support.
...
```
MarkDaoust, line 364:
```suggestion
    pip install --upgrade pip
```

Please revert.

site/en/install/pip.md
```diff
@@ -598,7 +598,7 @@ The following NVIDIA® software are only required for GPU support.
...
```
MarkDaoust, line 601:
```suggestion
    pip install --upgrade pip
```

Please revert.

site/en/install/pip.md
```diff
@@ -476,7 +476,7 @@ The following NVIDIA® software are only required for GPU support.
...
```
MarkDaoust, line 479: There are two commands a few paragraphs down, in this "windows native" tab that use `python3` instead of `python`. Can you updater those too?

MarkDaoust (changes requested):

site/en/install/pip.md
8bitmp3: @BLUE-DEVIL1134 PTAL

8bitmp3:
Let's reopen the PR once there's an update.

PR #2145, opened 2022-10-28
Updated syntax to set memory limit on all gpus

mohantym:
Updated syntax to set memory limit on all found GPU's instead of setting only on GPU i.e gpu[0] .  
Attached [gist](https://colab.sandbox.google.com/gist/mohantym/04a7cd00f399f0a058eaca102dec17fe/gpu.ipynb#scrollTo=8EMGuGKbNkc6) as Proof of work.

### DIFF
--- a/site/en/guide/gpu.ipynb
+++ b/site/en/guide/gpu.ipynb
@@ -397,10 +397,10 @@
         "if gpus:\n",
         "  # Create 2 virtual GPUs with 1GB memory each\n",
         "  try:\n",
+        "   for i in range(len(gpus)):",
         "    tf.config.set_logical_device_configuration(\n",
-        "        gpus[0],\n",
-        "        [tf.config.LogicalDeviceConfiguration(memory_limit=1024),\n",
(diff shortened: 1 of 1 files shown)


aliashori75:
> bot

AliAshori75

aliashori75:
> > bot
> 
> @aliashori75

site/en/guide/gpu.ipynb
```diff
@@ -397,10 +397,10 @@
         "if gpus:\n",
         "  # Create 2 virtual GPUs with 1GB memory each\n",
         "  try:\n",
+        "   for i in range(len(gpus)):",
```
MarkDaoust, line 400: The previous "LogicalDeviceConfiguration" was to limit the memory. This one is to simulate having multiple GPUs. If you actually have multiple GPUs, you don't need this. If you have just one, you want multiple LogicalDevices pinned to is. 

So I son't think this change aligns with what we're trying to do here.

PR #2149, opened 2022-11-17
Fixed tag sequence issue in errors.md

mohantym:
Fixed tag sequence issue in errors.md which might create problem while rendering the page in some devices.


8bitmp3 (approved):
Thanks @mohantym  👍

PR #2151, opened 2022-11-19
Fixed "NameError" in regression.ipynb.

nikhil25803:
In **regression.ipynb** file at **docs/site/en/tutorials/keras**, `horsepower_normalizer = layers.Normalization(input_shape=[1,],axis=None)` was used, throwing the `NameError: name 'layers' is not defined`. 
Replaced it with `tf.keras.layers.Normalization(input_shape=[1,],axis=None)` and now it is working fine :)

----

+ You can also find this missing on the [website](https://www.tensorflow.org/tutorials/keras/regression#linear_regression_with_one_variable)

image

### DIFF
--- a/site/en/tutorials/keras/regression.ipynb
+++ b/site/en/tutorials/keras/regression.ipynb
@@ -547,7 +547,7 @@
       "source": [
         "horsepower = np.array(train_features['Horsepower'])\n",
         "\n",
-        "horsepower_normalizer = layers.Normalization(input_shape=[1,], axis=None)\n",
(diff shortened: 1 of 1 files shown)


nikhil25803:
Hey @MarkDaoust | The same is in almost every model created in the notebook. Is it a general convention followed or really a typo :')

MarkDaoust:
Hi @nikhil25803,

Thanks for taking the time to make a PR.

When running a notebook you really need to run the cells in order. I think you missed this line up at the top:

```
from tensorflow.keras import layers
```

So I don't think we need to change anything here.

Have a good weekend!

PR #2157, opened 2022-12-01
Fix metrics in `compile_and_fit`

gilbd:
Passing a `loss` model into the `metrics` argument causes an error while trying to save and load the model:  
```
TypeError: __init__() got an unexpected keyword argument 'reduction' while trying to load keras model
```

Reproduce:  
1. Follow the guide and build a model
2. Save the model using `model.save(path)`
3. Load the model using `tf.keras.models.load_model(path)`


MarkDaoust (approved):
Thanks!

8bitmp3:
Thanks @gilbd and @MarkDaoust !

PR #2159, opened 2022-12-09
Use NEAREST NEIGHBOR method when resizing image mask

balvisio:
IIUC, when resizing the mask, the NEAREST NEIGHBOR interpolation method should be used to prevent the introduction of values of labels that are not present in the original mask.

Currently, if we print the values of sample mask we get:

```
>>> import numpy as np
>>> mask = iter(train_batches).next()[1]))
>>> print(np.unique(mask))

[0.0000000e+00 1.2207031e-04 3.6621094e-04 ... 1.9995728e+00 1.9999390e+00
 2.0000000e+00]
```

After the change:

```
>>> import numpy as np
>>> mask = iter(train_batches).next()[1]))
>>> print(np.unique(mask))

[0. 1. 2.]
```

### DIFF
--- a/site/en/tutorials/images/segmentation.ipynb
+++ b/site/en/tutorials/images/segmentation.ipynb
@@ -182,7 +182,11 @@
       "source": [
         "def load_image(datapoint):\n",
         "  input_image = tf.image.resize(datapoint['image'], (128, 128))\n",
-        "  input_mask = tf.image.resize(datapoint['segmentation_mask'], (128, 128))\n",
+        "  input_mask = tf.image.resize(\n",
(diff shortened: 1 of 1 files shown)


balvisio:
Hello, @MarkDaoust @8bitmp3 would it be possible for someone to review this? Thank you!

MarkDaoust (approved):

MarkDaoust:
Thanks for the ping.

8bitmp3 (approved):

balvisio:
The CI check is failing due to formatting issues in notebook `site/en/tutorials/video/transfer_learning_with_movinet.ipynb`.

PR #2161, opened 2022-12-16
Update segmentation.ipynb

EliaSchiavon:
`class_weight` is still supported for 3+ dimensional inputs with a binary target (e.g. binary classification of images). The problem is when the target is 3+ dimensional. Saying "`Model.fit` does not support the`class_weight` argument for inputs with 3+ dimensions" may generate confusion for the use case I mentioned.


8bitmp3:
Thank you @EliaSchiavon, will take a look

cc @MarkDaoust @markmcd @shilpakancharla

MarkDaoust (approved):
Thanks!

PR #2164, opened 2022-12-22
refactor: removing deprecated argument

gokullan:
The `untar` argument is now deprecated in favor of `extract` according to the information at https://www.tensorflow.org/api_docs/python/tf/keras/utils/get_file. This PR replaces the `untar` argument used in this tutorial's `keras.utils.get_file` method with `extract`.


8bitmp3 (approved):
LGTM

@MarkDaoust

8bitmp3 (approved):

PR #2166, opened 2022-12-27
Fix UFC to UCF errors

nio1814:
There were a few cases where the dataset was referred to as UFC rather than UCF.


8bitmp3 (approved):
Thanks @nio1814 ! 👍 👍

@shilpakancharla 😄

PR #2168, opened 2022-12-28
Simple audio tutorial: macOS fix, typo, do not mix label variables

AndiDog:
The [tutorial outcome](https://www.tensorflow.org/tutorials/audio/simple_audio#run_inference_on_an_audio_file) is currently confusing since the plot shows a totally different prediction than the speech file. In the graph, "go" is allegedly predicted:

Tutorial before

This fixes the bug by using the correct variable for plot labels.

Plus minor fixes (see commit message).

Not sure


8bitmp3:
Thanks @AndiDog

MarkDaoust (approved):
Thanks!

PR #2169, opened 2022-12-29
Improve tutorial classification on imbalanced data

lorentzenchr:
This PR improves the tutorial for classification on imbalanced data, https://www.tensorflow.org/tutorials/structured_data/imbalanced_data:
- Add proper scoring rules
- Add choice of threshold
- Remove oversampling (to be considered a bad practice)

See also https://discuss.tensorflow.org/t/improvements-to-the-tutorial-classification-on-imbalanced-data/13520.

### DIFF
--- a/site/en/tutorials/structured_data/imbalanced_data.ipynb
+++ b/site/en/tutorials/structured_data/imbalanced_data.ipynb
@@ -1233,7 +1233,7 @@
       "source": [
         "Here you can see that with class weights the accuracy and precision are lower because there are more false positives, but conversely the recall and AUC are higher because the model also found more true positives. Despite having lower accuracy, this model has higher recall (and identifies more fraudulent transactions than the baseline model at threshold 50%). Of course, there is a cost to both types of error (you wouldn't want to bug users by flagging too many legitimate transactions as fraudulent, either). Carefully consider the trade-offs between these different types of errors for your application.\n",
         "\n",
-        "Compared to the baseline model with changed threshold, the class weighted model is clearly inferior. The superiority of the baseline model is further confirmed by the lower test loss value (cross entropy and Brier score) and additionally can be seen by plotting the ROC curves of both models together."
+        "Compared to the baseline model with changed threshold, the class weighted model is clearly inferior. The superiority of the baseline model is further confirmed by the lower test loss value (cross entropy and mean squared error) and additionally can be seen by plotting the ROC curves of both models together."
       ]
     },
     {
--- a/site/en/tutorials/structured_data/imbalanced_data.ipynb
+++ b/site/en/tutorials/structured_data/imbalanced_data.ipynb
@@ -449,7 +449,7 @@
         "Good metrics to assess probabilistic predictions are, in fact, **proper scoring rules**. Their key property is that predicting the true probability is optimal. We give two well-known examples:\n",
         "\n",
         "*   **cross entropy** also known as log loss\n",
-        "*   **Brier score** also known as squared error\n",
+        "*   **Mean squared error** also known as the Brier score\n",
         "\n",
         "#### Metrics for deterministic 0/1 predictions\n",
         "\n",
--- a/site/en/tutorials/structured_data/imbalanced_data.ipynb
+++ b/site/en/tutorials/structured_data/imbalanced_data.ipynb
@@ -77,7 +77,7 @@
         "* Define and train a model using Keras (including setting class weights).\n",
         "* Evaluate the model using various metrics (including precision and recall).\n",
         "* Select a threshold for a probabilistic classifier to get a deterministic classifier.\n",
-        "* Try and compare with class weighted modelling"
+        "* Try and compare with class weighted modelling and oversampling."
       ]
     },
     {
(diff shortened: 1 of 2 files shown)


lorentzenchr:
@8bitmp3 Is there any chance to get a first feedback?

site/en/tutorials/structured_data/imbalanced_data.ipynb
```diff
@@ -414,8 +441,22 @@
         "### Understanding useful metrics\n",
         "\n",
         "Notice that there are a few metrics defined above that can be computed by the model that will be helpful when evaluating the performance.\n",
+        "These can be divided into three groups.\n",
+        "\n",
+        "#### Metrics for probability predictions\n",
+        "\n",
+        "As we train our network with the cross entropy as a loss function, it is fully capable of predicting class probabilities, i.e. it is a probabilistic classifier.\n",
+        "Metrics that assess probabilistic predictions and that are, in fact, **proper scoring rules** are:\n",
```
MarkDaoust, line 449: > **proper scoring rules**

This is the first I've seen this term, if it's worth mentioning we should give a brief description of what that means, and why it's important.

site/en/tutorials/structured_data/imbalanced_data.ipynb
```diff
@@ -458,7 +506,7 @@
         "Now create and train your model using the function that was defined earlier. Notice that the model is fit using a larger than default batch size of 2048, this is important to ensure that each batch has a decent chance of containing a few positive samples. If the batch size was too small, they would likely have no fraudulent transactions to learn from.\n",
```
MarkDaoust, line 506: Even though they're in a way equivalent, this line is a clue about why resampling if more effective than reweighting.

- reweighting you see a few important examples in each batch
- resampling you see a much wider distribution each batch.

site/en/tutorials/structured_data/imbalanced_data.ipynb
```diff
@@ -969,9 +1046,9 @@
         "id": "Y5twGRLfNwmO"
       },
       "source": [
-        "### Plot the AUPRC\r\n",
+        "### Plot the AUPRC\n",
```
MarkDaoust, line 1049:
```suggestion
        "### Plot the PRC\n",
```

site/en/tutorials/structured_data/imbalanced_data.ipynb
```diff
@@ -425,14 +466,21 @@
         ">   $\\frac{\\text{true positives}}{\\text{true positives + false positives}}$\n",
         "*   **Recall** is the percentage of **actual** positives that were correctly classified\n",
         ">   $\\frac{\\text{true positives}}{\\text{true positives + false negatives}}$\n",
+        "\n",
+        "**Note:** Accuracy is not a helpful metric for this task. You can have 99.8%+ accuracy on this task by predicting False all the time.  \n",
+        "\n",
+        "#### Other metrices\n",
+        "\n",
+        "The following metrics take into account all possible choices of thresholds $t$, but they are not proper scoring rules and only assess the ranking of predictions, not their absolute values.\n",
```
MarkDaoust, line 474: This sentence is hard to understand without a little more context on "proper scoring rules".

site/en/tutorials/structured_data/imbalanced_data.ipynb
```diff
@@ -1032,7 +1109,7 @@
       "source": [
         "### Calculate class weights\n",
         "\n",
-        "The goal is to identify fraudulent transactions, but you don't have very many of those positive samples to work with, so you would want to have the classifier heavily weight the few examples that are available. You can do this by passing Keras weights for each class through a parameter. These will cause the model to \"pay more attention\" to examples from an under-represented class."
+        "The goal is to identify fraudulent transactions, but you don't have very many of those positive samples to work with, so you would want to have the classifier heavily weight the few examples that are available. You can do this by passing Keras weights for each class through a parameter. These will cause the model to \"pay more attention\" to examples from an under-represented class. Note, however, that this does not increase in any way the amount of information of your dataset. In the end, using class weights is more or less equivalent to changing the output bias or to changing the threshold. Let's see how it works out."
```
MarkDaoust, line 1112: Yes.

site/en/tutorials/structured_data/imbalanced_data.ipynb
```diff
@@ -1153,7 +1230,9 @@
...
-        "Here you can see that with class weights the accuracy and precision are lower because there are more false positives, but conversely the recall and AUC are higher because the model also found more true positives. Despite having lower accuracy, this model has higher recall (and identifies more fraudulent transactions). Of course, there is a cost to both types of error (you wouldn't want to bug users by flagging too many legitimate transactions as fraudulent, either). Carefully consider the trade-offs between these different types of errors for your application."
+        "Here you can see that with class weights the accuracy and precision are lower because there are more false positives, but conversely the recall and AUC are higher because the model also found more true positives. Despite having lower accuracy, this model has higher recall (and identifies more fraudulent transactions than the baseline model at threshold 50%). Of course, there is a cost to both types of error (you wouldn't want to bug users by flagging too many legitimate transactions as fraudulent, either). Carefully consider the trade-offs between these different types of errors for your application.\n",
```
MarkDaoust, line 1233: It might be worth refering back to the fact, from the  previous comment, that this is equivalent to changing the threshold. And the different metric values here are mostly because of that skewing. 

To emphasize this it could be worth while to put markers on the curve plots at 10, 50, 90% thresholds so you can see that some of it is just moving the threshold values around.

MarkDaoust (changes requested):
Sorry about the delay, a lot of people were out for the Christmas holidays.

Thanks for taking the time to make the PR. Generally I support these changes, we jhust havea few little things to discuss. 

Mainly: I'm not convinced that removing the resampling example is the right approach here.

Yes, on the training data resampling/reweighting almost never beat the straight classifier. on the validation set resamplng does show improvements compared to the baseline, and does much better than reweighting. Given that resampling is working better than reweighting, I'm against removing it.

> Would it make sense to emphasize the cross entropy / log loss a bit more?

I don't think you should stop at CrossEntropy because because in many applications you do need to return a 0 or 1, and that has real-world value/costs and those are what you care about. I think the right thing to emphasize is the PRC curve and the relative values/costs of the different types of errors.

site/en/tutorials/structured_data/imbalanced_data.ipynb
lorentzenchr: I'll try to add a single sentence. Under "Read more" is a canonical reference. On top of that, I can recommend to read https://arxiv.org/abs/0912.0902 (knowing that scoring rules and scoring functions coincide for binary classification).

site/en/tutorials/structured_data/imbalanced_data.ipynb
lorentzenchr: Do you have a suggestion?
I added one sentence for proper scoring rules above. Then this clearly says that AUC only assesses the ranking of predictions. Otherwise stated: Best AUC does not guarantee to be close to the true probabilities.

lorentzenchr:
@MarkDaoust Thanks for looking into this PR and your feedback.

#### Probabilistic classifier
> > Would it make sense to emphasize the cross entropy / log loss a bit more?
>
> I don't think you should stop at CrossEntropy because because in many applications you do need to return a 0 or 1, and that has real-world value/costs and those are what you care about. I think the right thing to emphasize is the PRC curve and the relative values/costs of the different types of errors.

I would divide it into 2 steps. The first is modelling: Find a good probabilistic classifier. The statistical forecast literature clearly states that this is to be preferred over deterministic ones. The second step is then to make a decision, i.e. predict 0 or 1, given the predicted class probability. Note that given a good probabilistic classifier, there does not exist a (systematically/in expectation) better decision than the one based on it.

Without knowing the true cost (or cost ratio), the best one can do is - like in this tutorial - to demonstrate different thresholds and plot ROC curves.

In this regard, I noticed that the `EarlyStopping` is using the PRC-AUC (`'val_prc'`). I think this is not the best choice and indeed, setting it to `'val_loss'` gives better results in the end. I would like to chance this, too.

I also think that the differences in the final results, in particular for the over-sampling case, are due to estimation uncertainty, i.e. due to chance and not systematic (confidence intervals would proof it).

site/en/tutorials/structured_data/imbalanced_data.ipynb
lorentzenchr: Thanks, I really learned something new by you pointing this out.

lorentzenchr: Do you know any canonical literatur on this subject? How Ccass imbalance effects minibatch gradient descent?

lorentzenchr:
@MarkDaoust Any change to get this merged?

MarkDaoust:
Thanks for the ping, I give it a final look and try to get it merged.

site/en/tutorials/structured_data/imbalanced_data.ipynb
```diff
@@ -275,6 +274,31 @@
...
+      ]
+    },
+    {
+      "attachments": {},
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "8a_Z_kBmr7Oh"
+      },
+      "source": [
+        "Given the small number of positive labels, this seems about right. In case the average class probability deviates by a large amount, one could use a stratified splitting scheme.\n",
```
MarkDaoust, line 300:
```suggestion
        "Given the small number of positive labels, this seems about right.\n",
```

I try to avoid mentioning new techniques without more info or a link.

site/en/tutorials/structured_data/imbalanced_data.ipynb
```diff
@@ -414,8 +441,22 @@
...
         "Notice that there are a few metrics defined above that can be computed by the model that will be helpful when evaluating the performance.\n",
+        "These can be divided into three groups.\n",
+        "\n",
+        "#### Metrics for probability predictions\n",
+        "\n",
+        "As we train our network with the cross entropy as a loss function, it is fully capable of predicting class probabilities, i.e. it is a probabilistic classifier.\n",
+        "Good metrics to assess probabilistic predictions are, in fact, **proper scoring rules**. Their key property is that predicting the true probability is optimal. We give two well-known examples:\n",
         "\n",
+        "*   **cross entropy** also known as log loss\n",
+        "*   **Brier score** also known as squared error\n",
```
MarkDaoust, line 452:
```suggestion
        "*   **Mean squared error** also known as the Brier score\n",
```

I feel like MSE is the more common name.

site/en/tutorials/structured_data/imbalanced_data.ipynb
```diff
@@ -374,6 +398,8 @@
       "outputs": [],
       "source": [
         "METRICS = [\n",
+        "      keras.metrics.BinaryCrossentropy(name='cross entropy'),  # same as model's loss\n",
+        "      keras.metrics.MeanSquaredError(name='Brier score'),\n",
```
MarkDaoust, line 402:
```suggestion
        "      keras.metrics.MeanSquaredError(name='mse'),\n",
```

site/en/tutorials/structured_data/imbalanced_data.ipynb
MarkDaoust: Good call introducing proper scoring earlier.

> only assess the ranking of predictions, not their absolute values.

I don't understand this. Which ranking & values are we talking about?

> Best AUC does not guarantee to be close to the true probabilities.

Right, but if all you want is a deterministic classifier, we don't care about the true probabilities.

site/en/tutorials/structured_data/imbalanced_data.ipynb
MarkDaoust: I'm having trouble finding a reference. But it seems clear that it would reduce the gradient variance.

site/en/tutorials/structured_data/imbalanced_data.ipynb
```diff
@@ -1153,7 +1231,9 @@
...
+        "Here you can see that with class weights the accuracy and precision are lower because there are more false positives, but conversely the recall and AUC are higher because the model also found more true positives. Despite having lower accuracy, this model has higher recall (and identifies more fraudulent transactions than the baseline model at threshold 50%). Of course, there is a cost to both types of error (you wouldn't want to bug users by flagging too many legitimate transactions as fraudulent, either). Carefully consider the trade-offs between these different types of errors for your application.\n",
+        "\n",
+        "Compared to the baseline model with changed threshold, the class weighted model is clearly inferior. The superiority of the baseline model is further confirmed by the lower test loss value (cross entropy and Brier score) and additionally can be seen by plotting the ROC curves of both models together."
```
MarkDaoust, line 1236:
```suggestion
        "Compared to the baseline model with changed threshold, the class weighted model is clearly inferior. The superiority of the baseline model is further confirmed by the lower test loss value (cross entropy and mean squared error) and additionally can be seen by plotting the ROC curves of both models together."
```

MarkDaoust (approved):

site/en/tutorials/structured_data/imbalanced_data.ipynb
MarkDaoust, line 474:
```suggestion
        "The following metrics take into account all possible choices of thresholds $t$.\n",
```

MarkDaoust (dismissed):

lorentzenchr:
IMHO, https://github.com/tensorflow/docs/pull/2169/commits/885aeadf83ca214db18e3bc65c704e196ed56857 drops an important piece of information: "they only assess the ranking of predictions, not their absolute values". That's a real deficiency of those metrics.

MarkDaoust:
> IMHO, [885aead](https://github.com/tensorflow/docs/commit/885aeadf83ca214db18e3bc65c704e196ed56857) drops an important piece of information: "they only assess the ranking of predictions, not their absolute values", i.e. they are insensitive to (bad) calibration. That's a real deficiency of those metrics.

Thanks for the feedback. Could you help clarify?  Can you give a little more detail here? What do you mean concretely?

> "they only assess the ranking of predictions, not their absolute values"

I'm stumbling here on the fact that an AUPRC of 1.0 includes a perfect deterministic classifier, and a random classifier would give .. 0.5? 0.0? Those seem like absolute reference points to me.

 > insensitive to (bad) calibration
 
 I'm still lost here, can you give an example?

lorentzenchr:
Let's concentrate on AUC: If you add a constant (or multiply by a positive constant) to the probability prediction of a model, AUC does not change. More visual, AUC does tell nothing about a reliability diagram which assesses (auto-) calibration.
References out of my head (you'll notice fast why they are in my head):
- Gini Index and Friends https://dx.doi.org/10.2139/ssrn.4248143
- Model Comparison and Calibration Assessment: User Guide for Consistent Scoring Functions in Machine Learning and Actuarial Practice https://arxiv.org/abs/2202.12780

site/en/tutorials/structured_data/imbalanced_data.ipynb
lorentzenchr: In case you find a reference, I'm really very very interested because the whole "imbalanced class problem" eludes me and it is rarely stated what the actual problem is and, e.g., why on should ever use class weights.
https://doi.org/10.1214/14-AOS1220, for instance, focuses on computational costs.
My current understanding is that "the problem" is mainly a computational one, either CPU/memory efficiency or solver/optimization related, but not a statistical problem (except a small effective sample size).

site/en/tutorials/structured_data/imbalanced_data.ipynb
```diff
@@ -425,14 +466,21 @@
...
         "*   **AUC** refers to the Area Under the Curve of a Receiver Operating Characteristic curve (ROC-AUC). This metric is equal to the probability that a classifier will rank a random positive sample higher than a random negative sample.\n",
         "*   **AUPRC** refers to Area Under the Curve of the Precision-Recall Curve. This metric computes precision-recall pairs for different probability thresholds. \n",
         "\n",
-        "Note: Accuracy is not a helpful metric for this task. You can have 99.8%+ accuracy on this task by predicting False all the time.  \n",
         "\n",
-        "Read more:\n",
-        "*  [True vs. False and Positive vs. Negative](https://developers.google.com/machine-learning/crash-course/classification/true-false-positive-negative)\n",
-        "*  [Accuracy](https://developers.google.com/machine-learning/crash-course/classification/accuracy)\n",
+        "#### Read more:\n",
+        "*   [Strictly Proper Scoring Rules, Prediction, and Estimation](www.stat.washington.edu/people/raftery/Research/PDF/Gneiting2007jasa.pdf)\n",
```
markmcd, line 481: Include the protocol in the URL, or at least start with `//`. Otherwise this is interpreted as a relative path.

MarkDaoust:
```suggestion
        "*   [Strictly Proper Scoring Rules, Prediction, and Estimation](https://www.stat.washington.edu/people/raftery/Research/PDF/Gneiting2007jasa.pdf)\n",
```

MarkDaoust (approved):

PR #2171, opened 2023-01-01
Updates in Docker tutorial

mohantym:
Fixed broken link in Docker tutorial  at text  "GPU support" and Update 2.1 to 2.8 version Replaced "./gpu.md"
with  
https://www.tensorflow.org/install/pip

### DIFF
--- a/site/en/install/docker.md
+++ b/site/en/install/docker.md
@@ -1,36 +1,37 @@
 # Docker
 
-[Docker](https://docs.docker.com/install/){:.external} uses *containers* to
+[Docker](https://docs.docker.com/install/) uses *containers* to
 create virtual environments that isolate a TensorFlow installation from the rest
 of the system. TensorFlow programs are run *within* this virtual environment that
 can share resources with its host machine (access directories, use the GPU,
 connect to the Internet, etc.). The
-[TensorFlow Docker images](https://hub.docker.com/r/tensorflow/tensorflow/){:.external}
+[TensorFlow Docker images](https://hub.docker.com/r/tensorflow/tensorflow/)
 are tested for each release.
 
-Docker is the easiest way to enable TensorFlow [GPU support](./pip.md) on Linux since only the
-[NVIDIA® GPU driver](https://github.com/NVIDIA/nvidia-docker/wiki/Frequently-Asked-Questions#how-do-i-install-the-nvidia-driver){:.external}
+Docker is the easiest way to enable TensorFlow GPU support on Linux since only the
+[NVIDIA® GPU driver](https://github.com/NVIDIA/nvidia-docker/wiki/Frequently-Asked-Questions#how-do-i-install-the-nvidia-driver)
 is required on the *host* machine (the *NVIDIA® CUDA® Toolkit* does not need to
-be installed).
+be installed). If you don't require the use of Docker, refer to
+[this installation guide](./pip.md).
 
 
 ## TensorFlow Docker requirements
 
-1. [Install Docker](https://docs.docker.com/install/){:.external} on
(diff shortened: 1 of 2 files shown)


site/en/install/docker.md
```diff
@@ -8,7 +8,7 @@ connect to the Internet, etc.). The
 [TensorFlow Docker images](https://hub.docker.com/r/tensorflow/tensorflow/){:.external}
 are tested for each release.
 
-Docker is the easiest way to enable TensorFlow [GPU support](./gpu.md) on Linux since only the
+Docker is the easiest way to enable TensorFlow [GPU support](https://www.tensorflow.org/install/pip) on Linux since only the
```
8bitmp3, line 11: `gpu.md` redirects to `https://www.tensorflow.org/install/pip`, so let's do the following:

```suggestion
Docker is the easiest way to enable TensorFlow [GPU support](./pip.md) on Linux since only the
```

site/en/install/docker.md
```diff
@@ -37,7 +37,7 @@ using the following format:
 |-------------|----------------------------------------------------------------------------------------------------------------------|
 | `latest`    | The latest release of TensorFlow CPU binary image. Default.                                                          |
 | `nightly`   | Nightly builds of the TensorFlow image. (Unstable.)                                                                  |
-| *`version`* | Specify the *version* of the TensorFlow binary image, for example\: *2.1.0*                                          |
+| *`version`* | Specify the *version* of the TensorFlow binary image, for example\: *2.8.0*                                          |
```
8bitmp3, line 40:
```suggestion
| *`version`* | Specify the *version* of the TensorFlow binary image, for example\: *2.8.3*                                          |
```

@MarkDaoust @markmcd @shilpakancharla Should we update the TAG names in the table (see https://hub.docker.com/r/tensorflow/tensorflow/tags)

8bitmp3:
@mohantym Thanks!

8bitmp3 (approved):

site/en/install/docker.md
markmcd: I see that em + inline code isn't rendering on the site - I've filed b/264824237 internally to fix it.

> Should we update the TAG names in the table

What are you proposing we change?

site/en/install/docker.md
```diff
@@ -8,7 +8,7 @@ connect to the Internet, etc.). The
 [TensorFlow Docker images](https://hub.docker.com/r/tensorflow/tensorflow/){:.external}
 are tested for each release.
 
-Docker is the easiest way to enable TensorFlow [GPU support](./gpu.md) on Linux since only the
+Docker is the easiest way to enable TensorFlow [GPU support](./pip.md) on Linux since only the
 [NVIDIA® GPU driver](https://github.com/NVIDIA/nvidia-docker/wiki/Frequently-Asked-Questions#how-do-i-install-the-nvidia-driver){:.external}
 is required on the *host* machine (the *NVIDIA® CUDA® Toolkit* does not need to
 be installed).
```
8bitmp3, line 14:
```suggestion
Docker is the easiest way to enable TensorFlow GPU support on Linux since only the
[NVIDIA® GPU driver](https://github.com/NVIDIA/nvidia-docker/wiki/Frequently-Asked-Questions#how-do-i-install-the-nvidia-driver)
is required on the *host* machine (the *NVIDIA® CUDA® Toolkit* does not need to
be installed). If you don't require the use of Docker, refer to [this guide](/pip.md). 
```

@markmcd Rephrasing, following your comments. cc @MarkDaoust

8bitmp3 (approved):

PR #2172, opened 2023-01-03
Remove GPU package location for windows

sachinprasadhs:
Removed broken/non existing GPU python package location for windows, since from Tensorflow 2.11, GPU instructions are different and it is mentioned in the document.


8bitmp3 (approved):
Thank you!

@MarkDaoust @shilpakancharla @markmcd

PR #2173, opened 2023-01-05
Include print function in predictions

nakajimayoshi:
Adds a print function to the predictions section in case the reader wants to verify the functionality of the model using their own environment. 

Seeing is believing. I just found it odd the docs show what seems to be the expected return array, but don't provide the user by default a way to confirm this for themselves using a local environment. Optionally could extend this to the other examples on the page as well.

### DIFF
--- a/site/en/tutorials/quickstart/beginner.ipynb
+++ b/site/en/tutorials/quickstart/beginner.ipynb
@@ -189,7 +189,7 @@
       "outputs": [],
       "source": [
         "predictions = model(x_train[:1]).numpy()\n",
-        "predictions"
+        "print(predictions)"
       ]
     },
     {


MarkDaoust:
> I just found it odd the docs show what seems to be the expected return array, but don't provide the user by default a way to confirm this for themselves using a local environment.

The recomended way to run these in your own environment is to run it as a notebook.

Notebooks and python REPL environments both work the same way: If the last list doesn't assign the result to something, it gets printed. This is used _everywhere_ in our docs.

This particular cell looks a little funny because we catch the value in a variable and print it. 
It doesn't seem helpful to add these prints everywhere, so I'm closing.

Thanks for taking the time to send us a PR, we do appreciate it, but I'm not merging this one.

PR #2177, opened 2023-01-18
Fixing the broken link in tf_records.ipynb

tilakrayal:
Replaced the broken link in **tf_records.ipynb** with the right link


8bitmp3 (approved):
Thanks @tilakrayal 

The old link was for the R1 (TF1) doc, but it's now a 404. cc @MarkDaoust

MarkDaoust:
Hi, thanks for the PR.

This does fix the 404, but generally this old copy of the tf1 docs gets ~no traffic, so we don't really maintain them.

PR #2198, opened 2023-03-06
Fixed serialize_example() example description

eslam69:
- Fixed serialize_example() example description where the mentioned function `create_message()` was meant to be `serialize_example()`.
- Updated the code cell where the `serialize_example()` is used, the `example_observation` variable was initialized but not used, so I filled it with the data, and used it as an argument to  `serialize_example()`.


soosung80 (approved):
LGTM

8bitmp3 (approved):

markmcd:
Can you please run nbfmt, etc, on this PR? If you think we need `vscode` annotations in the notebooks, please open a separate PR or issue and we can discuss that there, so it's in the tooling and applies to all notebooks.

eslam69:
> Can you please run nbfmt, etc, on this PR? If you think we need `vscode` annotations in the notebooks, please open a separate PR or issue and we can discuss that there, so it's in the tooling and applies to all notebooks.

I've run nbfmt & made a new PR #2207, please check it.

PR #2199, opened 2023-03-11
Proto

MarkDaoust:
- Add support for proto-plus
- Make it possible to override the default filter list.


markmcd (approved):
What's the context for this? Does the TF API include `proto-plus` now?  If so, should we require it as a dep, so it's always present?

MarkDaoust:
> What's the context for this? Does the TF API include `proto-plus` now? If so, should we require it as a dep, so it's always present?

I was debugging it a local checkout, I'll explain more in the merge.

PR #2200, opened 2023-03-12
fix minor typo on the image segmentation tutorial

jhhan0:
tf.keras.losses.CategoricalCrossentropy should be changed to tf.keras.losses.SparseCategoricalCrossentropy since the pixel classes (labels) are integers, not one-hot encoded. Code is written correct, but the markdown should fix to prevent potential confusion to readers.


jhhan0:
Everything else is all perfect. Thank you for amazing image segmentation tutorial 👍

markmcd (approved):
LG, thanks!

PR #2203, opened 2023-03-13
Fixed typo

aoot:
I believe this is what the original author intended as the original statement did not make sense.

### DIFF
--- a/site/en/guide/intro_to_modules.ipynb
+++ b/site/en/guide/intro_to_modules.ipynb
@@ -403,7 +403,7 @@
         "id": "4eGaNiQWcK4j"
       },
       "source": [
-        "During distributed (multi-machine) training they can be sharded,  which is why they are numbered (e.g., '00000-of-00001').  In this case, though, there is only one shared.\n",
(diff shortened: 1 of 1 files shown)


soosung80 (dismissed):
LGTM

aoot:
Apologies, didn't know that additional edits would change the existing PR, thus I removed the additional edits.
The changes is now back to right before I dismissed the stale review.

8bitmp3 (approved):
Thanks

site/en/guide/intro_to_modules.ipynb
```diff
@@ -403,7 +403,7 @@
         "id": "4eGaNiQWcK4j"
       },
       "source": [
-        "During distributed (multi-machine) training they can be sharded,  which is why they are numbered (e.g., '00000-of-00001').  In this case, though, there is only have one shard.\n",
+        "During distributed (multi-machine) training they can be sharded,  which is why they are numbered (e.g., '00000-of-00001').  In this case, though, there is only one shared.\n",
```
8bitmp3, line 406:
```suggestion
        "During distributed (multi-machine) training they can be sharded,  which is why they are numbered (e.g., '00000-of-00001').  In this case, though, there is only one shard.\n",
```

8bitmp3 (approved):

PR #2205, opened 2023-03-15
Minor fix in type hint and stringdoc

felipetovarhenao:
*predict_next_note* currently says it returns an ID, where the return type is int, but it actually returns a tuple.


site/en/tutorials/audio/music_generation.ipynb
```diff
@@ -1073,8 +1073,8 @@
         "def predict_next_note(\n",
         "    notes: np.ndarray, \n",
         "    keras_model: tf.keras.Model, \n",
-        "    temperature: float = 1.0) -> int:\n",
-        "  \"\"\"Generates a note IDs using a trained sequence model.\"\"\"\n",
+        "    temperature: float = 1.0) -> Tuple:\n",
```
markmcd, line 1076:
```suggestion
        "    temperature: float = 1.0) -> tuple[int, float, float]:\n",
```

Two things here:
- prefer the built-in `tuple` over `typing.Tuple`, and
- provide type information about the tuple contents.

And the `Tuple` import on line 148 can be removed too.

markmcd (changes requested):
Thanks for the patch!

felipetovarhenao:
@markmcd Suggested changed added!

markmcd (approved):
LG, thanks

PR #2206, opened 2023-03-18
Change the example for non-strict graph execution

hao-peng:
Change the function from `tf.gather` to `tf.slice`, because `tf.gather` doesn't trigger `InvalidArgumentError` under GPU environment. Instead, "0 is stored in the corresponding output value" according to https://www.tensorflow.org/api_docs/python/tf/gather. Therefore there is no difference in the results of https://www.tensorflow.org/guide/intro_to_graphs#non-strict_execution.

After this change, as expected: tf.slice(x, [1], [1]) will trigger the error during eager execution under both CPU and GPU enviroments. It will not trigger the error for graph execution.


MarkDaoust (approved):
Thanks!

PR #2210, opened 2023-03-24
Update source.md

SuryanarayanaY:
I believe `--config=opt` not supported command.Getting the below error.

` ERROR: Config value 'opt' is not defined in any .rc file `

Hence proposing to remove this option from the example code above.


MarkDaoust:
I'm not convinced.


When I run `./configure` it still prints a line about `--config=opt`:

```
Please specify optimization flags to use during compilation when bazel option "--config=opt"
```

I'm running a build now using bazel 5.3, I'll let you know on Monday if it passes. But it's been going okay for 10min.

Did you run `.configure` ?

SuryanarayanaY:
Let me double check with running `./configure` and then will update.

PR #2212, opened 2023-03-29
Wrong x-axis label order

FrankXu1:
Just take a look at this chapter's graph
[www.tensorflow.org/tutorials/audio/simple_audio#run_inference_on_an_audio_file](http://www.tensorflow.org/tutorials/audio/simple_audio)
You may find It is wrong actually. Result is not campatible with the words.
It seems you posted the wrong order of x-axis labels.
The correct order is followed by initial characters by alphabet order.
So I propose it should be ['down', 'go', 'left', 'no', 'right', 'stop', 'up', 'yes']
I am not so sure whether there are any hidden possible order I did not notice from this.
If have, please tell me and then take the correction order.

### DIFF
--- a/site/en/tutorials/audio/simple_audio.ipynb
+++ b/site/en/tutorials/audio/simple_audio.ipynb
@@ -826,7 +826,7 @@
         "x = x[tf.newaxis,...]\n",
         "\n",
         "prediction = model(x)\n",
-        "x_labels = ['up', 'no', 'stop', 'right', 'down', 'left', 'yes', 'go']\n",
+        "x_labels = ['down', 'go', 'left', 'no', 'right', 'stop', 'up', 'yes']\n",
         "plt.bar(x_labels, tf.nn.softmax(prediction[0]))\n",
         "plt.title('No')\n",
         "plt.show()\n",


MaskedFaceDataSet:
I agreed to @FrankXu1.
After we reorder the label, the model predicted correctly.

I am not sure if we can put the label in this way or not. but seem the model work well after reorder the label and testing with other inferent data.

vijayasarathib:
This looks good to me

site/en/tutorials/audio/simple_audio.ipynb
```diff
@@ -826,7 +826,7 @@
         "x = x[tf.newaxis,...]\n",
         "\n",
         "prediction = model(x)\n",
-        "x_labels = ['no', 'yes', 'down', 'go', 'left', 'up', 'right', 'stop']\n",
+        "x_labels = ['down', 'go', 'left', 'no', 'right', 'stop', 'up', 'yes']\n",
         "plt.bar(x_labels, tf.nn.softmax(prediction[0]))\n",
```
markmcd, line 830:
```suggestion
        "plt.bar(label_names, tf.nn.softmax(prediction[0]))\n",
```

I think instead of hard-coding them here, we can just use the `label_names` variable that is created earlier, and used throughout this guide (so you need to delete the previous line too).

markmcd (changes requested):
LG, just one little request.

WangYihang:
Also there is another error at [Image Classification with TensorFlow Hub](https://www.tensorflow.org/hub/tutorials/image_classification).

In the last code block, the document says the object in the given image is `blowing glass` but actually it is a `leatherback turtle`.

image

I manually ran the notebook, the result shows as follows.

image

site/en/tutorials/audio/simple_audio.ipynb
8bitmp3: @markmcd @FrankXu1 PTAL

PR #2213, opened 2023-03-30
Fixing dataclasses compatibility with python prior to 3.7

eslam69:
fixes tensorflow/tensorflow#60137

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,9 +33,9 @@ REQUIRED_PKGS = [
     'pyyaml',
 ]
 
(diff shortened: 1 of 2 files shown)


markmcd (changes requested):

MarkDaoust:
Hi, thanks for taking the time to send a PR. 

But TensorFlow has dropped support for python <=3.7, and are planning to drop 3.8 for TF 2.13.

So I'm not convinced it makes sense to try and support older versions.

We should probably add `python_requires=">=3.8"` to the setup().

PR #2215, opened 2023-04-04
Make autoencoder easier to copy-paste

beyarkay:
This PR simply removes the explicit `(28, 28)` and `784` which represented the shape of the FMNIST dataset.

It rather uses a `shape` member variable which is passed into the constructor for the Autoencoder. This `shape` variable is set to be `shape = x_test.shape[1:]` by default.

I often come to this page, since I forget exactly how to structure an Autoencoder in TF, and I always end up writing this little snippet out where I extract the `shape` variable into a member variable of the Autoencoder class.


vijayasarathib:
LGTM

markmcd (approved):
Thanks for the submission - this is great.

site/en/tutorials/generative/autoencoder.ipynb
```diff
@@ -159,27 +159,29 @@
       },
       "outputs": [],
       "source": [
-        "latent_dim = 64 \n",
-        "\n",
-        "class Autoencoder(Model):\n",
-        "  def __init__(self, latent_dim):\n",
+        "class Autoencoder(keras.Model):\n",
```
markmcd, line 162: `keras.Model` isn't imported. `Model` is already imported (`from tensorflow.keras.models import Model`), so you could leave this as it was.

markmcd (approved):
👍

beyarkay:
Thanks for the approval! How often are the docs re-built?

markmcd:
This one is updated nightly, but before it's merged it requires another internal review, so could take a day or two.

site/en/tutorials/generative/autoencoder.ipynb
```diff
@@ -159,27 +159,29 @@
...
         "\n",
         "  def call(self, x):\n",
         "    encoded = self.encoder(x)\n",
         "    decoded = self.decoder(encoded)\n",
         "    return decoded\n",
-        "  \n",
-        "autoencoder = Autoencoder(latent_dim) "
+        "\n",
+        "\n",
+        "shape = x_test.shape\n",
```
markmcd, line 182:
```suggestion
        "shape = x_test.shape[1:]\n",
```

I tried to amend this PR myself but was denied - can you add this? You described this in the comment but didn't actually implement it in the code.

markmcd (changes requested):

site/en/tutorials/generative/autoencoder.ipynb
beyarkay: Ah, apologies, I'll fix this

beyarkay:
Thanks for helping out!

PR #2217, opened 2023-04-05
Fix pip install docs on linux and wsl2

maurerle:
The install on linux still failes if the extended docs are used, because the CUDNN_PATH variable is not set when executing the activate script

An alternative would be to use:

`echo 'export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:'"$CONDA_PREFIX/lib/:$CUDNN_PATH/lib" > $CONDA_PREFIX/etc/conda/activate.d/env_vars.sh`

which would also be idempotent. Yet I did stick to the current suggested way of appending to the existing file.

### DIFF
--- a/site/en/install/pip.md
+++ b/site/en/install/pip.md
@@ -260,8 +260,8 @@ The following NVIDIA® software are only required for GPU support.
 
     ```bash
     mkdir -p $CONDA_PREFIX/etc/conda/activate.d
-    CUDNN_PATH=$(dirname $(python -c "import nvidia.cudnn;print(nvidia.cudnn.__file__)"))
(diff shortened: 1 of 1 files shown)


haifeng-jin (approved):
Thanks for the fix!

8bitmp3 (approved):
@markmcd PTAL, just need LGTM from docs.
@maurerle thanks for the PR!

MarkDaoust (approved):

maurerle:
Thank you for merging this PR! When are the docs rebuild so that the fix is also available here:
https://www.tensorflow.org/install/pip ?

MarkDaoust:
Usually within 24h

PR #2219, opened 2023-04-13
Update macos brew installation instructions from build from source

SuryanarayanaY:
The command for installing brew for macos build instructions is outdated and returning 404 error. As suggested by @markmcd that we don't need to maintain brew installation instructions which may vary at times and we need to track it continuously. Hence I am proposing to remove the brew installation instructions as user can refer the attached link and can do on his own.

Please review and confirm whether this is OK. Thanks!


MarkDaoust (approved):
Thanks!

haifeng-jin (commented):
I do not have the knowlege to review this PR.  The rest of the reviewers can decide. Thanks!

PR #2220, opened 2023-04-20
Updated TF Windows build document

mraunak:
From the recent GitHub issues, we observed that TensorFlow Windows users are getting errors while building the TensorFlow package from the source due to incorrect path setup of the Environmental Variables. This pull request will update the https://www.tensorflow.org/install/source_windows with the steps and commands to set up the paths correctly and guide users to resolve similar issues. The document has also been updated with the latest output of running Python Configure file and GPU support on Native-Windows

This pull request also addresses one minor change in https://www.tensorflow.org/install/pip#windows-native.

### DIFF
--- a/site/en/install/pip.md
+++ b/site/en/install/pip.md
@@ -51,7 +51,7 @@ step-by-step instructions.
     supported GPU on native-Windows.
     Starting with TensorFlow `2.11`, you will need to install
     [TensorFlow in WSL2](https://tensorflow.org/install/pip#windows-wsl2),
-    or install `tensorflow-cpu` and, optionally, try the
+    or install `tensorflow` or `tensorflow-cpu` and, optionally, try the
     [TensorFlow-DirectML-Plugin](https://github.com/microsoft/tensorflow-directml-plugin#tensorflow-directml-plugin-)
 
     ```bash
--- a/site/en/install/pip.md
+++ b/site/en/install/pip.md
@@ -51,7 +51,7 @@ step-by-step instructions.
     supported GPU on native-Windows.
     Starting with TensorFlow `2.11`, you will need to install
     [TensorFlow in WSL2](https://tensorflow.org/install/pip#windows-wsl2),
-    or install `tensorflow` or `tensorflow-cpu` and, optionally, try the
+    or install `tensorflow-cpu` and, optionally, try the
     [TensorFlow-DirectML-Plugin](https://github.com/microsoft/tensorflow-directml-plugin#tensorflow-directml-plugin-)
 
     ```bash
--- a/site/en/install/pip.md
+++ b/site/en/install/pip.md
@@ -26,12 +26,9 @@ step-by-step instructions.
     for more information about this collaboration.
 
     ```bash
-    conda install -c conda-forge cudatoolkit=11.8.0
-    python3 -m pip install nvidia-cudnn-cu11==8.6.0.163 tensorflow==2.12.*
-    mkdir -p $CONDA_PREFIX/etc/conda/activate.d
-    echo 'CUDNN_PATH=$(dirname $(python -c "import nvidia.cudnn;print(nvidia.cudnn.__file__)"))' >> $CONDA_PREFIX/etc/conda/activate.d/env_vars.sh
-    echo 'export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$CONDA_PREFIX/lib/:$CUDNN_PATH/lib' >> $CONDA_PREFIX/etc/conda/activate.d/env_vars.sh
-    source $CONDA_PREFIX/etc/conda/activate.d/env_vars.sh
+    conda install -c conda-forge cudatoolkit=11.2.2 cudnn=8.1.0
+    export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$CONDA_PREFIX/lib/
+    python3 -m pip install tensorflow
     # Verify install:
     python3 -c "import tensorflow as tf; print(tf.config.list_physical_devices('GPU'))"
     ```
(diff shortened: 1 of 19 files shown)


haifeng-jin (changes requested):
Thanks for the PR! I believe the changes in the pip.md are totally accidents. Please change it back and only make the changes needed.

haifeng-jin:
@MarkDaoust Should we just remove the native windows tab on the top of pip page? since we do not support the latest TF on it any more.

site/en/install/source_windows.md
```diff
@@ -223,13 +248,14 @@ Although it is possible to build both CUDA and non-CUDA configs under the
...
 
+
 ### Install the package
 
 The filename of the generated `.whl` file depends on the TensorFlow version and
 your platform. Use `pip3 install` to install the package, for example:
 
 <pre class="devsite-terminal tfo-terminal-windows prettyprint lang-bsh">
-pip3 install C:/tmp/tensorflow_pkg/tensorflow-<var>version</var>-cp36-cp36m-win_amd64.whl
+pip3 install C:/tmp/tensorflow_pkg/tensorflow-<var>version</var>-cp310-cp310-win_amd64.whl
```
MarkDaoust, line 258: Is it possible to use a wild-card here so that the command to copy-paste doesn't depend on the python version.

site/en/install/source_windows.md
```diff
@@ -256,12 +282,12 @@ considered a Unix absolute path since it starts with a slash.)
 
 Add the Bazel and Python installation directories to your `$PATH` environmental
 variable. If Bazel is installed to `C:\tools\bazel.exe`, and Python to
-`C:\Python36\python.exe`, set your `PATH` with:
+`C:\Python310\python.exe`, set your `PATH` with:
```
MarkDaoust, line 285: Is there a way to find out this path so the commands don't change depending on the python version?

MarkDaoust (changes requested):
Thanks for taking the time to fix this. 

`source_windows.md` looks good, but something went wring with `install/pip.md`

It looks like you edited an older version of the file, and resolved the merge conflict by choosing your copy. This PR reverts all the changes that came in after you branched (CUDA 11.8 -> 11.2, TF2.12 -> 2.11). 

I think you need to look at your diff relative to where you branched off, and apply those changes to the version from master.

MarkDaoust:
> @MarkDaoust Should we just remove the native windows tab on the top of pip page? since we do not support the latest TF on it any more.

Not at all anymore? I was under the impression that there was still `tensorflow_cpu`, I could be totally wrong.

mraunak:
Hi @MarkDaoust and @haifeng-jin, I have made changes as per the suggestions.

site/en/install/pip.md
```diff
@@ -51,7 +51,7 @@ step-by-step instructions.
     supported GPU on native-Windows.
     Starting with TensorFlow `2.11`, you will need to install
     [TensorFlow in WSL2](https://tensorflow.org/install/pip#windows-wsl2),
-    or install `tensorflow-cpu` and, optionally, try the
+    or install `tensorflow` or `tensorflow-cpu` and, optionally, try the
```
MarkDaoust, line 54:
```suggestion
    or install `tensorflow-cpu` and, optionally, try the
```

It looks like there's a TensorFlow 2.12 package for windows here:

https://pypi.org/project/tensorflow/#files

But it's only 2kB -> broken.   I think we should fix that.  What happens right now if you "pip install tensorflow" on windows?

MarkDaoust (approved):

site/en/install/pip.md
MarkDaoust: I think we don't want a `tensorflow` here. I'll patch.

site/en/install/source_windows.md
```diff
@@ -69,6 +69,8 @@ Note: TensorFlow is tested against the *Visual Studio 2019*.
 See the Windows [GPU support](./gpu.md) guide to install the drivers and
 additional software required to run TensorFlow on a GPU.
 
+Note: GPU support on native-Windows is only available for 2.10 or earlier versions, starting in TF 2.11, CUDA build is not supported for Windows. For using TensorFlow GPU on Windows, you will need to build/install TensorFlow in WSL2 or use TensorFlow-DirectML-Plugin
```
MarkDaoust, line 72: "or use `tensorflow_cpu` with the TensorFlow-DirectML-Plugin."

site/en/install/source_windows.md
```diff
@@ -91,6 +93,28 @@ git checkout <em>branch_name</em>  # r1.9, r1.10, etc.
...
 a release branch that is known to work.
 
+## Optional: Environmental Variable Set Up
+Run following commands before running build command to avoid issue with package creation:
+(If the below commands were set up while installing the packages, please ignore them). Run `set` check if all the paths were set correctly, run `echo %Environmental Variable%` e.g. `echo %BAZEL_VC%` to check path set up for a specific Environmental Variable
+
+ Python path set up issue [tensorflow:issue#59943](https://github.com/tensorflow/tensorflow/issues/59943),[tensorflow:issue#9436](https://github.com/tensorflow/tensorflow/issues/9436),[tensorflow:issue#60083](https://github.com/tensorflow/tensorflow/issues/60083)
+
+<pre class="devsite-terminal tfo-terminal-windows devsite-click-to-copy">
+set PATH=path/to/python [e.g. (C:/Python310)]
```
MarkDaoust, line 103:
```suggestion
set PATH=path/to/python # [e.g. (C:/Python310)]
```

site/en/install/source_windows.md
```diff
@@ -91,6 +93,28 @@ git checkout <em>branch_name</em>  # r1.9, r1.10, etc.
...
+set PATH=path/to/python/Scripts [e.g. (C:/Python310/Scripts)]
```
MarkDaoust, line 104:
```suggestion
set PATH=path/to/python/Scripts # [e.g. (C:/Python310/Scripts)] 
```

site/en/install/source_windows.md
MarkDaoust: Is that the right way to comment these? Ideally this would work if someone copy-pasts the line.

site/en/install/source_windows.md
```diff
@@ -184,12 +200,21 @@ bazel build --config=opt //tensorflow/tools/pip_package:build_pip_package
 
 #### GPU support
 
+Note: GPU support on native-Windows is only available for 2.10 or earlier versions, starting in TF 2.11, CUDA build is not supported for Windows. For using TensorFlow GPU on Windows, you will need to build/install TensorFlow in WSL2 or use TensorFlow-DirectML-Plugin
```
MarkDaoust, line 203: `tensorflow_cpu`

MarkDaoust (changes requested):

MarkDaoust:
Thanks. Just a couple of minor things still to fix and we'll get this merged.

site/en/install/pip.md
mraunak: Hey Mark, TensorFlow on Windows is released by my Team at Intel. So 2kB is fine. It internally downloads tensorflow_intel https://pypi.org/project/tensorflow-intel/#files

MarkDaoust (approved):
I'm good with this. 

I thought you said `tensorflow` on windows was still working, but in 7dd83d7401b87e32c14bc61ce5b2f0a11c701863 you remove it? I'm a bit confused, but I'll move to merge and you can always send another PR if you'd prefer it with `tensorflow`.

Thanks!

mraunak:
Hi @MarkDaoust , if it is still not merged, please consider including this revert. I just put back tensorflow, it was removed by mistake on my end.

MarkDaoust (approved):
Let's go.

PR #2232, opened 2023-06-12
Fix incomplete comment

jorgeorpinel:
If the suggestion in the comment above is used as-is, the runtime returns an error due to incompatible data types.

### DIFF
--- a/site/en/guide/tensor.ipynb
+++ b/site/en/guide/tensor.ipynb
@@ -326,7 +326,7 @@
         "a = tf.constant([[1, 2],\n",
         "                 [3, 4]])\n",
         "b = tf.constant([[1, 1],\n",
-        "                 [1, 1]]) # Could have also said `tf.ones([2,2])`\n",
+        "                 [1, 1]]) # Could have also said `tf.ones([2,2], dtype=tf.dtypes.int32)`\n",
         "\n",
         "print(tf.add(a, b), \"\\n\")\n",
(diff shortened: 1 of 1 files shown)


site/en/guide/tensor.ipynb
```diff
@@ -326,7 +326,7 @@
         "a = tf.constant([[1, 2],\n",
         "                 [3, 4]])\n",
         "b = tf.constant([[1, 1],\n",
-        "                 [1, 1]]) # Could have also said `tf.ones([2,2])`\n",
+        "                 [1, 1]]) # Could have also said `tf.ones([2,2], dtype=tf.dtypes.int32)`\n",
```
markmcd, line 329:
```suggestion
        "                 [1, 1]]) # Could have also said `tf.ones([2,2], dtype=tf.int32)`\n",
```

Given this is an EOL comment, lets keep it concise.

markmcd (dismissed):
Make sure nbfmt passes, but otherwise LGTM. Thanks for the contribution!

jorgeorpinel:
Thanks @markmcd ! It looks like the format error is not related to my change though. Should I still patch that in this same PR?

image

markmcd:
> Thanks @markmcd ! It looks like the format error is not related to my change, though. Should I still patch that in this same PR?
> 
> image

oic, no that's OK. I'll send a follow-up.

markmcd (approved):

PR #2237, opened 2023-06-27
Added Activation Function and Increased Training Epochs in the CSV Tutorial

juelg:
This PR adds non-linear ReLU activation functions to the CSV tutorial in order to make the networks non-linear. I have noticed that the activation functions are missing as they are set to None when not explicitly passed and also, as to my knowledge, not added anywhere else. This makes the network linear which is probably not what we want in this context as we don't know whether the dataset is linear. Furthermore, stacking several layers does not make much sense in the linear case as it does not add any more capacity to the network.

Using a non-linear neural network also increases the training performance as shown in the evaluation below. However, the difference is only visible if trained for a little longer, this is why I increased the training epochs to 30.

Training on the abalone dataset for 30 epochs over $n=10$ runs **without** an activation function leads to the following loss (mean $\pm$ std):
$5.10\pm0.04$

Training with the same conditions but **with** ReLU as activation function:
$4.90\pm0.03$

The training examples later in the tutorial show a similar performance increase when ReLU is used.

### DIFF
--- a/site/en/tutorials/load_data/csv.ipynb
+++ b/site/en/tutorials/load_data/csv.ipynb
@@ -226,7 +226,7 @@
       "outputs": [],
       "source": [
         "abalone_model = tf.keras.Sequential([\n",
-        "  layers.Dense(64),\n",
+        "  layers.Dense(64, activation='relu'),\n",
         "  layers.Dense(1)\n",
         "])\n",
         "\n",
@@ -251,7 +251,7 @@
       },
       "outputs": [],
       "source": [
-        "abalone_model.fit(abalone_features, abalone_labels, epochs=10)"
+        "abalone_model.fit(abalone_features, abalone_labels, epochs=30)"
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


site/en/tutorials/load_data/csv.ipynb
```diff
@@ -251,7 +251,7 @@
       },
       "outputs": [],
       "source": [
-        "abalone_model.fit(abalone_features, abalone_labels, epochs=10)"
+        "abalone_model.fit(abalone_features, abalone_labels, epochs=30)"
```
MarkDaoust, line 254:
```suggestion
        "abalone_model.fit(abalone_features, abalone_labels, epochs=10)"
```

site/en/tutorials/load_data/csv.ipynb
```diff
@@ -337,14 +337,14 @@
...
-        "  layers.Dense(64),\n",
+        "  layers.Dense(64, activation='relu'),\n",
         "  layers.Dense(1)\n",
         "])\n",
         "\n",
         "norm_abalone_model.compile(loss = tf.keras.losses.MeanSquaredError(),\n",
         "                           optimizer = tf.keras.optimizers.Adam())\n",
         "\n",
-        "norm_abalone_model.fit(abalone_features, abalone_labels, epochs=10)"
+        "norm_abalone_model.fit(abalone_features, abalone_labels, epochs=30)"
```
MarkDaoust, line 347:
```suggestion
        "norm_abalone_model.fit(abalone_features, abalone_labels, epochs=10)"
```

site/en/tutorials/load_data/csv.ipynb
```diff
@@ -678,7 +678,7 @@
       },
       "outputs": [],
       "source": [
-        "titanic_model.fit(x=titanic_features_dict, y=titanic_labels, epochs=10)"
+        "titanic_model.fit(x=titanic_features_dict, y=titanic_labels, epochs=30)"
```
MarkDaoust, line 681:
```suggestion
        "titanic_model.fit(x=titanic_features_dict, y=titanic_labels, epochs=10)"
```

MarkDaoust (changes requested):
Right, thanks!

But I'd prefer to keep 10 epochs, just to keep things quick. The point is just to show that it works, not to get all the way to convergence.

MarkDaoust (approved):

MarkDaoust (approved):

PR #2246, opened 2023-07-16
fix: inconsistent single host multi-device distributed guide

YangWu1227:
In reference to the Keras [documentation](https://keras.io/guides/distributed_training/#singlehost-multidevice-synchronous-training) regarding single-host multi-device training, both model creation and compilation are advised to be executed within the `strategy.scope`. However, the official TensorFlow [guide](https://www.tensorflow.org/guide/distributed_training#use_tfdistributestrategy_with_keras_modelfit) shows `model.compile(...)` outside of `strategy.scope`. Possible inconsistency? Appreciate any clarification on this matter and I've opened this PR just in case. Thank you.


MarkDaoust (approved):
Thanks.

PR #2251, opened 2023-08-11
Update pip.md

ForeverZyh:
Swap the order of $CUDNN_PATH/lib and CONDA_PREFIX/lib.

I was hitting an error, `Loaded runtime CuD
NN library: 8.1.0 but source was compiled with: 8.6.0.`
I find out that my `CONDA_PREFIX/lib` contains `libcudnn8.1.0` so TF will first use this version instead of the `libcudnn8.6.0` installed in the `CONDA_PREFIX/lib`.


8bitmp3:
@markmcd

vijayasarathib:
Looks good to me! The change makes sense given the issue described with the CuDNN library versions. I approve this change.

markmcd (approved):
Not sure how prevalent this issue is, if you have any more context on how you got into that state it'd be helpful.

That said, this is harmless, and I am generally in favour of ordering `$PATH` vars by specificity, which this does.

PR #2257, opened 2023-08-26
pandas df loading tutorial: Fix incorrect super().__init__() call

vaharoni:
In the beginner's tutorials > load and process data > pandas.DataFrame there is the following code:
```python
class MyModel(tf.keras.Model):
  def __init__(self):
    # Create all the internal layers in init.
    super().__init__(self)

    self.normalizer = tf.keras.layers.Normalization(axis=-1)
    # ...
```
It seems like `super().__init__(self)` is incorrect here, sending `self` as one of the args to the base class `tf.keras.Model`. The documentation of `tf.keras.Model` uses `super().__init__()`, so I suggest to align it.


MarkDaoust (approved):

MarkDaoust:
Thanks. 

Yeah, how does that pass our tests?

PR #2258, opened 2023-08-27
Quickstart advanced tutorial: simplify super().__init__() call

vaharoni:
In the Quickstart for experts tutorial there is the following code:
```python
class MyModel(Model):
  def __init__(self):
    super(MyModel, self).__init__()
    # ...
```
I suggest to use `super().__init__()` instead as described everywhere else.


MarkDaoust (approved):
Yes. Thanks!

PR #2260, opened 2023-08-27
Loading text tutorial: fixed OOV handling

vaharoni:
In the Beginners tutorials > Load and preprocess data > Text, under Example 2, I believe there is an issue handling OOV tokens. The code does this:
```python
keys = vocab
values = range(2, len(vocab) + 2)  # Reserve `0` for padding, `1` for OOV tokens.

init = tf.lookup.KeyValueTensorInitializer(
    keys, values, key_dtype=tf.string, value_dtype=tf.int64)

num_oov_buckets = 1
vocab_table = tf.lookup.StaticVocabularyTable(init, num_oov_buckets)
```
However, note that `StaticVocabularyTable` does not return the value 1 for OOV, but rather `hash(<term>) % num_oov_buckets + vocab_size` per its documentation. In our case, this is equal to vocab_size, which is actually 10000 (and not 1 nor 10002). This in fact collides with the token in the original index 9998. To fix this, I suggest to first map the OOV to 10002 by providing vocabulary that includes the padding and OOV tokens inherently:
```python
# Reserve `0` for padding, `1` for OOV tokens.
keys = ['', '[UNK]'] + vocab
values = range(len(keys))
```
Then, in the subsequent cell remap the 10002 to 1. I'm using `len(keys)` here rather than `vocab_size + 2` as later in the tutorial we do `vocab_size += 2`.
```python
def preprocess_text(text, label):
  standardized = tf_text.case_fold_utf8(text)
  tokenized = tokenizer.tokenize(standardized)
  vectorized = vocab_table.lookup(tokenized)
  # StaticVocabularyTable returns the OOV token as vocab_size + 2. We overwrite it to be 1.
  vectorized = tf.where(vectorized == len(keys), tf.constant(1, dtype=tf.int64), vectorized)
  return vectorized, label
```

### DIFF
--- a/site/en/tutorials/load_data/text.ipynb
+++ b/site/en/tutorials/load_data/text.ipynb
@@ -1140,8 +1140,9 @@
       },
       "outputs": [],
       "source": [
-        "keys = vocab\n",
-        "values = range(2, len(vocab) + 2)  # Reserve `0` for padding, `1` for OOV tokens.\n",
+        "# Reserve `0` for padding, `1` for OOV tokens.\n",
+        "keys = ['', '[UNK]'] + vocab\n",
+        "values = range(len(keys))\n",
         "\n",
         "init = tf.lookup.KeyValueTensorInitializer(\n",
         "    keys, values, key_dtype=tf.string, value_dtype=tf.int64)\n",
@@ -1171,6 +1172,8 @@
         "  standardized = tf_text.case_fold_utf8(text)\n",
         "  tokenized = tokenizer.tokenize(standardized)\n",
         "  vectorized = vocab_table.lookup(tokenized)\n",
+        "  # StaticVocabularyTable returns the OOV token as vocab_size + 2. We overwrite it to be 1.\n",
+        "  vectorized = tf.where(vectorized == len(keys), tf.constant(1, dtype=tf.int64), vectorized)\n",
         "  return vectorized, label"
       ]
     },


MarkDaoust (dismissed):
Nice catch. Thanks for all the PRs!

I did a quick edit to use the builtin OOV indexes instead of remapping, and added a cell to test run it.

vaharoni:
This approach works for that section. However, note that in the subsequent section "Export the model", we do the following:
```text
To make the model capable of taking raw strings as input, you will create a Keras TextVectorization layer that performs the same steps as your custom preprocessing function. Since you have already trained a vocabulary, you can use TextVectorization.set_vocabulary (instead of TextVectorization.adapt), which trains a new vocabulary.
```
```python
preprocess_layer = TextVectorization(
    max_tokens=vocab_size,
    standardize=tf_text.case_fold_utf8,
    split=tokenizer.tokenize,
    output_mode='int',
    output_sequence_length=MAX_SEQUENCE_LENGTH)

preprocess_layer.set_vocabulary(vocab)

export_model = tf.keras.Sequential(
    [preprocess_layer, model,
     layers.Activation('sigmoid')])
```
We do not train `export_model`. We just export it. However, there is a subtle difference here. The trained model that used `StaticVocabularyTable` returns 10001 for the OOV token. The exported model using `TextVectorization` returns 1 for the OOV token. This could impact inference, and the statement "you will create a Keras TextVectorization layer that performs the same steps as your custom preprocessing function" seems incorrect.

MarkDaoust:
Thanks again. Why is it setup that way?  I'll fix it.

MarkDaoust:
Okay, I've removed my commits, let's merge this, and I'll fix the weirdness in a separate PR.

PR #2267, opened 2023-09-03
Fix predict_next_note() param

MacalusoJeff:
The parameter name is `keras_model` but the function uses `model`, updating the parameter name to match the function
image


MarkDaoust:
Yes. Thanks!

PR #2284, opened 2023-11-09
Update time_series.ipynb

Muneeb21-hub:
indices not indexes


pcoet:
@Muneeb21-hub Thanks for the PR! To be honest, I prefer "indices" too, but the Google Dev Style prefers "indexes": https://developers.google.com/style/word-list#index. Let's leave this as is, since it doesn't seem to be an actual mistake.

PR #2288, opened 2023-11-27
Fix oob index in sparse example

ericcanton:
A sparse matrix with shape [4, 10] is being assigned an element at location [7, 0]. This changes that to the valid [3, 0].

Thanks for your time.


8bitmp3:
thanks @ericcanton, can you also check the CLA - see the message from `google-cla` above for details :+1:

ericcanton:
Thanks for your comment @8bitmp3 . I had to re-make the commit, to use the email address that's on file with my CLA. Find that updated PR [here](https://github.com/tensorflow/docs/pull/2290). I'll close this one!

PR #2289, opened 2023-11-30
Added csv reading

Formiga57:
I'm a aerospace engineering student actually searching for ways to measure sensors influence within a wing with help of ML. Given a lot of researches based on gathering sensors data stored in .csv files or binary ones, we thought it would be a nice feature to have a csv loader or even more formats of data files from sensors or researches, not only mainly Images, Audio or Text.
Then we implemented a "csv_dataset_from_directory()", by now as just an example, method to load a directory with multiple classes of csv, notwithstanding resulting in good results with a tiny amount of our dataset.

Please feel free to give any comments or suggestions about this implementation, we had only about 12 hours to think in a project cause we're in a Hackathon right now. Let us know if you guys had issues inputting that kind of data, therefore for our use in the aerospace engineering would be the exactly feature we're looking for!

[Pull Request into Keras](https://github.com/keras-team/keras/pull/18853)


MarkDaoust:
Hi, thanks for the PR.

Quick feedback: 

1. We can't accept this until `csv_dataset_from_directory` is merged into Keras.
2. I'd prefer not to add a whole new section, to demonstrate csv loading I'd add this function to load_data/csv
3. To simplify reviews, here we submit notebooks without outputs.

PR #2290, opened 2023-12-07
(Take 2) Fix oob index in sparse example

ericcanton:
Re-making this PR commit after updating my Google CLA and GitHub email settings to make this commit using the email address I signed with.

From [my other PR](https://github.com/tensorflow/docs/pull/2288):

A sparse matrix with shape [4, 10] is being assigned an element at location [7, 0]. This changes that to the valid [3, 0].

Thanks for your time.


MarkDaoust (approved):
Why isn't that an error?

PR #2291, opened 2023-12-13
Update cudnn version of TF2.15v

SuryanarayanaY:
The documentation of source.md mentioned `cudnn` version `8.8` is needed for Tf2.15v. However, after installing that, and running TensorFlow, it complains:

2023-12-12 16:48:40.306928: E external/local_xla/xla/stream_executor/cuda/cuda_dnn.cc:447] Loaded runtime CuDNN library: 8.8.0 but source was compiled with: 8.9.4. CuDNN library needs to have matching major version and equal or higher minor version. If using a binary install, upgrade your CuDNN library. If building from sources, make sure the library loaded at runtime is compatible with the version specified during compile .

Even setup.py file contains cudnn as 8.9.4.

https://github.com/tensorflow/tensorflow/blob/0b15fdfcb3fddcb4b8a095f26d0f786e316163d7/tensorflow/tools/pip_package/setup.py#L171

Hence updating it `8.9` version(Ignoring minor version 8.9.4 , hope it is OK).

Fixes TF #[62626](https://github.com/tensorflow/tensorflow/issues/62626)

### DIFF
--- a/site/en/install/source.md
+++ b/site/en/install/source.md
@@ -462,7 +462,7 @@ Success: TensorFlow is now installed.
 
 <table>
 <tr><th>Version</th><th>Python version</th><th>Compiler</th><th>Build tools</th><th>cuDNN</th><th>CUDA</th></tr>
-<tr><td>tensorflow-2.15.0</td><td>3.9-3.11</td><td>Clang 16.0.0</td><td>Bazel 6.1.0</td><td>8.8</td><td>12.2</td></tr>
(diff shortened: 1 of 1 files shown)


haifeng-jin (commented):
I don't have the exact information about this.
We need someone from the dev infra team to confirm the cudnn version here.

SuryanarayanaY:
CC: @MarkDaoust

MarkDaoust:
It'll be easier to get their attention on the merge.

MarkDaoust (approved):

PR #2294, opened 2024-01-29
Add support for `google` in notebook metadata.

markmcd:
These fields will be used to support document-specific metadata that needs to be pushed to the publishing system.


8bitmp3 (approved):
Thanks @markmcd 
cc @MarkDaoust @shilpakancharla

MarkDaoust (approved):

MarkDaoust:
> import/copybara — An error happened while migrating the change

That's a new kind of error.

markmcd:
Looks like this needs reapproval since I renamed the file.

PR #2295, opened 2024-02-13
fixed errors - converted train_features and test_features to tensor, …

gitshiva:
…then updated Horsepower references to specific columns in array, otherwise this nb was failing due to data type mismatches

closes: https://github.com/tensorflow/docs/pull/2313


8bitmp3 (commented):
@markmcd @MarkDaoust

MarkDaoust:
Hi, thanks for pointing this out and taking the time to send a PR.

I always prefer referencing columns by name instead of index, how about we do this:

https://github.com/gitshiva/tensorflow-docs/pull/2

PR #2300, opened 2024-04-08
Update the Segmentation Tutorial Colab to be compatible with Keras 3

kinarr:
I've identified a couple of issues that occur under specific conditions when using TensorFlow 2.16 and Keras 3's plotting utilities, along with their respective solutions. 

#### Issue 1: AttributeError with `plot_model`
- **Issue:** When executing `tf.keras.utils.plot_model(model, show_shapes=True)`, users encounter an AttributeError stating that a 'list' object has no attribute 'shape'.
- **Fix:** Enhancing the `plot_model` function call by adding keyword arguments `expand_nested=True` and `dpi=64`. This not only resolves the AttributeError but also improves the visualization output (might be a little verbose but works without needing any changes to the plotting APIs). The adjusted function call looks like this:
```python
tf.keras.utils.plot_model(model, show_shapes=True, expand_nested=True, dpi=64)
```

#### Issue 2: AttributeError with `SparseCategoricalCrossentropy`
- **Issue:** The following code sequence leads to an AttributeError similar to Issue 1, indicating a 'list' object has no attribute 'shape':
```python
loss = tf.keras.losses.SparseCategoricalCrossentropy(from_logits=True, reduction=tf.keras.losses.Reduction.NONE)
loss(label, prediction, sample_weight).numpy()
```
This occurs when `label` and `prediction` are defined as lists, e.g., `label = [0,0]` and `prediction = [[-3., 0], [-3, 0]]`.
- **Fix:** Converting `label` and `prediction` objects into numpy arrays resolves the issue. The corrected definitions should be as follows:
```python
label = np.array([0,0])
prediction = np.array([[-3., 0], [-3, 0]])
```

**Summary of Changes:**
- Added `expand_nested=True` and `dpi=64` keyword arguments to `plot_model` for enhanced visualization and error prevention.
- Converted `y_true` and `y_pred` objects into numpy arrays before passing them to `SparseCategoricalCrossentropy` to fix the AttributeError.

### DIFF
--- a/site/en/tutorials/images/segmentation.ipynb
+++ b/site/en/tutorials/images/segmentation.ipynb
@@ -817,8 +817,8 @@
   "metadata": {
     "accelerator": "GPU",
     "colab": {
-      "toc_visible": true,
-      "provenance": []
+      "name": "segmentation.ipynb",
+      "toc_visible": true
     },
     "kernelspec": {
       "display_name": "Python 3",
--- a/site/en/tutorials/images/segmentation.ipynb
+++ b/site/en/tutorials/images/segmentation.ipynb
@@ -97,7 +97,10 @@
       },
       "outputs": [],
       "source": [
-        "!pip install git+https://github.com/tensorflow/examples.git"
+        "!pip install git+https://github.com/tensorflow/examples.git\n",
+        "!pip install -U keras\n",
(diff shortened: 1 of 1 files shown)


8bitmp3 (approved):
@MarkDaoust @markmcd

PR #2313, opened 2024-05-30
save model as .tf & labels/features astype float32

mahsanghani:
save model as .tf for h5 bug in saving keras models

linear model train labels and features astype float32


MarkDaoust:
I think I'm merging: https://github.com/tensorflow/docs/pull/2295

mahsanghani:
hi can we merge this so the notebook is not failing, thanks! @MarkDaoust

PR #2316, opened 2024-06-23
Update regression.ipynb to explicitly cast one-hot encoding values

jperk224:
As currently written, the one-hot encoding step leaves the user with boolean dummy values that cause errors later in the tutorial as presently written, when passing the np.array() argument to Normalization.adapt().  The values need to be cast to a numerical type (e.g., int) either at the one-hot encoding step or the adapt() step.


8bitmp3:
@MarkDaoust @markmcd @shilpakancharla PTAL

PR #2317, opened 2024-07-25
Update pandas_dataframe.ipynb

Ayman250:
Passing a dataframe into normalizer.adapt() yields the following error: 
```local variable 'input_shape' referenced before assignment``` 
even if  it can be converted to a tensor with tf.convert_to_tensor. It seems you have to actually pass in the converted tensor or ```np.array(df)``` .

Since it's no longer a dataframe you also can't use iloc.


Ayman250:
The guide is not compatible with TensorFlow 2.16+

PR #2320, opened 2024-08-21
Fix stale issue label

markmcd:
Woops, I used spaces instead of hyphens.


MarkDaoust (approved):

MarkDaoust:
@markmcd .. these files are GH only? so refresh my memory, do we skip the usual import process and just squash/merge here?

MarkDaoust:
> @markmcd .. these files are GH only? so refresh my memory, do we skip the usual import process and just squash/merge here?

There's no copybara check, so yes.

PR #2322, opened 2024-08-28
fixed the tf issue #74564

sanskarmodi8:
I looked into the issue and found out that after the evaluation we are getting a list of size 4 where only the last one is the accuracy of the export_model. So ignored all the other output to only get the required loss and accuracy


site/en/tutorials/keras/text_classification.ipynb
```diff
@@ -861,7 +861,7 @@
         ")\n",
         "\n",
         "# Test it with `raw_test_ds`, which yields raw strings\n",
-        "loss, accuracy = export_model.evaluate(raw_test_ds)\n",
+        "_, _, loss, accuracy = export_model.evaluate(raw_test_ds)\n",
```
MarkDaoust, line 864:
```suggestion
        "metrics = export_model.evaluate(raw_test_ds, return_dict=True)\n",
```

site/en/tutorials/keras/text_classification.ipynb
```diff
@@ -861,7 +861,7 @@
...
         "print(accuracy)"
```
MarkDaoust, line 865:
```suggestion
        "print(metrics)"
```

MarkDaoust (changes requested):

MarkDaoust (approved):

MarkDaoust:
Thanks for the fix! 

I actually always prefer `return_dict=True` because then the response labels each output. It makes it much clearer what's what.

sanskarmodi8:
> Thanks for the fix! 
> 
> I actually always prefer `return_dict=True` because then the response labels each output. It makes it much clearer what's what.

Thank you, Mark! I'll keep that in mind and make sure to use return_dict=True moving forward.
It definitely helps in making the output clearer.

PR #2323, opened 2024-09-04
Fix Text classification with TensorFlow Hub

dnoliver:
* install tf_keras
* use tf_keras model and layers

Run into a problem while running the cell:

> Only instances of `keras.Layer` can be added to a Sequential model. Received: <tensorflow_hub.keras_layer.KerasLayer

Applied the fix described in https://github.com/tensorflow/tensorflow/issues/63849.
After the change, the notebook works end to end. 
But I think this actually downgrades the tf_keras support from 3 to 2, as noted in https://github.com/tensorflow/tensorflow/issues/63849#issuecomment-2011748852.

If the notebook is supposed to work as is, without this fix applied, then this PR shows how to apply the workaround to the notebook until the main issue is figured out :)

### DIFF
--- a/site/en/tutorials/keras/text_classification_with_hub.ipynb
+++ b/site/en/tutorials/keras/text_classification_with_hub.ipynb
@@ -121,7 +121,8 @@
       "outputs": [],
       "source": [
         "!pip install tensorflow-hub\n",
-        "!pip install tensorflow-datasets"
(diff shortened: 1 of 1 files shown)


8bitmp3:
Thank you @dnoliver . Please review the CLA .

@MarkDaoust @markmcd `keras.` instead of `tf_keras.` or both in layer names?

PR #2327, opened 2024-09-11
save checkpoint as .weights.h5

mahsanghani:
When using `save_weights_only=True` in `ModelCheckpoint`, the filepath provided must end in `.weights.h5` (Keras weights format). Received: filepath=./training_checkpoints/ckpt_{epoch}

PR #2328, opened 2024-09-11
optimizer attribute learning_rate

mahsanghani:
model.optimizer.learning_rate.numpy()

to avoid Adam attribute error for "lr"

AttributeError: 'Adam' object has no attribute 'lr'


MarkDaoust:
Merged into #2327

PR #2329, opened 2024-09-11
checkpoint must be saved as a .keras file

mahsanghani:
checkpoint must be saved as a .keras file

checkpoint_prefix = os.path.join(checkpoint_dir, "ckpt_{epoch}.keras")

tf.train.latest_checkpoint returns None
model must be loaded with the appropriate extension

model.load_weights('/content/training_checkpoints/ckpt_12.keras')


8bitmp3:
@MarkDaoust @markmcd

MarkDaoust (changes requested):
Hi, thanks for the fixes, but in the future when there are multiple fixes in one file could you consolidate them into a single PR?

It's easier to review and merge without conflicts. Thanks!

MarkDaoust:
Merged into #2327

PR #2355, opened 2025-02-05
Expose where the issue was found when the linter finds an issue.

Giom-V:


markmcd (approved):
LGTM. we do get false positives in b64 image data sometimes, so this will churn out a crapload of text but this is still good.

PR #2359, opened 2025-02-18
Update dataset_dir path on text_classification.ipynb

Andre-Andreati:
The path on the 'dataset_dir' initialization was wrong, leading to a "FileNotFoundError" in the next cell.


8bitmp3 (approved):

8bitmp3:
@MarkDaoust @markmcd PTAL

PR #2372, opened 2025-04-02
Fix TypeError in simple_audio.ipynb

MahdiHakkakan:
change` x = data_dir/'no/01bb6a2a_nohash_0.wav'`
to `x = data_dir+'/no/01bb6a2a_nohash_0.wav'`
to prevent a TypeError at line 819, caused by attempting to concatenate a Path object with a string.


MarkDaoust (changes requested):
Path/str is legal. 
This notebook is failing at a bunch of other places. 

We can only accept changes it it runs in colab top-to bottom.

PR #2373, opened 2025-04-03
Update images.ipynb

donwanhk:
replace glob with rglob


MarkDaoust:
Please explain why?

donwanhk:
The current notebook is not working, i think the extract process inserted an extra folder, which broke the code.

PR #2374, opened 2025-04-10
Bug fix: Update time_series.ipynb

nectostr:
Running this notebook in Google Colab produces an error because the file being unzipped in a folder and previous code is unable to find it.


MarkDaoust (changes requested):
Please escape the quotes and test.

PR #2379, opened 2025-05-07
tfhub.dev --> kaggle.com/models in overview.md

jorgeorpinel:
As the home page of TF Hub docs, it would be ideal to point to the correct model catalog at this point, IMO.

### DIFF
--- a/site/en/hub/overview.md
+++ b/site/en/hub/overview.md
@@ -2,10 +2,10 @@
 # TensorFlow Hub
 
 TensorFlow Hub is an open repository and library for reusable machine learning.
-The [tfhub.dev](https://tfhub.dev) repository provides many pre-trained models:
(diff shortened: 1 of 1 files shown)


site/en/hub/overview.md
```diff
@@ -14,7 +14,7 @@ code.
 ```python
 import tensorflow_hub as hub
 
-model = hub.KerasLayer("https://tfhub.dev/google/nnlm-en-dim128/2")
+model = hub.KerasLayer("https://www.kaggle.com/models/google/nnlm/TensorFlow2/en-dim128/2")
```
jorgeorpinel, line 17: Note: I based this on the example in https://www.kaggle.com/models/google/nnlm but did not confirm whether this specific code works.

PR #2385, opened 2025-06-11
Update source.md according to Filesystem Hierarchy Standard

cretingame:
According to the [Filesystem Hierarchy Standard (FHS)](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s09.html), /usr/local is specifically intended for software manually installed by the system administrator (i.e., outside the distribution's package manager), which is the case for the clang + LLVM 17 tarball.

## Why /usr/local is preferred over /usr

 * Avoids conflicts: /usr is reserved for distribution-managed software. Installing to /usr might overwrite files from system packages or be overwritten in future package updates.
* Cleaner management: Tools like checkinstall or system backups can more easily identify "local" vs. "distro" files.
* Conformance: Installing to /usr/local conforms to Unix and Linux conventions and best practices.

### DIFF
--- a/site/en/install/source.md
+++ b/site/en/install/source.md
@@ -70,7 +70,7 @@ package sources:
 <code class="devsite-terminal">sudo apt-get update && sudo apt-get install -y llvm-17 clang-17</code>
 </pre>
 
-Now that `/usr/lib/llvm-17/bin/clang` is the actual path to clang in this case.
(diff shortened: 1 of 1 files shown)


haifeng-jin (commented):
I do not have  a good idea on this. Please refer to the rest of the reviewers. Thanks for the PR!

PR #2405, opened 2025-12-09
docs: add 3.13 python support to notes

clintonsteiner:


haifeng-jin (commented):
Thanks for the PR!

I have no idea if 3.13 is well supported or not.
So, please rely on the rest of the reviewers.

